PR #3, opened 2019-09-09
initial redis feature work

aidenkeating:
## Overview

Jira: INTLY-3192

## Verification

- Clone this branch
- Run `make cluster/prepare`
- Run `make run`
- Ensure...


aidenkeating:
@austincunningham Let's split this verification, as there are two different deployment methods in this PR (AWS and OpenShift)?

I'm happy to take the AWS one.

I'm guessing for OpenShift, the deployment steps are as follows:

- Run `oc new-project intly-3192`
- Run `make cluster/prepare`
- Run `make run`
- Ensure the operator picks up on the `Redis` custom resource
- Wait for the `Redis` custom resource status to be updated with a `secretRef`
- Check the `Secret` from the `secretRef` and ensure it contains connection details (internal service url)
- Delete the `Redis` custom resource
- Ensure all redis Kube resources are deleted
- Ensure the Secret is deleted
- Ensure the custom resource is deleted

Makefile
```diff
@@ -43,6 +44,10 @@ cluster/seed/smtp:
 cluster/seed/blobstorage:
 	oc apply -f ./deploy/crds/integreatly_v1alpha1_blobstorage_cr.yaml -n $(NAMESPACE)
 
+.PHONY: cluster/seed/redis
+cluster/seed/redis:
+	oc apply -f ./deploy/crds/integreatly_v1alpha1_redis_crd.yaml -n $(NAMESPACE)
```
austincunningham: Should this be the cr rather that the CRD

aidenkeating: It should

austincunningham:
## AWS Verification
### Install
``` 
oc new-project intly-3192
make cluster/prepare
make cluster/seed/redis #issue with this command doing the same as `make cluster/prepare`
make run
```
- [x] confirmed the Redis custom resource is created
### Apply the CR
edited the cr
```yaml
kind: Redis
metadata:
  name: example-redis
spec:
  secretRef:
    name: redis-test
  tier: development
  type: managed
```
Apply the cr
```bash
oc apply -f ./deploy/crds/integreatly_v1alpha1_redis_cr.yaml
```
- [x] confirm the redis-test secret is created and contain connection url and port

### Delete the redis crd 
- [x] Ensure all redis Kube resources are deleted
- [x] Ensure the Secret is deleted
- [x] Ensure the custom resource is deleted

deploy/crds/integreatly_v1alpha1_redis_cr.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: integreatly.org/v1alpha1
+kind: Redis
+metadata:
+  # name must be between 1-40 characters
+  name: example-redis
+spec:
+  # i want my redis storage information output in a Secret named test in the Namespace cloud-resource-operator
+  secretRef:
+    name: redis-test
```
maleck13: do we want to allow this to target another namespace for the secret? For example ```openshift-3scale ```

deploy/crds/integreatly_v1alpha1_smtpcredentialset_cr.yaml
```diff
@@ -1,13 +1,14 @@
 apiVersion: integreatly.org/v1alpha1
 kind: SMTPCredentialSet
 metadata:
+  # name must be between 1-40 characters
   name: example-smtpcredentialset
 spec:
-  # i want my blob storage information output in a Secret named test in the Namespace cloud-resource-operator
+  # i want my smtp cred information output in a Secret named test in the Namespace cloud-resource-operator
```
maleck13: same with all secret refs, I think we want the secret created in the product namespace?

dimitraz: We do actually do that here: https://github.com/integr8ly/cloud-resource-operator/pull/3/files/a22b5b0546b2e1a574283d992c9ac51e984cc4b7#diff-1e673976e07ef62921f71dadba0ac6f4R148 

So I'll remove the `secretRef` attribute from the CRDs and auto-generate the secret name as well

deploy/crds/integreatly_v1alpha1_redis_cr.yaml
dimitraz: We had that initially but realised it allowed users to create a secret in any random namespace. We're just creating it in the product namespace now

aidenkeating: @maleck13 As the RHMI operator would be creating the CR and reading the secret to then populate whatever the product needs, I don't see the reason for allowing multiple namespaces involved?

pkg/controller/redis/redis_controller.go
```diff
@@ -0,0 +1,177 @@
...
+// Reconcile reads that state of the cluster for a Redis object and makes changes based on the state read
+// and what is in the Redis.Spec
+// TODO(user): Modify this Reconcile function to implement your Controller logic.  This example creates
+// a Pod as an example
+// Note:
+// The Controller will requeue the Request to be processed again if the returned error is non-nil or
+// Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
+func (r *ReconcileRedis) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	ctx := context.TODO()
+	providerList := []providers.RedisProvider{aws.NewAWSRedisProvider(r.client, r.logger), openshift.NewOpenShiftRedisProvider(r.client, r.logger)}
```
maleck13: do we want to do it this way? Generally would avoid creating new objects like this within a function. I would recommend creating the objects outside and passing them in as dependencies.

pkg/providers/aws/provider_redis.go
```diff
@@ -0,0 +1,280 @@
...
+		resources.RemoveFinalizer(&r.ObjectMeta, DefaultFinalizer)
+		if err := p.Client.Update(ctx, r); err != nil {
+			return errorUtil.Wrapf(err, "failed to update instance as part of finalizer reconcile")
+		}
+		return nil
+	}
+	// check if replication group exists and is available
+	if *foundCache.Status == "available" {
+		// delete the redis cluster that was created by the provider
+		_, err = cacheSvc.DeleteReplicationGroup(&elasticache.DeleteReplicationGroupInput{
```
maleck13: How do we decide whether to do a final snapshot after delete?

pkg/providers/aws/provider_redis.go
```diff
@@ -0,0 +1,280 @@
...
+			}
+		}
+	}
+	return nil
+}
+
+// poll for replication groups
+func getReplicationGroups(cacheSvc elasticacheiface.ElastiCacheAPI) ([]*elasticache.ReplicationGroup, error) {
+	var rgs []*elasticache.ReplicationGroup
+	err := wait.PollImmediate(time.Second*5, time.Minute*5, func() (done bool, err error) {
```
maleck13: If we fail here will it end up a requeue?

pkg/controller/redis/redis_controller.go
aidenkeating: We can move this logic to `newReconciler`, and make the internal reconcile function the public function. We would have been testing `reconcile` here

pkg/providers/aws/provider_redis.go
aidenkeating: @maleck13 There is a separate Jira to address final snapshotting during deletion as this ticked ended up being bloated due to OpenShift and AWS being bundled into the same task

pkg/providers/aws/provider_redis_test.go
```diff
@@ -0,0 +1,243 @@
...
+		Region:      "eu-west-1",
+		RawStrategy: json.RawMessage("{}"),
+	}
+	rawStratCfg, err := json.Marshal(sc)
+	if err != nil {
+		t.Fatal("failed to marshal strategy config", err)
+	}
+
+	cases := []struct {
+		name          string
```
maleck13: Do you find breaking it up this way means quite a log of duplication? Would it cause less duplication to have create delete etc as cases under a test_redis method?

dimitraz: Sure, we can look into refactoring that

pkg/providers/aws/provider_redis.go
dimitraz: No we don't, I'll change this

pkg/controller/redis/redis_controller.go
dimitraz: :+1:

deploy/crds/integreatly_v1alpha1_redis_cr.yaml
maleck13: The main reason is to allow the product to use the secret directly rather than having to copy that secret around.  It would work similar to the credential minter. The controlling factor here is having permission to create the resource in the operators namespace.

deploy/crds/integreatly_v1alpha1_smtpcredentialset_cr.yaml
maleck13: So this creates it in the same namespace as the resource is created in. I think this is fine but as long as we expect the resource to be created in the product namespace and as long as we limit the operator group to our product namespaces. We wouldn't want any one in the cluster to be able to create the resource.

go.sum
```diff
@@ -229,6 +229,7 @@ github.com/imdario/mergo v0.3.5/go.mod h1:2EnlNZ0deacrJVfApfmtdGgDfMuh/nq6Ok1EcJ
 github.com/imdario/mergo v0.3.6/go.mod h1:2EnlNZ0deacrJVfApfmtdGgDfMuh/nq6Ok1EcJh5FfA=
 github.com/imdario/mergo v0.3.7 h1:Y+UAYTZ7gDEuOfhxKWy+dvb5dRQ6rJjFSdX2HZY1/gI=
 github.com/imdario/mergo v0.3.7/go.mod h1:2EnlNZ0deacrJVfApfmtdGgDfMuh/nq6Ok1EcJh5FfA=
+github.com/inconshreveable/mousetrap v1.0.0 h1:Z8tu5sraLXCXIcARxBp/8cbvlwVa7Z1NHg9XEKhtSvM=
 github.com/inconshreveable/mousetrap v1.0.0/go.mod h1:PxqpIevigyE2G7u3NXJIT2ANytuPF1OarO4DADm73n8=
```
austincunningham: Same with this one should it be committed, I'm asking as a curiosity

pkg/providers/aws/provider_redis.go
```diff
@@ -0,0 +1,282 @@
...
+
+	"github.com/integr8ly/cloud-resource-operator/pkg/providers"
+
+	errorUtil "github.com/pkg/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	defaultCacheNodeType     = "cache.t2.micro"
+	defaultEngineVersion     = "2.8.24"
```
austincunningham: May need to change these values, believe that 3scale uses redis version 3.2.13, but AWS only supports 3.2.10 see https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/supported-engine-versions.html for more info

go.sum
dimitraz: `make code/fix` runs a `go mod tidy` command so only the dependencies that are used are comitted. So I'm guessing they should be

pkg/providers/aws/provider_redis_test.go
dimitraz: Left it separated for now because we're testing different functions

PR #8, opened 2019-09-24
[INTLY-3551] Initial openshift postgres feature

dimitraz:
## Overview

Deploy an instance of Postgres to Openshift
Jira: https://issues.jboss.org/browse/INTLY-3551

### Creation 
- Run `make cluster/prepare` and `make cluster/seed/postgres`
- Run `make run`
- Ensure the operator picks up on the `Postgres` custom resource
- Wait for the `Postgres` custom resource status to be updated with a `secretRef`
- Check the `Secret` from the `secretRef` and ensure it contains connection details for postgres
- Ensure that the database can be connected to with these details
- Attempt to override the spec for one or more of the provisioned resources by adding the json spec to the `cloud-resources-openshift-strategies` configmap in the `kube-system` namespace. For example, update the user credentials: 
```
{"development": { "strategy": {"secretData":{"user":"test-user","password":"test-password"}} }}
```
- Verify that the `postgres-creds` secret and the `secretRef` secret are updated accordingly with the new credentials 

### Deletion
- Delete the `Postgres` custom resource
- Ensure all resources are deleted (secret, service, pv, pvc, deployment, custom resource)


pkg/controller/postgres/postgres_controller.go
```diff
@@ -0,0 +1,168 @@
...
+			return reconcile.Result{}, nil
+		}
+
+		ps, err := p.CreatePostgres(r.ctx, instance)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+		if ps == nil {
+			r.logger.Info("Secret data is still reconciling, postgres instance is nil")
+			return reconcile.Result{}, errorUtil.New("secret data is still reconciling")
```
ciaranRoche: Would it be worth removing this error here, and replacing it with something similar as we have done with Redis, 

`reconcile.Result{Requeue: true, RequeueAfter: time.Second * 30}, nil`

IMO the error makes it look like a problem while we are just waiting on the postgres instance to become available. We will catch any other errors in the provider. Open to suggestions, but we should make sure we follow the same pattern here across all the controllers.

aidenkeating: let's move to using the suggestion @ciaranRoche made here.

however, we still should have a way of communicating back to anyone who wants detail on the provision progress of the cloud resource. so i think it would make sense here to add a `phase` and `message` block to the status of our resources. `phase` being a one-word indicator of the cloud resources status and `message` being something like `"Waiting for Postgres resources to deploy"`.

we can do the status block updates in a separate pr. @ciaranRoche would you mind creating a ticket around that?

ciaranRoche: Created that Jira -- https://issues.jboss.org/browse/INTLY-3705

dimitraz: :+1:

PR #9, opened 2019-09-30
format aws redis secret to match openshift redis secret

ciaranRoche:
## Overview
Current the redis secret created for AWS and Openshift are different. This change formats the secrets so that they are the same

### Creation 
- Run `make cluster/prepare` and `make cluster/seed/redis`
- Run `make run`
- Verify the secret created is split to `uri` and `port`

### DIFF
--- a/pkg/providers/aws/provider_redis.go
+++ b/pkg/providers/aws/provider_redis.go
@@ -3,7 +3,6 @@ package aws
 import (
 	"context"
 	"encoding/json"
-	"strconv"
 	"time"
 
 	"github.com/aws/aws-sdk-go/service/elasticache/elasticacheiface"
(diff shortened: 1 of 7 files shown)


pkg/providers/aws/provider_redis.go
```diff
@@ -130,7 +129,8 @@ func createRedisCluster(cacheSvc elasticacheiface.ElastiCacheAPI, redisConfig *e
 			logrus.Info("found existing redis cluster")
 			primaryEndpoint := foundCache.NodeGroups[0].PrimaryEndpoint
 			connData := map[string][]byte{
-				"connection": []byte(fmt.Sprintf("%s:%s", *primaryEndpoint.Address, strconv.FormatInt(*primaryEndpoint.Port, 10))),
+				"uri":  []byte(*primaryEndpoint.Address),
```
aidenkeating: To standardise this format we could move this out to a struct, similar to what was done for the SMTP credentials. wdyt?

https://github.com/integr8ly/cloud-resource-operator/blob/52eb01c7bd6f3b8618027054e033c57c755da4cb/pkg/providers/types.go#L29

https://github.com/integr8ly/cloud-resource-operator/blob/master/pkg/providers/aws/provider_smtpcredentialset.go#L36-L52

dimitraz: Oh, I'll update the postgres PR to follow that pattern as well

ciaranRoche: Yeah sounds good :+1:

PR #11, opened 2019-09-30
update delete reconcile to requeue

ciaranRoche:
## Overview
Currently we do not `Requeue on deletion`, which leads to problems when a resource is slow to be deleted. This change adds the `requeue` parameter and increased the reconcile to 30 seconds while the resource is being deleted

## Verification

- Clone this branch
- Run `make cluster/prepare`
- Run `make run`
- Verify the elastiCache cluster is being created
- Delete the redis `example-redis` 
- Verify the log message changes, 
- Verify elastiCache deletes

### DIFF
--- a/pkg/controller/smtpcredentialset/smtpcredentialset_controller.go
+++ b/pkg/controller/smtpcredentialset/smtpcredentialset_controller.go
@@ -116,8 +116,8 @@ func (r *ReconcileSMTPCredentialSet) reconcile(ctx context.Context, request reco
 			if err = p.DeleteSMTPCredentials(ctx, instance); err != nil {
 				return reconcile.Result{}, errorUtil.Wrapf(err, "failed to run delete handler for smtp credentials instance %s", instance.Name)
 			}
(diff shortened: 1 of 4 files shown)


pkg/controller/smtpcredentialset/smtpcredentialset_controller.go
```diff
@@ -116,8 +116,8 @@ func (r *ReconcileSMTPCredentialSet) reconcile(ctx context.Context, request reco
 			if err = p.DeleteSMTPCredentials(ctx, instance); err != nil {
 				return reconcile.Result{}, errorUtil.Wrapf(err, "failed to run delete handler for smtp credentials instance %s", instance.Name)
 			}
-			r.logger.Infof("deletion handler for smtp credential instance %s successful, ending reconciliation", instance.Name)
-			return reconcile.Result{}, nil
+			r.logger.Infof("waiting on successful deletion handler for smtp credential instance %s", instance.Name)
```
aidenkeating: Can we change this log to be similar to the other controllers? `Waiting for SMTP credentials to successfully delete`

PR #15, opened 2019-10-02
[INTLY-3705] Add phase details to custom resources

dimitraz:
## Overview
Add a status phase and message to the custom resources
Jira: https://issues.jboss.org/browse/INTLY-3705

## Verification
To do


pkg/apis/integreatly/v1alpha1/types.go
```diff
@@ -13,9 +19,13 @@ type ResourceTypeSpec struct {
...
 // ResourceTypeStatus Represents the basic status information provided by a resource provider
 type ResourceTypeStatus struct {
-	Strategy  string     `json:"strategy,omitempty"`
-	Provider  string     `json:"provider,omitempty"`
-	SecretRef *SecretRef `json:"secretRef,omitempty"`
+	Strategy      string      `json:"strategy,omitempty"`
+	Provider      string      `json:"provider,omitempty"`
+	SecretRef     *SecretRef  `json:"secretRef"`
+	StatusPhase   StatusPhase `json:"statusPhase,omitempty"`
+	StatusMessage string      `json:"statusMessage,omitempty"`
```
aidenkeating: We can probably leave out the `status` prefix here, as the keys are in the `status` block

pkg/controller/postgres/postgres_controller.go
```diff
@@ -130,10 +130,22 @@ func (r *ReconcilePostgres) Reconcile(request reconcile.Request) (reconcile.Resu
 		// create the postgres instance
 		ps, err := p.CreatePostgres(r.ctx, instance)
 		if err != nil {
+			instance.Status.SecretRef = &v1alpha1.SecretRef{}
+			instance.Status.StatusPhase = v1alpha1.PhaseFailed
+			instance.Status.StatusMessage = "failed to create postgres instance"
```
aidenkeating: If we let the provider functions return a message, or simply expect them to set the message directly on the struct. We could have very specific information being returned in the message, which could be good.

pkg/apis/integreatly/v1alpha1/types.go
dimitraz: Cool

pkg/controller/postgres/postgres_controller.go
dimitraz: Ok makes sense

pkg/apis/integreatly/v1alpha1/zz_generated.openapi.go
```diff
@@ -442,7 +481,20 @@ func schema_pkg_apis_integreatly_v1alpha1_SMTPCredentialSetStatus(ref common.Ref
...
+						SchemaProps: spec.SchemaProps{
+							Type:   []string{"string"},
+							Format: "",
+						},
+					},
+					"statusMessage": {
+						SchemaProps: spec.SchemaProps{
+							Type:   []string{"string"},
+							Format: "",
+						},
```
dimitraz: test

pkg/controller/blobstorage/blobstorage_controller.go
```diff
@@ -146,5 +176,34 @@ func (r *ReconcileBlobStorage) Reconcile(request reconcile.Request) (reconcile.R
...
+	instance.Status.Phase = v1alpha1.PhaseInProgress
+	instance.Status.Message = msg
+	return r.client.Status().Update(r.ctx, instance)
+}
+
+func (r *ReconcileBlobStorage) updatePhaseDeletionInProgress(instance *v1alpha1.BlobStorage, msg v1alpha1.StatusMessage) error {
+	instance.Status.Phase = v1alpha1.PhaseDeleteInProgress
+	instance.Status.Message = msg
+	return r.client.Status().Update(r.ctx, instance)
+}
```
aidenkeating: Although it's only three lines I'd be tempted to de-duplicate these four functions into.

```
func (r *ReconcileBlobStorage) updatePhaseInProgress(inst *v1alpha1.BlobStorage, msg v1alpha1.StatusMessage) error {
    return updatePhase(inst, v1alpha1.PhaseInProgress, msg)
}

...

func (r *ReconcileBlobStorage) updatePhase(inst *v1alpha1.BlobStorage, phase v1alpha1.StatusPhase, msg v1alpha1.StatusMessage) error {
    inst.Status.Phase = phase
    inst.Status.Message = msg
    if err := r.client.Status().Update(r.ctx, inst); err != nil {
        return errors.Wrap(err, "failed to update resource status phase and message")
    }
    return nil
}
```

This'd allow us to add a bit more content into the main body, like wrapping the error, without having to repeat ourselves.

I'm also not overly familiar with it, but we may be able to use the `runtime` `Field` and `SetField` functions for making this generic, instead of specific to `BlobStorage`.

PR #17, opened 2019-10-04
Add an AWS reconciler for Postgres

ciaranRoche:
Co-authored-by: dimitraz <dzuccare@redhat.com>

## Overview

https://issues.jboss.org/browse/INTLY-3090


pkg/providers/aws/provider_postgres.go
```diff
@@ -127,23 +131,24 @@ func createPostgresService(stratCfg *StrategyConfig, providerCreds *AWSCredentia
...
 	if err := p.verifyPostgresConfig(ctx, cr, postgresCfg); err != nil {
-		return nil, errorUtil.Wrap(err, "failed to verify aws postgres cluster configuration")
+		msg := "failed to verify aws postgres cluster configuration"
+		return nil, v1alpha1.StatusMessage(msg), errorUtil.Wrap(err, msg)
 	}
 
 	credSec := &v1.Secret{}
 	if err := p.Client.Get(ctx, types.NamespacedName{Name: cr.Name + defaultCredentialsSec, Namespace: cr.Namespace}, credSec); err != nil {
-		return nil, errorUtil.Wrap(err, "failed to retrieve postgres credential secret")
+		return nil, "failed to retrieve postgres credential secret", errorUtil.Wrap(err, "failed to retrieve postgres credential secret")
```
dimitraz: Do you want to bring this out into a `msg` variable

cover.html
```diff
@@ -0,0 +1,1988 @@
+
```
aidenkeating: Can we remove this?

ciaranRoche: Will do, I thought I added it to gitignore

pkg/providers/aws/provider_postgres.go
```diff
@@ -0,0 +1,487 @@
...
+		updateFound = true
+	}
+	if !updateFound {
+		return nil
+	}
+	return mi
+}
+
+// verify postgres create config
+func (p *AWSPostgresProvider) verifyPostgresCreateConfig(ctx context.Context, pg *v1alpha1.Postgres, postgresCreateConfig *rds.CreateDBInstanceInput) error {
```
aidenkeating: The func name here suggests to me there'd be an error if `pg` or `postgresCreateConfig` was invalid. `buildPostgresCreateStrategy` might be a more descriptive name. wdyt?

pkg/providers/aws/provider_postgres.go
```diff
@@ -0,0 +1,487 @@
...
+	return rds.New(sess)
+}
+
+func (p *AWSPostgresProvider) createPostgresInstance(ctx context.Context, cr *v1alpha1.Postgres, rdsSvc rdsiface.RDSAPI, postgresCfg *rds.CreateDBInstanceInput) (*providers.PostgresInstance, v1alpha1.StatusMessage, error) {
+	// the aws access key can sometimes still not be registered in aws on first try, so loop
+	pi, err := getPostgresInstances(rdsSvc)
+	if err != nil {
+		// return nil error so this function can be requeued
+		logrus.Info("error getting replication groups : ", err)
+		return nil, "error getting replication groups", err
```
aidenkeating: Can we de-dupe this message?

pkg/providers/aws/provider_postgres.go
```diff
@@ -0,0 +1,487 @@
...
+
+	// check if the cluster has already been created
+	var foundInstance *rds.DBInstance
+	for _, i := range pi {
+		if *i.DBInstanceIdentifier == *postgresCfg.DBInstanceIdentifier {
+			foundInstance = i
+			break
+		}
+	}
+	if foundInstance != nil {
```
aidenkeating: As the longer branch of logic is there being a found instance. We could inverse this and keep this a bit flatter

```
if foundInstance == nil {
    // handle create
    // ...
    if _, err = rdsSvc.CreateDBInstance(postgresCfg); err != nil {
        // handle create err
    }
    // handle create success
}
if foundInstance.DBInstanceStatus != "available" {
    return nil, fmt.Sprintf("rds instance creation in progress, current status is %s", foundInstance.DBInstanceStatus), nil
}
// check for update ... 
``

File: pkg/providers/aws/provider_postgres.go [hunk 6]
```diff
@@ -0,0 +1,487 @@
...
+		if *foundInstance.DBInstanceStatus == "available" {
```
aidenkeating: I don't think we need the dereference here, we should be able to reference everything just fine without it

pkg/providers/aws/provider_postgres.go
```diff
@@ -0,0 +1,487 @@
...
+	}
+	return nil, "", nil
+}
+
+func (p *AWSPostgresProvider) DeletePostgres(ctx context.Context, r *v1alpha1.Postgres) (v1alpha1.StatusMessage, error) {
+	// resolve postgres information for postgres created by provider
+	postgresCreateConfig, postgresDeleteConfig, stratCfg, err := p.getPostgresConfig(ctx, r)
+	if err != nil {
+		msg := "failed to retrieve aws postgres config"
+		return v1alpha1.StatusMessage(msg), err
```
aidenkeating: If we aren't providing `msg` in two places we can just use it inline

pkg/providers/aws/provider_postgres.go
```diff
@@ -0,0 +1,487 @@
...
+		resources.RemoveFinalizer(&pg.ObjectMeta, DefaultFinalizer)
+		if err := p.Client.Update(ctx, pg); err != nil {
+			msg := "failed to update instance as part of finalizer reconcile"
+			return v1alpha1.StatusMessage(msg), errorUtil.Wrapf(err, msg)
+		}
+		return "", nil
+	}
+
+	// check if instance db exists and is available
+	if *foundInstance.DBInstanceStatus == "available" {
```
aidenkeating: We can inverse this to keep the whole block of code much flatter

pkg/providers/aws/provider_postgres.go
```diff
@@ -0,0 +1,487 @@
...
+		if *foundInstance.DeletionProtection {
```
aidenkeating: Why do we dereference here?

pkg/providers/aws/provider_postgres.go
```diff
@@ -0,0 +1,487 @@
...
+			_, err = instanceSvc.ModifyDBInstance(&rds.ModifyDBInstanceInput{
+				DBInstanceIdentifier: postgresDeleteConfig.DBInstanceIdentifier,
+				DeletionProtection:   aws.Bool(false),
+			})
+			if err != nil {
+				msg := "waiting on AWS ModifyDBInstance permission"
```
aidenkeating: We're not necessarily waiting on that permission here. Something more general like `failed to remove deletion protection` may be a better approach to this. The operator logs will reveal more to anyone who needs to know

pkg/providers/aws/provider_postgres.go
```diff
@@ -0,0 +1,487 @@
...
+		if err != nil && !isAwsErr {
+			msg := fmt.Sprintf("failed to delete rds instance : %s", err)
+			return v1alpha1.StatusMessage(msg), errorUtil.Wrapf(err, msg)
+		}
+		if err != nil && isAwsErr {
+			if postgresErr.Code() != rds.ErrCodeDBInstanceNotFoundFault {
+				msg := fmt.Sprintf("failed to delete rds instance : %s", err)
+				return v1alpha1.StatusMessage(msg), errorUtil.Wrapf(err, msg)
+			}
+		}
```
aidenkeating: Maybe, something like (didn't test):

```
_, err = instanceSvc.DeleteDBInstance(postgresDeleteConfig)
postgresErr, isAwsErr := err.(awserr.Error)
if err != nil && (!isAwsErr || postgresErr.Code() != rds.ErrCodeDBInstanceNotFoundFault) {
    msg := fmt.Sprintf("failed to delete rds instance : %s", err)
    return v1alpha1.StatusMessage(msg), errorUtil.Wrapf(err, msg)
}
```

pkg/providers/aws/provider_postgres.go
```diff
@@ -0,0 +1,487 @@
...
+
+	postgresDeleteConfig := &rds.DeleteDBInstanceInput{}
+	if err := json.Unmarshal(stratCfg.DeleteStrategy, postgresDeleteConfig); err != nil {
+		return nil, nil, nil, errorUtil.Wrap(err, "failed to unmarshal aws postgres cluster configuration")
+	}
+	return postgresCreateConfig, postgresDeleteConfig, stratCfg, nil
+}
+
+// verifies if there is a change between a found instance and the configuration from the instance strat
+func verifyPostgresChange(postgresConfig *rds.CreateDBInstanceInput, foundConfig *rds.DBInstance) *rds.ModifyDBInstanceInput {
```
aidenkeating: Maybe `buildPostgresUpdateStrategy`? This would follow the same convention as `buildPostgresCreateStrategy` (and `buildPostgresDeleteStrategy` and infer that we're changing stuff rather than checking stuff

pkg/resources/finalizers.go
```diff
@@ -18,7 +18,7 @@ func RemoveFinalizer(om *controllerruntime.ObjectMeta, finalizer string) {
 	om.SetFinalizers(remove(om.GetFinalizers(), finalizer))
 }
 
-func contains(list []string, s string) bool {
+func Contains(list []string, s string) bool {
```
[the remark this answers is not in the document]
aidenkeating: Ah, we use it elsewhere now

ciaranRoche:
@aidenkeating syncing on this in the morning sounds good. The majority of changes you suggested will need to be back ported to some of the other providers.

pkg/providers/aws/provider_postgres.go
```diff
@@ -0,0 +1,487 @@
...
+)
+
+const (
+	postgresProviderName       = "aws-rds"
+	defaultAwsIdentifierLength = 40
+	// default create options
+	defaultAwsPostgresDeletionProtection = true
+	defaultAwsPostgresPort               = 5432
+	defaultAwsPostgresUser               = "postgres"
+	defaultAwsAllocatedStorage           = 20
```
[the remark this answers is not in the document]
ciaranRoche: According to the [docs](https://docs.aws.amazon.com/sdk-for-go/api/service/rds/#CreateDBInstanceInput) it will not scale without the max allocated storage set. 
Aurora has the feature for auto scaling without setting max allocated storage.

Could this be a candidate to bring out of the strat and move to the spec?

pkg/providers/aws/provider_postgres.go
ciaranRoche: Yeah we need to dereference these as they are `aws` variables, they use pointers.

pkg/providers/aws/provider_postgres.go
ciaranRoche: Yeah we need to dereference these as they are aws variables, they use pointers.

austincunningham:
@ciaranRoche and @dimitraz 
Anything beyond the following to verifiy
## Verify
- Clone this branch
- Run make `cluster/clean`
- Run make `cluster/prepare`
- Run make `cluster/seed/postgres`
- Run make `run`

ciaranRoche:
@austincunningham just verify an RDS instance is created in AWS, then remove the postgres CR and verify the RDS instance gets deleted, and both secrets are removed from the namespace

austincunningham:
Verified Created
- [x] RDS instance created on AWS
- [x] Secrets created in the namespace
Verified Deleted when delete cr
- [x] RDS instance removed on AWS
- [x] Secrets deleted in the namespace

PR #20, opened 2019-10-09
OLM Manifest

aidenkeating:
Add Travis spec for building and pushing operator images and manifests.

verification:
- create:

```
apiVersion: operators.coreos.com/v1
kind: OperatorSource
metadata:
  name: integreatly-operators
spec:
  authorizationToken: {}
  displayName: Integreatly Operators
  endpoint: 'https://quay.io/cnr'
  publisher: Integreatly Publisher
  registryNamespace: aidenkeating
  type: appregistry
```

- create the operator from the catalog
- test it out


pkg/resources/provider.go
```diff
@@ -0,0 +1,69 @@
...
+		}
+		e.Data = d
+		e.Type = v1.SecretTypeOpaque
+		return nil
+	})
+	if err != nil {
+		if updateErr := UpdatePhase(ctx, r.Client, o, v1alpha1.PhaseFailed, "failed to reconcile instance secret"); updateErr != nil {
+			return updateErr
+		}
+		return errors.Wrapf(err, "failed to reconcile smtp credential set instance secret %s", sec.Name)
```
dimitraz: Can be generic/deduped

PR #26, opened 2019-10-18
replace underscores in config names

dimitraz:


aidenkeating:
What's the cause for these showing up in the first place? Is it in the hash at the end?

ciaranRoche:
@aidenkeating testing on a cluster, we ended up with `_` in the hash. Could of been just a snowflake, but we added the function just in case

PR #27, opened 2019-10-28
fix in cluster postgres provisioning

aidenkeating:
currently we do not give the correct permissions to the user we
create in the postgres instance for openshift deployments. in rhmi
codeready workspaces requires the user to be a superuser.

this defaults the created user to be a superuser.

verification:
- have an openshift 4 cluster running
- have this operator running
- have the rhmi operator running from austincunningham:INTLY-3146
- create an installation cr

```
kind: Installation
metadata:
  name: example-installation
spec:
  type: workshop
  namespacePrefix: integreatly-
  routingSubdomain: apps.example.com
  masterUrl: http://console.apps.example.com
  selfSignedCerts: true
  useExternalResources: true
```

- ensure codeready comes up
- ensure codeready has no self-deployed database
- ensure codeready migrations complete successfully


deploy/olm-catalog/cloud-resources/cloud-resource-operator.clusterserviceversion.yaml
```diff
@@ -115,7 +115,7 @@ spec:
                       fieldPath: metadata.name
                 - name: OPERATOR_NAME
                   value: cloud-resource-operator
-                image: quay.io/integreatly/cloud-resource-operator:0.1.0
+                image: quay.io/aidenkeating/cloud-resource-operator:0.1.0
```
ciaranRoche: Do we not want to be pointing to the intly Quay?

PR #28, opened 2019-10-29
[WIP] INTLY-3773 - e2e testing

ciaranRoche:
## Overview

Jira: https://issues.jboss.org/browse/INTLY-3773

### Prerequisite
- oc logged into a cluster
- operator-sdk version: v0.10.0

### Steps
- Run `make image/build`
- Run `make image/push IMAGE_ORG=<your quay>`
- Update `deploy/operator.yaml` `image` to point to your image
- Run `make test/e2e`

### DIFF
--- a/test/e2e/cro_test.go
+++ b/test/e2e/cro_test.go
@@ -1,0 +1,88 @@
+package e2e
+
+import (
+	"testing"
+	"time"
+
+	"github.com/integr8ly/cloud-resource-operator/pkg/apis"
+	"github.com/integr8ly/cloud-resource-operator/pkg/apis/integreatly/v1alpha1"
+	"github.com/operator-framework/operator-sdk/pkg/test/e2eutil"
+
+	framework "github.com/operator-framework/operator-sdk/pkg/test"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var (
+	retryInterval = time.Second * 5
+	timeout       = time.Second * 30
+)
+
+func TestCRO(t *testing.T) {
+	redisList := &v1alpha1.Redis{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Redis",
+			APIVersion: "integreatly.org/v1alpha1",
+		},
+	}
+	if err := framework.AddToFrameworkScheme(apis.AddToScheme, redisList); err != nil {
+		t.Fatalf("failed to add Redis custom resource scheme to framework: %v", err)
+	}
+	postgresList := &v1alpha1.Postgres{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Postgres",
+			APIVersion: "integreatly.org/v1alpha1",
+		},
+	}
+	if err := framework.AddToFrameworkScheme(apis.AddToScheme, postgresList); err != nil {
+		t.Fatalf("failed to add Postgres custom resource scheme to framework: %v", err)
+	}
+	blobstorageList := &v1alpha1.BlobStorage{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "BlobStorage",
+			APIVersion: "integreatly.org/v1alpha1",
+		},
+	}
+	if err := framework.AddToFrameworkScheme(apis.AddToScheme, blobstorageList); err != nil {
+		t.Fatalf("failed to add Blob Storage custom resource scheme to framework: %v", err)
+	}
+	smtpList := &v1alpha1.SMTPCredentialSet{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "SMTPCredentialSet",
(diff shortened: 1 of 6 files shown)


test/e2e/cro_test.go
```diff
@@ -0,0 +1,149 @@
...
+			Type: "workshop",
+		},
+	}
+	// create postgres resource
+	if err := f.Client.Create(goctx.TODO(), examplePostgres, getCleanupOptions(t)); err != nil {
+		return errorUtil.Wrapf(err, "could not create example Postgres")
+	}
+
+	// wait from postgres deployment
+	if err := e2eutil.WaitForDeployment(t, f.KubeClient, namespace, "example-postgres", 1, retryInterval, timeout); err != nil {
```
aidenkeating: Do we need to wait for this at all anymore?

test/e2e/cro_test.go
```diff
@@ -0,0 +1,149 @@
...
+		return errorUtil.Wrapf(err, "could not get deployment")
+	}
+
+	// get created secret
+	sec := v1.Secret{}
+	err = wait.Poll(retryInterval, timeout, func() (done bool, err error) {
```
aidenkeating: Should we poll on the postgres custom resource here instead? We expect that once the status phase is complete that the secret should exist.

The e2e test could test this by polling the custom resource, waiting for the status to complete and then trying to get the secret from the secret ref in the status

test/e2e/cro_test.go
```diff
@@ -0,0 +1,149 @@
...
+		if sec.Data[k] == nil {
+			return errorUtil.New(fmt.Sprintf("secret %v value not found", k))
+		}
+	}
+
+	// delete postgres resource
+	if err := f.Client.Delete(goctx.TODO(), examplePostgres); err != nil {
+		return errorUtil.Wrapf(err, "failed to delete example Postgres")
+	}
+
```
aidenkeating: Can we perform a check that at least the deployment, secret and pvc no longer exist?

test/e2e/cro_test.go
ciaranRoche: We don't, but I think for logging it might be better to catch the error here, if there is a problem with the deployment, over a timeout on the secret. What do you think?

test/e2e/cro_test.go
ciaranRoche: Ok, that makes sense now, polling the CR, we can then remove the poll on the deployment, and return the status if there is a timeout. Ill make that change.

test/e2e/cro_test.go
ciaranRoche: Yeah, makes sense, good catch :+1:

test/e2e/cro_test.go
```diff
@@ -0,0 +1,190 @@
...
+	// wait from postgres deployment
+	if err := e2eutil.WaitForDeployment(t, f.KubeClient, namespace, postgresName, 1, retryInterval, timeout); err != nil {
+		return errorUtil.Wrapf(err, "could not get deployment")
+	}
+	t.Logf("%s deployment created", postgresName)
+
+	// poll cr for complete status phase
+	pcr := &v1alpha1.Postgres{}
+	err = wait.Poll(retryInterval, time.Minute*6, func() (done bool, err error) {
+		if err := f.Client.Get(goctx.TODO(), types.NamespacedName{Namespace: namespace, Name: postgresName}, pcr); err != nil {
```
aidenkeating: could we add a log statement in here letting the user know we're polling for complete and the current status phase is, just so we know what's happening when looking at this from a ci perspective?

test/e2e/cro_test.go
ciaranRoche: Going to leave this poll for now, till we sort the reconcile time

test/e2e/cro_test.go
```diff
@@ -0,0 +1,190 @@
...
+		return false, nil
+	})
+	if err != nil {
+		return err
+	}
+	t.Logf("postgres status phase %s", pcr.Status.Phase)
+
+	// get created secret
+	sec := v1.Secret{}
+	if err := f.Client.Get(goctx.TODO(), types.NamespacedName{Namespace: namespace, Name: "example-postgres-sec"}, &sec); err != nil {
```
aidenkeating: we won't get the full value of this test until we're polling the cr, as we can't reference the `secretRef` yet. let's hold off on merging this until we can update this to check the CR

test/e2e/cro_test.go
ciaranRoche: I had it there, but the logs, are only printed after the test is run, so it added noise to the final output. 
Ill have a look to see if i can get them printed out in real time, and update.

aidenkeating: ah that's sucky, but good to know for people when they're writing tests.

it makes sense to keep it clean then.

PR #29, opened 2019-10-29
add colours to make test/unit

austincunningham:
## Overview
add colours to `make test/unit`

## Verification

- Clone this branch
- run `make test/unit`
should see something like 
image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,6 +36,7 @@ jobs:
     - stage: test
       script:
         - go get github.com/mattn/goveralls
+        - make test/unit/setup
         - make code/check
         - make test/unit/ci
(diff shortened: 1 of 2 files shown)


aidenkeating:
This seems fine, can we add a `make test/unit/setup` target that runs `go get -u github.com/rakyll/gotest`? Not everyone will have the tool by default

austincunningham:
@aidenkeating  it was added here https://github.com/integr8ly/cloud-resource-operator/blob/5887eb9ee6b9d0253a73af762228504bc29ea07a/tools.go#L6 previously but I noticed Travis isn't picking it up so might be an option.

PR #31, opened 2019-10-30
remove namespace from postgres cr and set the finilizers to the same name

austincunningham:
## Overview
bug fix, remove namespace from postgres cr and set the finilizers to the same name

### Setup
- Clone this branch
- Change the name space in the Makefile `integreatly-operator`
- Deploy `managed` and `workshop` by changing the postgres cr
- Run `make cluster/prepare`
- Run `make cluster/seed/postgres`
- Run `make run`
### Verify
- make sure that the postgres cr points at `integreatly-operator`
- make sure that when the postgres cr is deleted all resources are removed .

### DIFF
--- a/deploy/crds/integreatly_v1alpha1_postgres_cr.yaml
+++ b/deploy/crds/integreatly_v1alpha1_postgres_cr.yaml
@@ -7,7 +7,6 @@ spec:
   # i want my postgres storage information output in a Secret named test in the Namespace cloud-resource-operator
   secretRef:
     name: example-postgres-sec
-    namespace: cloud-resource-operator
   # i want a postgres storage of a development-level tier
   tier: development
   # i want a postgres storage for the type managed
--- a/pkg/providers/aws/config.go
+++ b/pkg/providers/aws/config.go
@@ -22,7 +22,7 @@ const (
 	DefaultConfigMapName      = "cloud-resources-aws-strategies"
 	DefaultConfigMapNamespace = "kube-system"
 
-	DefaultFinalizer = "finalizers.aws.cloud-resources-operator.integreatly.org"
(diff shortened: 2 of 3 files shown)


dimitraz:
I attempted to verify this and it seems to be ok, just a few things related to my testing environment.. 

- My cluster name seems to cause problems for the `buildInfraNameFromObject` hash, so the RDS name is usually invalid. Even if the actual RDS instance doesn't come up, the `Postgres` object is still deleting without issues. I created a JIRA for that here: https://issues.jboss.org/browse/INTLY-3963
- When I try to provision an in-cluster postgres (and any other in-cluster resources), I'm running in to pod networking issues (even though my nodes are healthy). Because of this, the pod gets stuck in `container creating` mode, and as a result the deletion also gets stuck, so I can't verify that the delete actually works for in-cluster postgres resources.. I think we can assume it does, but we probably should address the container health issue, which reminds me of [this issue](https://github.com/integr8ly/cloud-resource-operator/blob/master/pkg/providers/openshift/provider_redis.go#L137) in openshift resource deletion functions where we only delete resources if the deployment is healthy, and if it isn't, the deletion never completes :thinking: Sorry, that was a long sentence

aidenkeating:
@dimitraz Thanks for creating that issue. Do you have the name of the resource you created so that once fixed we can include it in unit tests for regression testing purposes?

For the second issue, I've created https://issues.jboss.org/browse/INTLY-3975 to address this. It made sense to do this approach in AWS, but in OpenShift, we probably don't need to wait for a successful state before deletion.

@dimitraz @austincunningham I don't think this blocks the purpose of the current PR though. Let's merge this and address the two issues in another PR.

austincunningham:
@aidenkeating agree

PR #33, opened 2019-10-31
Change from kube-system namespace by default

austincunningham:
## Overview

Jira: https://issues.jboss.org/browse/INTLY-3896

## Verification

- Clone this branch
- Run `make cluster/prepare`
- Run `make cluster/seed/postgres`
- Run `make run`
- Ensure that cro configmap's are created in the `cloud-resources-operator` namespace
- Change the `NAMESPACE=` var in the makefile
- Run `make cluster/prepare`
- Run `make cluster/seed/postgres`
- Run `make run`
- Ensure that cro configmap's are created in the new namespace
- Ensure no cro resources are created under the `kube-system` namespace

### DIFF
--- a/deploy/examples/cloud_resource_config.yaml
+++ b/deploy/examples/cloud_resource_config.yaml
@@ -2,7 +2,6 @@ kind: ConfigMap
 apiVersion: v1
 metadata:
   name: cloud-resource-config
-  namespace: kube-system
 data:
   managed: |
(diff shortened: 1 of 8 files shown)


aidenkeating:
@austincunningham i think there may be something missing in this pr. we want to ensure that the operator will check in the namespace in which the operator is running instead of something static like `cloud-resource-operator`. so basically we want to override the default namespace with the namespace of the custom resource.

let's sync on this tomorrow. the ticket may not have been described very well.

PR #34, opened 2019-10-31
add provider specified reconcile timeouts

aidenkeating:
currently we have a global override for the reconcile timings in
each resource type. however this should be more individual instead
of allowing a controller to decide what the timings should be.

this change performs a slight refactor to the current reconcile
time override function and also ensures each provider can return
it's own reconcile time based on knowledge of the cr state.

### DIFF
--- a/pkg/controller/smtpcredentialset/smtpcredentialset_controller_test.go
+++ b/pkg/controller/smtpcredentialset/smtpcredentialset_controller_test.go
@@ -107,6 +107,9 @@ func TestReconcileSMTPCredentialSet_Reconcile(t *testing.T) {
 								},
 							}, "", nil
 						},
+						GetReconcileTimeFunc: func(smtpCreds *v1alpha1.SMTPCredentialSet) time.Duration {
+							return time.Second * 10
+						},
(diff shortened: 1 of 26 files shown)


pkg/controller/smtpcredentialset/smtpcredentialset_controller_test.go
```diff
@@ -107,6 +107,9 @@ func TestReconcileSMTPCredentialSet_Reconcile(t *testing.T) {
 								},
 							}, "", nil
 						},
+						GetReconcileTimeFunc: func(smtpCreds *v1alpha1.SMTPCredentialSet) time.Duration {
+							return time.Second * 10
```
austincunningham: Much as I like the readability of `* 10` I think we should add the numbers as variables

aidenkeating: good point, will do

aidenkeating: going to address this and other refactoring in a separate pr. merging to unblock e2e tests

PR #35, opened 2019-11-03
ensure unsupported provided and tiers are included in status block

aidenkeating:
currently, when there is not a provider for the type of the cr a
message is returned.

this change ensures the message returned is a standard.

currently, when an error occurs in a provider, only an overview
message is returned to the status block of the cr.

this change adds a WrapError function to act on the StatusMessage
struct that includes the returned string from the Error function
of an underlying error. This makes the status block slightly more
verbose in error situations.

currently, it is possible that the string returned from the util
string shortening function, ShortenString, is not url or aws
resource name-safe. this is due to a number of reasons that allow
for characters such as underscores due to base64 encoding or double
hyphens due to the way there was no preprocessing on the provided
string.

this change swaps to base32 encoding which does not include any
special characters and ensures all non-alphanumeric characters are
removed from the original provided string before processing the
string.


aidenkeating:
@ciaranRoche i think this may cover at least some of the verbose logging work

@dimitraz this should resolve the issues with resource naming you were seeing in aws

i'll add some verification steps to this tomorrow morning. if you wouldn't mind taking a look, that'd be great.

there's a few deduping tasks that can be created, i'll do that tomorrow morning.

PR #40, opened 2019-11-05
generate default postgres password, add cluster/seed/all

austincunningham:
## Overview

Jira: https://issues.jboss.org/browse/INTLY-3759

## Verification

- Clone this branch
- Run `make cluster/prepare`
- Run `make run`
- Run `make cluster/seed/all`
- Ensure all generated secrets have a random password

### DIFF
--- a/pkg/providers/openshift/provider_postgres.go
+++ b/pkg/providers/openshift/provider_postgres.go
@@ -35,7 +35,7 @@ var (
 	// default openshift create paramaters
 	defaultPostgresPort        = 5432
 	defaultPostgresUser        = "postgresuser"
-	defaultPostgresPassword    = "password"
+	defaultPostgresPassword, _ = resources.GeneratePassword()
 	defaultPostgresUserKey     = "user"
 	defaultPostgresPasswordKey = "password"
 	defaultPostgresDatabaseKey = "database"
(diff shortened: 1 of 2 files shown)


pkg/providers/openshift/provider_postgres.go
```diff
@@ -35,7 +35,7 @@ var (
 	// default openshift create paramaters
 	defaultPostgresPort        = 5432
 	defaultPostgresUser        = "postgresuser"
-	defaultPostgresPassword    = "password"
+	defaultPostgresPassword, _ = resources.GeneratePassword()
```
ciaranRoche: Do we need this here? We generate the password and handle the error when creating the [secret](https://github.com/integr8ly/cloud-resource-operator/blob/master/pkg/providers/openshift/provider_postgres.go#L108). We only use this `defaultPostgresPassword` if the `password` is nil after we generate it.

austincunningham: Only spotted it here https://github.com/integr8ly/cloud-resource-operator/blob/6c7aaf3c5c6dd75fe33d913a59a7763c5a848e90/pkg/providers/openshift/provider_postgres.go#L505 but I see what you mean https://github.com/integr8ly/cloud-resource-operator/blob/6c7aaf3c5c6dd75fe33d913a59a7763c5a848e90/pkg/providers/openshift/provider_postgres.go#L108

Agree, will close this pr as no value.

PR #55, opened 2019-11-19
update vpc peering docs for security groups

aidenkeating:
ensure the vpc peering steps mentions the security group updates required


aidenkeating:
@dimitraz @austincunningham mind taking a quick look?

PR #57, opened 2019-11-21
Updating operator-sdk to v0.12.0

davidffrench:
## Overview
https://issues.jboss.org/browse/INTLY-4216

Unfortunately there is one potential issue in this PR that I haven't been able to get resolved. The CRDs now have a `$ref` field for the spec and status instaed of the full object tree. I have tried everything I know about the generation to try and get this working again but it needs someone with a better understanding of the k8s code-generator (https://github.com/kubernetes/code-generator).

## Verification

- Clone this branch
- Run `make cluster/prepare`
- Run `make run`
- Ensure everything works :-)


aidenkeating:
@davidffrench Thanks for putting the work in on this!

Do you see the need for vendoring dependencies as a necessity here? It has become a non-default option in go1.13

davidffrench:
@aidenkeating @dimitraz Thanks for the review. I was not aware vendoring is now the non default in 1.13. Happy to remove the vendor folder if this is what you would like?

aidenkeating:
@davidffrench Let's remove it for now, due to Go Modules now being widely available in go1.13, we can probably leverage it's defaults unless we have a reason to change.

If we have the need to provide vendored dependencies in the future, we can add it in a separate PR.

davidffrench:
done now @aidenkeating .

Makefile
```diff
@@ -40,10 +40,10 @@ code/audit:
...
 cluster/prepare:
 	oc new-project $(NAMESPACE) || true
-	oc apply -f ./deploy/crds/integreatly_v1alpha1_blobstorage_crd.yaml
-	oc apply -f ./deploy/crds/integreatly_v1alpha1_smtpcredentialset_crd.yaml
-	oc apply -f ./deploy/crds/integreatly_v1alpha1_redis_crd.yaml
-	oc apply -f ./deploy/crds/integreatly_v1alpha1_postgres_crd.yaml
+	oc apply -f ./deploy/crds/integreatly.org_blobstorages_crd.yaml
+	oc apply -f ./deploy/crds/integreatly.org_smtpcredentialsets_crd.yaml
+	oc apply -f ./deploy/crds/integreatly.org_redis_crd.yaml
+	oc apply -f ./deploy/crds/integreatly.or__postgres_crd.yaml
```
ciaranRoche: Small typo here. `oc apply -f ./deploy/crds/integreatly.org_postgres_crd.yaml`

deploy/crds/integreatly.org_blobstorages_crd.yaml
```diff
@@ -28,41 +29,12 @@ spec:
...
-            tier:
-              type: string
-            type:
-              type: string
-          required:
-          - type
-          - tier
-          - secretRef
-          type: object
+          $ref: '#/definitions/github.com~1integr8ly~1cloud-resource-operator~1pkg~1apis~1integreatly~1v1alpha1~1types~0ResourceTypeSpec'
```
ciaranRoche: Should we be defining `$ref` in `types`? I am not sure here so I am just asking :smile_cat:

davidffrench: no, unfortunately this does allow the CRD to be applied to a cluster. It is because all the types are using the format `type SMTPCredentialSetSpec types.ResourceTypeSpec`

Makefile
davidffrench: good spotting, thanks @ciaranRoche

deploy/crds/integreatly.org_blobstorages_crd.yaml
davidffrench: Fixed now

ciaranRoche:
/test e2e 

Just looks like a timeout on prows side

davidffrench:
Thanks @ciaranRoche

PR #58, opened 2019-11-22
feat: Add monitoring label to operator service monitor

davidffrench:
## Overview
None

Adding monitoring-key:middleware to the service monitor created by the operator itself so the metrics can be fed into the middleware monitoring prometheus.

## Verification

- Checkout this branch
- Run `make cluster/prepare`
- Modify `operator.yaml` image to  `quay.io/davidffrench/cloud-resource-operator:add_monitoring_label`
- Run `kubectl deploy -f deploy/operator.yaml`
- Check the service-monitor created by the operator in the same namespace has the correct `monitoring-key:middleware` label

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,12 +49,16 @@ code/audit:
 
 .PHONY: cluster/prepare
 cluster/prepare:
-	oc new-project $(NAMESPACE) || true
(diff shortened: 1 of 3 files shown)


davidffrench:
@david-martin FYI, PR to add the label to the operator service monitor

PR #72, opened 2019-12-06
add more details to the vpc peering documentation

JameelB:
## Overview

Jira: N/A

The following changes adds more details to the current VPC peering documentation located in the README.

## Verification

- Follow the process stated in the documentation and ensure that VPC peering works successefully

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,11 +59,34 @@ $ make cluster/clean
 
 ## VPC Peering 
 Currently AWS resources are deployed into a separate Virtual Private Cloud (VPC) than the VPC that the cluster is deployed into. In order for these to communicate, a `peering connection` must be established between the two VPCS. To do this:
-- In the AWS VPC console, create a new peering connection between the two VPCs. This is a two-way communication channel, so only one needs to be created
-- Select the newly created connection, then click `actions > accept request` to accept the peering request
-- Edit the cluster VPC route table. Create a new route that contains the resource/default VPC's CIDR block as the `destination` and the newly created peering connection as the `target`
-- Edit the resource VPC's route table. Create a new route that contains the CIDR block of the cluster VPC as the `destination` and the peering connection as the `target`. 
-- Edit the Security Groups associated with each VPC to ensure database and cache traffic can pass between the two VPCs.
+1. Create a new peering connection between the two VPCs.
+  - Go to `VPC` > `Peering Connections`
+  - Click on `Create Peering Connection`
+    - **NOTE**: This is a two-way communication channel so only one needs to be created.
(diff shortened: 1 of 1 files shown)


JameelB:
/cc @dimitraz @aidenkeating 

Added some more details to the vpc peering doc, please let me know if the steps i've added in are correct and if i miss anything else.

README.md
```diff
@@ -59,11 +59,33 @@ $ make cluster/clean
...
-- Edit the Security Groups associated with each VPC to ensure database and cache traffic can pass between the two VPCs.
+1. Create a new peering connection between the two VPCs.
+  - Go to `VPC` > `Peering Connections`
+  - Click on `Create Peering Connection`
+    - **NOTE**: This is a two-way communication channel so only one needs to be created.
+  - Select the newly created connection, then click `Actions` > `Accept Request` to accept the peering request.
+2. Edit the cluster VPC route table
+  - Go to `VPC` > `Your VPCs`
+  - Find the cluster VPC and click on its route table under the heading `Main Route Table`
+    - The cluster VPC usually has an empty name or is called `default`
```
dimitraz: The cluster VPC is actually the one with your name, and the resources VPC is usually the default one because resources get provisioned there by default

README.md
```diff
@@ -59,11 +59,33 @@ $ make cluster/clean
...
+  - Go to `VPC` > `Your VPCs`
+  - Find your VPC and click on its route table under the heading `Main Route Table`
+  - Select the cluster route table and click on `Action` > `Edit routes`
+  - Add a new route with the following details:
+    ```
+    Destination: <cluster VPC CIDR block>
+    Target: <newly created peering connection>
+    ```
+  - Click on `Save routes`
+4. Edit the Security Groups associated with each VPC to ensure database and cache traffic can pass between the two VPCs.
```
dimitraz: Should we update this to say modify the resources VPC? Since in the end I don't think we needed to update the cluster VPC security groups, as the access would be from cluster VPC --> resource VPC

README.md
JameelB: if this is actually your VPC, is the destination/target correct? I remember setting the destination/target to point to these for the `default` VPC.

JameelB: Ah ignore the comment above, i got confused which CIDR block was which :) Think this is right :+1:

README.md
JameelB: When you say `modify the resources VPC` what exactly do you mean? We didn't update the security groups yesterday because this was already updated by @aidenkeating? However, it was still needed to be updated to allow traffic to pass between the two VPCs on the initial setup right?

dimitraz: Sorry, I meant I think we only need to update the security security groups associated with the resources VPC, and not the cluster VPC. That says to update both

PR #74, opened 2019-12-06
Intly 3968 all tagging

austincunningham:
## Overview

Jira: https://issues.jboss.org/browse/INTLY-3968

## Verification

- Clone this branch
- Run `make cluster/prepare`
- Add the cr's to the cloud-resources-operator namespace for postgres,redis and blobstorage of type managed
e.g.
```yaml
kind: Postgres
metadata:
  # name must be between 1-40 characters
  name: example-postgres
  labels:
    # label for the product we are installing , subject to change
    productname: productName
spec:
  # i want my postgres storage information output in a secret named example-postgres-sec
  secretRef:
    name: example-postgres-sec
  # i want a postgres storage of a development-level tier
  tier: development
  # i want a postgres storage for the type workshop
  type: managed
```
- to run locally
```bash
export DEFAULT_ORGANIZATION_TAG=integreatly.org/
```

- Run `make run`
- Ensure Tags are added to the instances
- Ensure Tags are added to the snapshots for your postgres DB instance


pkg/providers/aws/provider_blobstorage.go
```diff
@@ -156,6 +156,49 @@ func (p *BlobStorageProvider) CreateStorage(ctx context.Context, bs *v1alpha1.Bl
...
 	}
 
+	// Adding tags to s3
+	p.Logger.Infof("bucket %s found, Adding tags to bucket", *bucketCreateCfg.Bucket)
+
+	// get the environment from the CR
+	defaultOrganizationTag := bs.Spec.Env.Value
+
+	// new session but don't need credentials
+	s3svctag := s3.New(session.Must(session.NewSession(&aws.Config{Region: aws.String(stratCfg.Region)})))
```
ciaranRoche: We have a helper func for creating a session `createS3Session`

pkg/providers/aws/provider_blobstorage.go
```diff
@@ -156,6 +156,49 @@ func (p *BlobStorageProvider) CreateStorage(ctx context.Context, bs *v1alpha1.Bl
 		},
 	}
 
+	// Adding tags to s3
```
ciaranRoche: Would it be worth, abstracting this logic out into its own `reconcileBucketTagging` function. IMO it would read better and follow the same pattern as reconciling the bucket creation

pkg/providers/aws/provider_postgres.go
```diff
@@ -203,6 +202,77 @@ func (p *AWSPostgresProvider) createRDSInstance(ctx context.Context, cr *v1alpha
...
+	// get the environment from the CR
+	defaultOrganizationTag := cr.Spec.Env.Value
+
+	// new session but don't need credentials
+	// Get the region from AvailabilityZone
+	rdsRegion := *foundInstance.AvailabilityZone
+	regionLen := len(rdsRegion)
+	rdsRegion = rdsRegion[:regionLen-1]
+	// get provider aws creds so the postgres instance can be deleted
+	rdsSvc = rds.New(session.Must(session.NewSession(&aws.Config{
```
ciaranRoche: Why does this session not need credentials? this makes me think we can reuse the already created session.

pkg/providers/aws/provider_postgres.go
```diff
@@ -203,6 +202,77 @@ func (p *AWSPostgresProvider) createRDSInstance(ctx context.Context, cr *v1alpha
 		}
 		return nil, "modify instance in progress", nil
 	}
+	// Add Tags to Aws Postgres resources
```
ciaranRoche: I feel this function is getting quite long. Maybe we should look at abstracting the tagging and snapshot logic to its own function.

pkg/providers/aws/provider_postgres.go
```diff
@@ -203,6 +202,77 @@ func (p *AWSPostgresProvider) createRDSInstance(ctx context.Context, cr *v1alpha
...
+	logrus.Infof("Adding Tags to RDS instance %s", *foundInstance.DBInstanceIdentifier)
+	// get the environment from the CR
+	defaultOrganizationTag := cr.Spec.Env.Value
+
+	// new session but don't need credentials
+	// Get the region from AvailabilityZone
+	rdsRegion := *foundInstance.AvailabilityZone
+	regionLen := len(rdsRegion)
```
ciaranRoche: Does this need to be a variable?

pkg/providers/aws/provider_postgres.go
```diff
@@ -203,6 +202,77 @@ func (p *AWSPostgresProvider) createRDSInstance(ctx context.Context, cr *v1alpha
...
+	inputRdsInstance := &rds.AddTagsToResourceInput{
+		ResourceName: aws.String(*foundInstance.DBInstanceArn),
+		Tags:         rdsTag,
+	}
+
+	// adding tags to rds postgres instance
+	_, err = rdsSvc.AddTagsToResource(inputRdsInstance)
+	if err != nil {
+		msg := "Failed to add Tags to RDS instance"
+		logrus.Errorf("Error : %s : %s", msg, err)
```
ciaranRoche: Why are we only logging this error?

pkg/providers/aws/provider_postgres.go
```diff
@@ -203,6 +202,77 @@ func (p *AWSPostgresProvider) createRDSInstance(ctx context.Context, cr *v1alpha
...
+	}
+
+	// Get a list of Snapshot objects for the DB instance
+	rdsSnapshotAttributeInput := &rds.DescribeDBSnapshotsInput{
+		DBInstanceIdentifier: aws.String(*foundInstance.DBInstanceIdentifier),
+	}
+	rdsSnapshotList, err := rdsSvc.DescribeDBSnapshots(rdsSnapshotAttributeInput)
+	if err != nil {
+		logrus.Infof("Can't get Snapshot info : %s", err)
```
ciaranRoche: Same as above, should we be returning this error?

pkg/providers/aws/provider_postgres.go
```diff
@@ -203,6 +202,77 @@ func (p *AWSPostgresProvider) createRDSInstance(ctx context.Context, cr *v1alpha
...
+	for _, snapshotList := range rdsSnapshotList.DBSnapshots {
+		inputRdsSnapshot := &rds.AddTagsToResourceInput{
+			ResourceName: aws.String(*snapshotList.DBSnapshotArn),
+			Tags:         rdsTag,
+		}
+		// Adding Tags to RDS Snapshot
+		_, err = rdsSvc.AddTagsToResource(inputRdsSnapshot)
+		if err != nil {
+			msg := "Failed to add Tags to RDS Snapshot"
+			logrus.Errorf("Error : %s : %s", msg, err)
```
ciaranRoche: Same as above.

pkg/providers/aws/provider_redis.go
```diff
@@ -158,6 +159,66 @@ func (p *AWSRedisProvider) createElasticacheCluster(ctx context.Context, r *v1al
...
+		}
+		input := &elasticache.AddTagsToResourceInput{
+			ResourceName: aws.String(arn),
+			Tags:         cacheTag,
+		}
+		// add tags
+		_, err = cacheSvcTag.AddTagsToResource(input)
+
+		//TODO when snapshots are created add tags to them
+		//inputDescirbe := &elasticache.DescribeSnapshotsInput{
```
ciaranRoche: Do we need to leave this commented out code?

pkg/providers/aws/provider_redis.go
```diff
@@ -158,6 +159,66 @@ func (p *AWSRedisProvider) createElasticacheCluster(ctx context.Context, r *v1al
 		return nil, "modify elasticache cluster in progress", nil
 	}
 
+	// Add Tags to AWS Elasticache
+	if err != nil {
+		logrus.Errorf("Error Failed to get strategy config: %s", err)
+	}
+	cacheSvcTag := elasticache.New(session.Must(session.NewSession(&aws.Config{Region: aws.String(stratCfg.Region)})))
```
ciaranRoche: Same as others, do we need to create a new session for the tagging?

pkg/providers/openshift/provider_postgres.go
```diff
@@ -443,7 +444,7 @@ func buildDefaultPostgresPodContainers(ps *v1alpha1.Postgres) []v1.Container {
 	return []v1.Container{
 		{
 			Name:  ps.Name,
-			Image: "registry.redhat.io/rhscl/postgresql-96-rhel7",
+			Image: "registry.redhat.io/rhscl/postgresql-10-rhel7",
```
ciaranRoche: Why are we changing this image?

version/version.go
```diff
@@ -1,5 +1,5 @@
 package version
 
 var (
-	Version = "0.2.0"
+	Version = "0.3.0"
```
ciaranRoche: Is this related to this change?

deploy/crds/integreatly_v1alpha1_postgres_cr.yaml
```diff
@@ -11,3 +11,10 @@ spec:
   tier: development
   # i want a postgres storage for the type workshop
   type: workshop
+  labels:
```
ciaranRoche: Comment around how we are doing this. Do we need to add this info to the CR. Can we not generate these dynamically? 
I would lean toward not having them added to the CR, but think maybe we should have more of a discussion around this

pkg/providers/openshift/provider_postgres.go
dimitraz: Looks like a cherry pick from my old branch which should be merged now, just needs a rebase I think

version/version.go
dimitraz: Same thing

pkg/providers/aws/provider_redis.go
ciaranRoche: I just noticed this work is done here https://github.com/integr8ly/cloud-resource-operator/pull/75

Can we look at merging these together, to speed up the verification

pkg/providers/aws/provider_blobstorage.go
```diff
@@ -156,6 +156,49 @@ func (p *BlobStorageProvider) CreateStorage(ctx context.Context, bs *v1alpha1.Bl
...
+	// get the environment from the CR
+	defaultOrganizationTag := bs.Spec.Env.Value
+
+	// new session but don't need credentials
+	s3svctag := s3.New(session.Must(session.NewSession(&aws.Config{Region: aws.String(stratCfg.Region)})))
+
+	//get Cluster Id
+	clusterId, _ := resources.GetClusterId(ctx, p.Client)
+	// Tagging input
+	bucketTaggingInput := &s3.PutBucketTaggingInput{
```
dimitraz: Reading up about `PutBucketTagging`, it seems to override existing tags if there are any. Do we want this? Would it be safer to get a list of tags and merge our tags with that existing list? 

wdyt @austincunningham @ciaranRoche

deploy/crds/integreatly_v1alpha1_postgres_cr.yaml
dimitraz: Would be pretty nice to be able to tag based on flexible list of key/value pairs, that way we aren't adding labels that are very integreatly specific. 

For integration with integreatly we could have a cro util function that accepts a map of key value pairs, and updates the cr dynamically with the labels. If labels exist on the cr we can proceed to tag the aws resources.

pkg/apis/integreatly/v1alpha1/types/types.go
```diff
@@ -18,12 +20,23 @@ type SecretRef struct {
...
+}
+
 // ResourceTypeSpec Represents the basic information required to provision a resource type
 // +k8s:openapi-gen=true
 type ResourceTypeSpec struct {
 	Type      string     `json:"type"`
 	Tier      string     `json:"tier"`
 	SecretRef *SecretRef `json:"secretRef"`
+	Env       *Env       `json:"env"`
+	Labels    *Labels    `json:"labels"`
```
dimitraz: Personally I think these fields should be optional. If the labels are there, we can use them to tag our resources. If the organisation tag is there, we can use that too, or otherwise set a reasonable/empty fallback value.

pkg/providers/openshift/provider_postgres.go
ciaranRoche: Why was the image changed, have the products changed to this version?

deploy/crds/integreatly_v1alpha1_postgres_cr.yaml
ciaranRoche: For `env` I think this could be an `env var`, if set then we label with the organisation. If not we ignore this label. 
For the other label, I know we need to be able to list resources from cluster X and product Y. So the product may need to be set externally in the CR, as I can't see how CRO will know what product the resource belongs too. Unless can we some how get the product name from the name passed to the CR?

pkg/providers/aws/provider_blobstorage.go
ciaranRoche: I do not think we would like to override any existing tags. As we do not know if our SRE team will add their own tags after the fact.

pkg/apis/integreatly/v1alpha1/types/types.go
ciaranRoche: I agree here, if we enforce these labels, we will lose backwards compatible, this will force us to have to create a new version of the CRD `v1alpha2` to avoid completely breaking work already carried out in the intly operator.

deploy/crds/integreatly_v1alpha1_postgres_cr.yaml
dimitraz: Yeah we'll need to pass it in from integreatly. My idea is we have a map of labels in the integreatly product, something like :
```go
labels := &map[string]string{
     "productName": "3scale",
     "clusterId": clusterId,
     ...
}
```

Then when we create the cro resource in integreatly, we label the cr in the modification function:
```go
postgres, err := croUtil.ReconcilePostgres(ctx, client, r.installation.Spec.Type, tier, postgresName, ns, postgresName, ns, func(cr metav1.Object) error {
	// add the labels
	cr.Spec.Labels = labels 

	// add the owner
	ownerutil.EnsureOwner(cr, r.installation)
	return nil
})
```

In cro aws providers, if `cr.Spec.Labels != nil`, we can loop through the map and dynamically add the `key`/`value` pairs 

```go
tags := []*elasticache.Tag{}
for k, v := range cr.Spec.Labels {
    tag := *elasticache.Tag{
        Key: fmt.Sprintf("%s/%s", org, k),
        Value: v,
    }
    tags = append(tags, tag)
}
```

I haven't tried it myself obviously, but I don't see why it wouldn't work

pkg/providers/openshift/provider_postgres.go
dimitraz: This is to fix the bug with 3scale where the system-app pods kept crashing. 3scale 2.6+ requires postgres 10+

ciaranRoche: Perfect, thanks dimitra.

deploy/crds/integreatly_v1alpha1_postgres_cr.yaml
dimitraz: We could also define a struct in the integreatly operator with fixed fields like `ProductName`, `ClusterID`, etc, so that we enforce a standard for key names. We can then have a util function there too which dumps the struct values into a map that we can use to label the cr

pkg/providers/aws/provider_blobstorage.go
austincunningham: @ciaranRoche tried it and it gives a 403, as the comment says don't need credentials, but should probably say doesn't work with credentials.

pkg/providers/aws/provider_blobstorage.go
austincunningham: had that in one of the closed pr but had issues with testing due to it not needing credentials to connect to aws.

pkg/providers/aws/provider_postgres.go
austincunningham: found it didn't work with a the created session either.

pkg/providers/aws/provider_postgres.go
austincunningham: did this in a previous pr as stated above test fail as creating a new session with no credentials.

pkg/providers/aws/provider_redis.go
austincunningham: As they are separate jira's, I believe they should remain separate. Not in favour of large pr's, they tend to make tracking changes difficult and also reverting.

pkg/providers/aws/provider_redis.go
austincunningham: same as above

pkg/providers/aws/provider_postgres.go
austincunningham: probably not but I feel adding a variable makes it easier to read.

version/version.go
austincunningham: added as part of a rebase to one of the close pr, these pr's have been open a long time.

pkg/providers/aws/provider_blobstorage.go
ciaranRoche: It sounds like a permissions issue, AWS can be poor at times with the error codes. Have you tired adding any extra permissions to our credentials https://docs.aws.amazon.com/AmazonS3/latest/dev/using-with-s3-actions.html

Its just so strange it allows for tagging with out credentials at all.

deploy/crds/integreatly_v1alpha1_postgres_cr.yaml
austincunningham: Adding the productName to the CR was outlined in the jira https://issues.redhat.com/browse/INTLY-3968

env , I have no objection to it being a type omit on empty, I think there was a need for it to be an environment variable, and this seemed to be best practice https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/ to set it in the spec but maybe it can be add to another spec rather that the CR

There is this ticket in the backlog for integreatly integration of tagging, https://issues.redhat.com/browse/INTLY-4377, it may be the best place for a discussion. 

ClusterID is got dynamically in CRO so don't need to pass it from Integreatly in the CR.
The operator-sdk seems to give out about `&map[string]string` types as an api violation, But I didn't play around with it too much after the first error.

pkg/providers/aws/provider_blobstorage.go
austincunningham: yea it was a strange to me as well until I found this example https://github.com/aws/aws-sdk-go/blob/master/service/s3/examples_test.go#L1704L1720, the only thing our implementation complained about was it was missing a region with the old new.Session function , so change to the non deprecated method for session and added a region.  Haven't changed anything to the permissions.

pkg/providers/aws/provider_blobstorage.go
austincunningham: As discussed above using the existing session with credential give a 403. Creating a new session with Credential give a 403. Creating a new session with region without credentials works, but this breaks the tests when in a separate function.

pkg/providers/aws/provider_postgres.go
austincunningham: Agree it's quite long , a necessary evil of using the same session for tag snapshot as well, but may be worth a refactor

pkg/providers/aws/provider_blobstorage.go
austincunningham: My reading of the docs is that tag keys have to be unique so it would not allow over writing. Can't put my hands on where I read that now.

austincunningham: found it https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/allocation-tag-restrictions.html

pkg/apis/integreatly/v1alpha1/types/types.go
austincunningham: I have no problem adding omit on empty to these.

deploy/crds/integreatly_v1alpha1_postgres_cr.yaml
dimitraz: Reading that Jira again, it seems like the labels should be added to the `metadata.labels` field in the custom resource: 

```yml
metadata:
  labels:
    integreatly.org/product-name: 3scale
```

I think that's actually really nice because it won't require any breaking changes to the CRD, so it  also won't require as much refactoring on the integreatly side. 

I also personally would be a fan of leaving the env out because it could be easily done as a variable on the integreatly side as well. Not sure what you guys think?

pkg/providers/aws/provider_postgres.go
austincunningham: looks like I missed one

pkg/providers/aws/provider_blobstorage.go
```diff
@@ -156,10 +157,68 @@ func (p *BlobStorageProvider) CreateStorage(ctx context.Context, bs *v1alpha1.Bl
...
+		p.Logger.Infof("env var DEFAULT_ORGANIZATION_TAG : %s", defaultOrganizationTag)
+	}
+
+	// new session but don't need credentials
+	s3svctag := s3.New(session.Must(session.NewSession(&aws.Config{Region: aws.String(stratCfgRegion)})))
+
+	//get Cluster Id
+	clusterId, _ := resources.GetClusterId(ctx, p.Client)
+	// Tagging input
+	for _, productName := range bs.ObjectMeta.Labels {
```
aidenkeating: from my reading of this, we go through each label in the kube resource, and use that label as the `<org-tag>/product-name` tag in AWS, is that correct?

if so, with a kube resource like:

```
metadata:
    name: test
    labels:
        productName: myProduct
        somethingElse: notMyProduct
```

would this perform two tag updates and result in the AWS tag being `<org-tag>/product-name=notMyProduct`?

i know the ticket discussed the idea of allowing  tags to be added arbitrarily based on the label, but I think for now we can just set `<org-tag>/product-name` to be `bs.ObjectMeta.Labels["productName"]` if it exists and isn't an empty string. then we can avoid any looping for now.

if it's an empty string, let's just not add the tag for now. if that makes sense?

pkg/providers/aws/provider_postgres.go
```diff
@@ -214,6 +219,88 @@ func (p *AWSPostgresProvider) createRDSInstance(ctx context.Context, cr *v1alpha
...
+	rdsRegion := *foundInstance.AvailabilityZone
+	rdsRegion = rdsRegion[:(len(rdsRegion))-1]
+	// get provider aws creds so the postgres instance can be deleted
+	rdsSvc := rds.New(session.Must(session.NewSession(&aws.Config{
+		Region: aws.String(rdsRegion),
+	})))
+	//get Cluster Id
+	clusterId, _ := resources.GetClusterId(ctx, p.Client)
+	// Set the Tag values
+	for _, productName := range cr.ObjectMeta.Labels {
```
aidenkeating: same question/suggestion as the one for blob storage

pkg/providers/aws/provider_redis.go
```diff
@@ -166,6 +179,84 @@ func (p *AWSRedisProvider) createElasticacheCluster(ctx context.Context, r *v1al
...
+	// need arn in the following format arn:aws:elasticache:us-east-1:1234567890:cluster:my-mem-cluster
+	arn := "arn:aws:elasticache:" + cacheRegion + ":" + accountId + ":cluster:" + cacheClusterId
+
+	// Set the Tag values
+	defaultOrganizationTag, exists := os.LookupEnv("DEFAULT_ORGANIZATION_TAG")
+	if exists {
+		// Print the value of the environment variable
+		p.Logger.Infof("env var DEFAULT_ORGANIZATION_TAG : %s", defaultOrganizationTag)
+	}
+	for _, productName := range r.ObjectMeta.Labels {
```
aidenkeating: same question/suggestion as the one for blob storage

pkg/providers/aws/provider_redis.go
```diff
@@ -166,6 +179,84 @@ func (p *AWSRedisProvider) createElasticacheCluster(ctx context.Context, r *v1al
...
+func (p *AWSRedisProvider) TagElasticache(ctx context.Context, r *v1alpha1.Redis, stratCfg StrategyConfig, cacheClusterId string, preferredAvailabilityZone string) (types.StatusMessage, error) {
+	cacheSvcTag := elasticache.New(session.Must(session.NewSession(&aws.Config{Region: aws.String(stratCfg.Region)})))
+	cacheRegion := preferredAvailabilityZone
+	cacheRegion = cacheRegion[:(len(preferredAvailabilityZone))-1]
+	// get the account number
+	svc := sts.New(session.New())
+	inputSts := &sts.GetCallerIdentityInput{}
+	id, err := svc.GetCallerIdentity(inputSts)
+	if err != nil {
+		logrus.Error("Error GetCallerIdentity returned an error : s%", err)
```
aidenkeating: should this be `%s`?

pkg/providers/aws/provider_redis.go
```diff
@@ -166,6 +179,84 @@ func (p *AWSRedisProvider) createElasticacheCluster(ctx context.Context, r *v1al
 	}}, croType.StatusMessage(fmt.Sprintf("creation successful, aws elasticache status is %s", *foundCache.Status)), nil
 }
 
+// Add Tags to AWS Elasticache
+func (p *AWSRedisProvider) TagElasticache(ctx context.Context, r *v1alpha1.Redis, stratCfg StrategyConfig, cacheClusterId string, preferredAvailabilityZone string) (types.StatusMessage, error) {
+	cacheSvcTag := elasticache.New(session.Must(session.NewSession(&aws.Config{Region: aws.String(stratCfg.Region)})))
+	cacheRegion := preferredAvailabilityZone
+	cacheRegion = cacheRegion[:(len(preferredAvailabilityZone))-1]
```
aidenkeating: i'm a bit out of the loop here, i'm not sure i understand the redeclaration here, what does `cacheRegion` look like at each of these declarations?

pkg/providers/aws/provider_redis.go
```diff
@@ -166,6 +179,84 @@ func (p *AWSRedisProvider) createElasticacheCluster(ctx context.Context, r *v1al
...
+	}
+	accountId := *id.Account
+	// need arn in the following format arn:aws:elasticache:us-east-1:1234567890:cluster:my-mem-cluster
+	arn := "arn:aws:elasticache:" + cacheRegion + ":" + accountId + ":cluster:" + cacheClusterId
+
+	// Set the Tag values
+	defaultOrganizationTag, exists := os.LookupEnv("DEFAULT_ORGANIZATION_TAG")
+	if exists {
+		// Print the value of the environment variable
+		p.Logger.Infof("env var DEFAULT_ORGANIZATION_TAG : %s", defaultOrganizationTag)
```
aidenkeating: do we need to perform this log, or was it for debugging?

pkg/providers/aws/provider_redis.go
```diff
@@ -166,6 +179,84 @@ func (p *AWSRedisProvider) createElasticacheCluster(ctx context.Context, r *v1al
...
+		}
+		//When snapshots are created add tags to them
+		inputDescirbe := &elasticache.DescribeSnapshotsInput{
+			CacheClusterId: aws.String(cacheClusterId),
+		}
+		snapshotList, _ := cacheSvcTag.DescribeSnapshots(inputDescirbe)
+		if snapshotList.Snapshots != nil {
+			for _, snapshot := range snapshotList.Snapshots {
+				snapshotArn := "arn:aws:elasticache:" + cacheRegion + ":" + accountId + ":snapshot:" + *snapshot.SnapshotName
+				logrus.Infof("Snapshots name : %s", *snapshot.SnapshotName)
```
aidenkeating: i think having a log in this block makes sense, however we could make it more verbose, for example

```
adding operator tags to snapshot <snapshot-name> (arn: <snapshot-arn>)
```

i also think there may be some ability to extract tagging out to a util function, however i will create a separate low priority ticket for this.

pkg/providers/aws/provider_redis.go
```diff
@@ -166,6 +179,84 @@ func (p *AWSRedisProvider) createElasticacheCluster(ctx context.Context, r *v1al
...
+	cacheSvcTag := elasticache.New(session.Must(session.NewSession(&aws.Config{Region: aws.String(stratCfg.Region)})))
+	cacheRegion := preferredAvailabilityZone
+	cacheRegion = cacheRegion[:(len(preferredAvailabilityZone))-1]
+	// get the account number
+	svc := sts.New(session.New())
+	inputSts := &sts.GetCallerIdentityInput{}
+	id, err := svc.GetCallerIdentity(inputSts)
+	if err != nil {
+		logrus.Error("Error GetCallerIdentity returned an error : s%", err)
+		id.Account = aws.String("")
```
aidenkeating: if `id.Account` is `""` can we still perform the tagging? if not, we can probably return an error here

pkg/providers/aws/provider_postgres.go
```diff
@@ -214,6 +219,88 @@ func (p *AWSPostgresProvider) createRDSInstance(ctx context.Context, cr *v1alpha
...
 }
 
+// Tags RDS resources
+func (p *AWSPostgresProvider) TagRDSPostgres(ctx context.Context, cr *v1alpha1.Postgres, foundInstance *rds.DBInstance) (croType.StatusMessage, error) {
+	logrus.Infof("Adding Tags to RDS instance %s", *foundInstance.DBInstanceIdentifier)
+	// get the environment from the CR
+	defaultOrganizationTag, exists := os.LookupEnv("DEFAULT_ORGANIZATION_TAG")
+	if exists {
+		// Print the value of the environment variable
+		p.Logger.Infof("env var DEFAULT_ORGANIZATION_TAG : %s", defaultOrganizationTag)
```
aidenkeating: i think we can remove this log, unless there is a plan to perform some other task in this block?

pkg/providers/aws/provider_blobstorage.go
```diff
@@ -156,10 +157,68 @@ func (p *BlobStorageProvider) CreateStorage(ctx context.Context, bs *v1alpha1.Bl
...
 }
 
+func (p *BlobStorageProvider) TagBlobStorage(ctx context.Context, bucketName string, bs *v1alpha1.BlobStorage, stratCfgRegion string) (croType.StatusMessage, error) {
+	p.Logger.Infof("bucket %s found, Adding tags to bucket", bucketName)
+
+	// get the environment from the CR ,
+	defaultOrganizationTag, exists := os.LookupEnv("DEFAULT_ORGANIZATION_TAG")
+	if exists {
+		// Print the value of the environment variable
+		p.Logger.Infof("env var DEFAULT_ORGANIZATION_TAG : %s", defaultOrganizationTag)
```
aidenkeating: i think this block can be removed

Makefile
```diff
@@ -7,6 +7,7 @@ PREV_VERSION=0.3.0
 VERSION=0.4.0
 COMPILE_TARGET=./tmp/_output/bin/$(IMAGE_NAME)
 OPERATOR_SDK_VERSION=0.12.0
+DEFAULT_ORGANIZATION_TAG=integreatly.org/
```
aidenkeating: i'm thinking we can rename this to something more generic, such as 

```
TAG_KEY_PREFIX
```

if unset or empty, we can default to `integreatly.org/` instead of requiring it to be set in the Makefile or as an env var. if that makes sense?

README.md
```diff
@@ -132,6 +132,22 @@ spec:
   type: managed
 ```
 
+## Resource tagging
```
aidenkeating: awesome, this covers the majority of https://issues.redhat.com/browse/INTLY-4687

pkg/apis/integreatly/v1alpha1/types/types.go
```diff
@@ -20,6 +22,10 @@ type SecretRef struct {
 	Namespace string `json:"namespace,omitempty"`
 }
 
+//type Labels struct {
+//	ProductName string `json:"productname,omitempty"`
```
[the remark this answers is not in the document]
austincunningham: :+1:

deploy/crds/integreatly_v1alpha1_blobstorage_cr.yaml
```diff
@@ -3,6 +3,8 @@ kind: BlobStorage
 metadata:
   # name must be between 1-40 characters
   name: example-blobstorage
+  labels:
+    productname: ProductNameHere
```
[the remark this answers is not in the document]
austincunningham: :+1:

pkg/providers/aws/provider_blobstorage.go
austincunningham: @aidenkeating something like 
```
if bs.ObjectMeta.Labels["productName"] != "" && len(bs.ObjectMeta.Labels["productName"]) > 0 {do stuff}
``` 
or would the length greater that zero be sufficient

pkg/providers/aws/provider_redis.go
austincunningham: :+1:

pkg/providers/aws/provider_redis.go
austincunningham: Basically preferredAvailabilityZone returns the sub zone like eu-west1b, by dropping the last character we get the main zone eu-west1,which is what the api requires, is a bit hacky

Makefile
austincunningham: :+1:

pkg/providers/aws/provider_redis.go
austincunningham: I think this was added to satisfy a test, returning a error will break the same test,  I will have to look into it.

pkg/providers/aws/provider_redis.go
austincunningham: @aidenkeating Originally I had the arn in the log, but removed it as it gives account name, I deemed this a security risk and removed it. Am I being too paranoid ?

aidenkeating: @austincunningham I think that's valid, let's just keep the snapshot name

pkg/providers/aws/provider_redis.go
aidenkeating: Ah ok, can we add that in as a comment in the code? We could hardcode a lookup table. But I believe this approach will work, makes sense.

pkg/providers/aws/provider_blobstorage.go
aidenkeating: @austincunningham It slipped my mind that strings default to an empty string in go, so we can just check for the empty string here, like

```
// define the default list of tags
if bs.ObjectMeta.Labels["productName"] != "" {
    // add the product name tag to the list of tags
}
```

pkg/providers/aws/provider_redis.go
austincunningham: nope its fine

pkg/providers/aws/provider_blobstorage.go
austincunningham: :+1:

pkg/providers/aws/provider_redis.go
austincunningham: Nope breaks tests in CI :( but is fine locally, looks like GetCallerIdentity is getting my local credentials which don't exist in ci.

README.md
austincunningham: credit to you @aidenkeating  it's almost a straight copy of your demo.

pkg/providers/aws/provider_redis.go
aidenkeating: Do you have a link to the test that's failing? I'm not familiar with GetCallerIdentity but I'd like to understand if there are going to be any issues around this when running this in a pod instead of locally.

My cluster setup is still happening, so I can test this out more during verification once it's ready.

austincunningham: https://travis-ci.com/integr8ly/cloud-resource-operator/builds/142886183
```
--- FAIL: TestAWSRedisProvider_TagElasticache (0.00s)
    --- FAIL: TestAWSRedisProvider_TagElasticache/test_tags_reconcile_completes_successfully (0.00s)
        provider_redis_test.go:427: TagElasticache() error = NoCredentialProviders: no valid providers in chain. Deprecated.
            	For verbose messaging see aws.Config.CredentialsChainVerboseErrors, wantErr false
```

pkg/providers/aws/provider_postgres.go
```diff
@@ -214,6 +219,87 @@ func (p *AWSPostgresProvider) createRDSInstance(ctx context.Context, cr *v1alpha
...
+	if !exists {
+		defaultOrganizationTag = TagKeyPrefex
+	}
+
+	// new session but don't need credentials
+	// Get the region from AvailabilityZone
+	rdsRegion := *foundInstance.AvailabilityZone
+	rdsRegion = rdsRegion[:(len(rdsRegion))-1]
+	// get provider aws creds so the postgres instance can be deleted
+	rdsSvc := rds.New(session.Must(session.NewSession(&aws.Config{
```
aidenkeating: We need to define the credentials here and/or define the RDS/Elasticache sessions the same way as we do in existing parts of the code. Otherwise AWS attempts to use your on-machine credentials which won't exist in a pod.

There are existing examples of using these credentials in the pod. Happy to contribute a change/pair up if wanted.

This is the error I end up seeing in the CRs

```
  message: "Failed to add tags to RDS resources: Failed to add Tags to RDS instance: NoCredentialProviders: no valid providers in chain. Deprecated.\n\tFor verbose messaging see aws.Config.CredentialsChainVerboseErrors"
```

pkg/providers/aws/provider_blobstorage.go
```diff
@@ -43,6 +44,7 @@ const (
 	DetailsBlobStorageCredentialKeyID     = "credentialKeyID"
 	DetailsBlobStorageCredentialSecretKey = "credentialSecretKey"
 	defaultForceBucketDeletion            = false
+	TagKeyPrefex                          = "integreatly.org/"
```
dimitraz: Small typo: `TagKeyPrefix`

pkg/providers/aws/provider_blobstorage.go
```diff
@@ -156,10 +158,67 @@ func (p *BlobStorageProvider) CreateStorage(ctx context.Context, bs *v1alpha1.Bl
...
 }
 
+func (p *BlobStorageProvider) TagBlobStorage(ctx context.Context, bucketName string, bs *v1alpha1.BlobStorage, stratCfgRegion string) (croType.StatusMessage, error) {
+	p.Logger.Infof("bucket %s found, Adding tags to bucket", bucketName)
+
+	// get the environment from the CR ,
+	defaultOrganizationTag, exists := os.LookupEnv("TAG_KEY_PREFIX")
+	if !exists {
+		defaultOrganizationTag = TagKeyPrefex
+	}
```
dimitraz: Could use a nice `envOrDefault` util like func here:

```
func getEnv(key, fallback string) string {
    if value, ok := os.LookupEnv(key); ok {
        return value
    }
    return fallback
}
```

deploy/crds/integreatly_v1alpha1_blobstorage_cr.yaml
```diff
@@ -3,6 +3,8 @@ kind: BlobStorage
 metadata:
   # name must be between 1-40 characters
   name: example-blobstorage
+  labels:
```
dimitraz: I think these can be regenerated

pkg/providers/aws/provider_redis.go
```diff
@@ -166,6 +179,85 @@ func (p *AWSRedisProvider) createElasticacheCluster(ctx context.Context, r *v1al
...
+			Tags:         cacheTag,
+		}
+		// add tags
+		_, err = cacheSvcTag.AddTagsToResource(input)
+		if err != nil {
+			msg := "Failed to add tags to AWS Elasticache :"
+			return types.StatusMessage(msg), err
+		}
+		//When snapshots are created add tags to them
+		inputDescirbe := &elasticache.DescribeSnapshotsInput{
```
dimitraz: Small typo: `inputDescribe`

pkg/providers/aws/provider_blobstorage.go
austincunningham: :+1:

pkg/providers/aws/provider_redis.go
austincunningham: :+1:

pkg/providers/aws/provider_blobstorage.go
austincunningham: :+1:

pkg/providers/aws/provider_postgres.go
austincunningham: :+1:

pkg/providers/aws/provider_postgres.go
```diff
@@ -214,6 +218,79 @@ func (p *AWSPostgresProvider) createRDSInstance(ctx context.Context, cr *v1alpha
...
+	}
+	if cr.ObjectMeta.Labels["productName"] != "" {
+		productTag := &rds.Tag{
+			Key:   aws.String(defaultOrganizationTag + "product-name"),
+			Value: aws.String(cr.ObjectMeta.Labels["productName"]),
+		}
+		rdsTag = append(rdsTag, productTag)
+	}
+
+	inputRdsInstance := &rds.AddTagsToResourceInput{
```
ciaranRoche: I don't think we need to declare this as a variable, and can just pass it directly into `AddTagsToResource`

I know it makes it less readable, but it does sorten the code. We have a lot of logic to add labels. 

What do you think?

pkg/providers/aws/provider_postgres_test.go
```diff
@@ -3,27 +3,23 @@ package aws
...
-	"sigs.k8s.io/controller-runtime/pkg/client/fake"
-
-	"testing"
-
 	"github.com/aws/aws-sdk-go/service/rds"
 	"github.com/aws/aws-sdk-go/service/rds/rdsiface"
 	"github.com/integr8ly/cloud-resource-operator/pkg/apis/integreatly/v1alpha1"
-	"github.com/integr8ly/cloud-resource-operator/pkg/providers"
+	croType "github.com/integr8ly/cloud-resource-operator/pkg/apis/integreatly/v1alpha1/types"
+	//"github.com/integr8ly/cloud-resource-operator/pkg/providers"
```
ciaranRoche: We can remove this

pkg/resources/util.go
```diff
@@ -154,3 +155,11 @@ func ReconcileRedis(ctx context.Context, client client.Client, deploymentType, t
 
 	return r, nil
 }
+
+// checks to see if env var exists
+func EnvOrDefault(key string, fallback string) string {
```
ciaranRoche: I think we can remove this util function now.

pkg/resources/util_test.go
```diff
@@ -148,3 +148,31 @@ func TestReconcileBlobStorage(t *testing.T) {
 		})
 	}
 }
+
+func TestEnvOrDefault(t *testing.T) {
```
ciaranRoche: This too can be removed.

pkg/resources/util.go
austincunningham: already done

pkg/resources/util_test.go
austincunningham: done

pkg/providers/aws/provider_postgres_test.go
austincunningham: :+1:

pkg/providers/aws/provider_postgres.go
austincunningham: I like readability but yea the var is only used once, so will remove it

pkg/providers/aws/provider_postgres.go
```diff
@@ -203,6 +202,11 @@ func (p *AWSPostgresProvider) createRDSInstance(ctx context.Context, cr *v1alpha
 		}
 		return nil, croType.StatusMessage(fmt.Sprintf("changes detected, modifyDBInstance() in progress, current aws rds resource status is %s", *foundInstance.DBInstanceStatus)), nil
 	}
+	// Add Tags to Aws Postgres resources
+	_, err = p.TagRDSPostgres(ctx, cr, rdsSvc, foundInstance)
```
ciaranRoche: Just thinking, but shouldn't we be handling a status here? As this is the last step, it should return `creation successful` or else an error status.

aidenkeating: @ciaranRoche i'm not fully understanding your comment here, we handle the returned error state from this function call, on L212 we perform the last return step. are you talking about handling the first, currently ignored, value that's returned by the `#TagRDSPostgres` function?

dimitraz:
Verified! 

/lgtm

PR #75, opened 2019-12-09
INTLY-4355 fix issue with redis snapshot, and tag the snapshot

austincunningham:
## Overview

Jira: https://issues.redhat.com/browse/INTLY-4355

## Verification

- Clone this branch
- Run `make cluster/prepare`
- Run `make run`
- Wait for redis to provision on aws
- delete the redis CR 
- Ensure a snapshot is take on deletion
- Check to see that tag are created on the snapshot, by selecting Backups and selecting the back up created you should see
image

### DIFF
--- a/pkg/providers/aws/provider_redis.go
+++ b/pkg/providers/aws/provider_redis.go
@@ -208,13 +208,24 @@ func (p *AWSRedisProvider) createElasticacheCluster(ctx context.Context, r *v1al
 		// add tags
 		_, err = cacheSvcTag.AddTagsToResource(input)
 
-		//TODO when snapshots are created add tags to them
-		//inputDescirbe := &elasticache.DescribeSnapshotsInput{
-		//	CacheClusterId:      aws.String(cacheClusterId),
-		//}
-		//snapshotList, err := cacheSvcTag.DescribeSnapshots(inputDescirbe)
-		//logrus.Infof("Snapshots List : %s", snapshotList)
+		//When snapshots are created add tags to them
+		inputDescirbe := &elasticache.DescribeSnapshotsInput{
+			CacheClusterId: aws.String(cacheClusterId),
+		}
+		snapshotList, err := cacheSvcTag.DescribeSnapshots(inputDescirbe)
+		if snapshotList.Snapshots != nil {
+			for _, snapshot := range snapshotList.Snapshots {
+				snapshotArn := "arn:aws:elasticache:" + cacheRegion + ":" + accountId + ":snapshot:" + *snapshot.SnapshotName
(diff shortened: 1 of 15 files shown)


pkg/providers/aws/provider_redis.go
```diff
@@ -349,7 +421,7 @@ func (p *AWSRedisProvider) buildElasticacheDeleteConfig(ctx context.Context, r v
 	if err != nil {
 		return errorUtil.Wrapf(err, "failed to retrieve rds config")
 	}
-	if elasticacheDeleteConfig.FinalSnapshotIdentifier != nil && *elasticacheDeleteConfig.FinalSnapshotIdentifier == "" {
+	if elasticacheDeleteConfig.FinalSnapshotIdentifier == nil {
```
ciaranRoche: I think we do not want to change this. As the api for elasti-cache is not as powerfull as the rds api. We do not have a parameter that allows us to trigger the creation of the snapshot. Because of this we decided on using the `FinalSnapshotIdentifer` from the api, if it has any value other then an empty string then set it as our generated identifer.

austincunningham: will try running with the operator-sdk up local with a namespace. 
might just be the `!= nil` is the issue

ciaranRoche: Yes, it is worth trying it with passing a namespace to `make run`, I only tested it with `make code/run/service_account`, Just an FYI though, we want that check, having it `== nil` will mean we take a snapshot every delete. We do not want this behavior, especially during development.
Currently we just want to check if it is not equal to nil, and if it is an empty string, to generate the indentifier, otherwise just use the passed one. 
This was the approach we decided on, as we lack the `SkipFinalSnapshot` we get with `RDS` parameter to easily trigger or disable a snapshot.

austincunningham: Has our backup policy changed since this jira https://issues.redhat.com/browse/INTLY-3568 ? If so backup on delete is happening with RDS, should elasticache be different ?

ciaranRoche: No it is the exact same, we just want a way to turn this off. We can turn off RDS by passing `SkipFinalSnapshot:false`, we can turn off elasticache by not passing `FinalSnapshotIdentifier`

This was the solution we decided on, as we are trying to leverage `DeleteReplicationGroupOutput` with out having to create our own `struct`

austincunningham: I get you , Thanks @ciaranRoche

austincunningham: @ciaranRoche verified adding namespace seems to work, will close this ticket and move the snapshot tagging to all tagging pr.

PR #76, opened 2019-12-11
add new run service account target

ciaranRoche:
## Overview

Jira: https://issues.redhat.com/browse/INTLY-4503

## Verification

- Clone this branch
- Run `make cluster/prepare`
- Seed the cluster with some resources 
- Run `make code/run/service_account`
- Ensure the operator behaves as expected

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,19 @@ build:
 run:
 	RECTIME=30 operator-sdk up local --namespace=""
 
+.PHONY: setup/service_account
+setup/service_account:
+	@oc replace --force -f deploy/role.yaml -n $(NAMESPACE)
+	@oc replace --force -f deploy/cluster_role.yaml -n $(NAMESPACE)
+	@oc replace --force -f deploy/service_account.yaml -n $(NAMESPACE)
+	@oc replace --force -f deploy/role_binding.yaml -n $(NAMESPACE)
+	@cat deploy/role_binding.yaml | sed "s/namespace: cloud-resource/namespace: $(NAMESPACE)/g" | oc replace --force -f -
+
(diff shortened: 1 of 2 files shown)


deploy/role.yaml
```diff
@@ -88,4 +88,96 @@ rules:
   verbs:
   - '*'
   resources:
-  - credentialsrequests
\ No newline at end of file
+  - credentialsrequests
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
dimitraz: We have a separate file for cluster roles - https://github.com/integr8ly/cloud-resource-operator/blob/master/deploy/cluster_role.yaml

deploy/role_binding.yaml
```diff
@@ -9,3 +9,16 @@ roleRef:
   kind: Role
   name: cloud-resource-operator
   apiGroup: rbac.authorization.k8s.io
+---
+kind: ClusterRoleBinding
```
dimitraz: I would be in favour of moving this out to a separate file

Makefile
```diff
@@ -26,6 +26,18 @@ build:
 run:
 	RECTIME=30 operator-sdk up local --namespace=""
 
+.PHONY: setup/service_account
+setup/service_account:
```
austincunningham: Should we have a make target to remove the service account as well ? Did have issues with this in the passed with integreatly's service account locking me out. 

Wait its tied to the namespace, so as soon as the namespace is gone it should be gone right ?

ciaranRoche: Yeah it is all added to the `namespace`, just have to remember to log yourself back in after running this target and everything should be good

PR #77, opened 2019-12-16
add skip create logic to aws redis and postgres

ciaranRoche:
## Overview

Jira: https://issues.redhat.com/browse/INTLY-4345

## Verification

- Clone this branch
- Run `make cluster/prepare`
- Seed the cluster with the following
```
kind: Redis
metadata:
  name: example-redis
spec:
  secretRef:
    name: example-redis-sec
  tier: development
  type: managed
  skipCreate: true
```
```
kind: Postgres
metadata:
  name: example-postgres
spec:
  secretRef:
    name: example-postgres-sec
  tier: development
  type: managed
  skipCreate: true
```
- Run `make run`
- Ensure both Postgres and Redis creation has not begun in AWS and the `Status` is updated to `paused`
- Remove `skipCreate: true` from CR's or set `skipCreate:false`
- Ensure both Postgres and Redis creation starts in AWS and the `Status` changed to reflect that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,11 @@ Currently AWS resources are deployed into a separate Virtual Private Cloud (VPC)
 
 The two VPCs should now be able to communicate with each other. 
 
+## Skip Create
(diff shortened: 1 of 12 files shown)


dimitraz:
:eyes:

PR #83, opened 2020-01-13
Allow arbitrary redis and elasticache snapshots

ciaranRoche:
## Overview

Jira: https://issues.redhat.com/browse/INTLY-3697

## Verification 

- Run `make cluster/prepare`
- Add the following CR's 
```
kind: PostgresSnapshot
metadata:
  name: example-postgressnapshot
spec:
  # The postgres resource name for the snapshot you want to take
  resourceName: example-postgres
```
```
kind: RedisSnapshot
metadata:
  name: example-redissnapshot
spec:
  # The redis resource name for the snapshot you want to take
  resourceName: example-redis
```
- Ensure both redis and postgres create with no errors
- Ensure once resource are created, the snapshots begin. This can be seen from the logs, or the status of the `RedisSnapshot` CR or the `PostgresSnapshot` CR. 
- Verify in AWS the snapshots are beinging created. 
- *Note* in elastcache we will only create a snapshot of the primary node.


README.md
```diff
@@ -90,6 +90,20 @@ Currently AWS resources are deployed into a separate Virtual Private Cloud (VPC)
...
+apiVersion: integreatly.org/v1alpha1
+kind: RedisSnapshot
+metadata:
+  name: my-redis-snapshot
+spec:
+  # The redis resource name for the snapshot you want to take
+  resourceName: my-redis-resource
+
+```  
+*Note* You may experience some downtime in the resource during the creation of the Snapshot
```
aidenkeating: Is this a reference to the Elasticache documentation around snapshotting?

Does RDS have the same warning around possible impacted service?

pkg/controller/postgressnapshot/postgressnapshot_controller.go
```diff
@@ -0,0 +1,226 @@
...
+
+// Reconcile reads that state of the cluster for a PostgresSnapshot object and makes changes based on the state read
+// and what is in the PostgresSnapshot.Spec
+func (r *ReconcilePostgresSnapshot) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	r.logger.Info("reconciling postgres snapshot")
+	ctx := context.TODO()
+
+	// Fetch the PostgresSnapshot instance
+	instance := &integreatlyv1alpha1.PostgresSnapshot{}
+	err := r.client.Get(context.TODO(), request.NamespacedName, instance)
```
aidenkeating: Can we just use `ctx` here instead of repeating `context.TODO()`?

pkg/controller/postgressnapshot/postgressnapshot_controller.go
```diff
@@ -0,0 +1,226 @@
...
+			// Return and don't requeue
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		return reconcile.Result{}, err
+	}
+
+	// check status, if complete return
+	if instance.Status.Phase == croType.PhaseComplete {
+		r.logger.Infof("found existing rds snapshot for %s", instance.Name)
```
aidenkeating: I think this could cause some confusion when reading the logs. We haven't found an existing snapshot here, we're skipping creation of a snapshot as the phase is complete. I think we could change the wording here to something like that.

I think this check is fine to keep as we cannot create a snapshot from a previous time/date so if a user goes into AWS and removes the snapshot, we cannot really recreate it.

It would be nice if we could report on a snapshot for this time having been cleaned up and have a status to reflect that. But I don't think that was part of this ticket.

pkg/controller/postgressnapshot/postgressnapshot_controller_test.go
```diff
@@ -0,0 +1,154 @@
...
+
+func (m *mockRdsClient) CreateDBSnapshot(*rds.CreateDBSnapshotInput) (*rds.CreateDBSnapshotOutput, error) {
+	return &rds.CreateDBSnapshotOutput{
+		DBSnapshot: m.dbSnapshot,
+	}, nil
+}
+
+func TestReconcilePostgresSnapshot_createSnapshot(t *testing.T) {
+	//ctx := context.TODO()
+	//snapshotName, err := croAws.BuildTimestampedInfraNameFromObjectCreation(ctx, r.client, snapshot.ObjectMeta, croAws.DefaultAwsIdentifierLength)
```
aidenkeating: Can we remove these comment blocks?

pkg/controller/redissnapshot/redissnapshot_controller.go
```diff
@@ -0,0 +1,246 @@
...
+	if err != nil {
+		errMsg := fmt.Sprintf("failed to get redis cr : %s", err.Error())
+		if updateErr := resources.UpdateSnapshotPhase(ctx, r.client, instance, croType.PhaseFailed, croType.StatusMessage(errMsg)); updateErr != nil {
+			return reconcile.Result{}, updateErr
+		}
+	}
+
+	// check redis cr deployment type is aws
+	if redisCr.Status.Strategy != providers.AWSDeploymentStrategy {
+		errMsg := "none supported deployment strategy"
```
aidenkeating: This message is difficult to understand. Maybe something like:

`the resource <resourceName> uses an unsupported provider strategy <strategy>, only resources using the aws provider are valid`

pkg/controller/redissnapshot/redissnapshot_controller.go
```diff
@@ -0,0 +1,246 @@
...
+		return croType.PhaseInProgress, "snapshot started", nil
+	}
+
+	// if snapshot status complete update status
+	if *foundSnapshot.SnapshotStatus == "available" {
+		return croType.PhaseComplete, "snapshot created", nil
+	}
+
+	msg := fmt.Sprintf("current snapshot status : %s", *foundSnapshot.SnapshotStatus)
+	r.logger.Info(msg)
```
aidenkeating: We don't reuse `msg`, we can probably just keep it inline

pkg/providers/aws/provider_postgres_test.go
```diff
@@ -265,8 +264,6 @@ func TestAWSPostgresProvider_createPostgresInstance(t *testing.T) {
 				t.Errorf("createRDSInstance() error = %v, wantErr %v", err, tt.wantErr)
 				return
 			}
-			fmt.Println(got)
-			fmt.Println(tt.want)
```
aidenkeating: :+1:

pkg/controller/redissnapshot/redissnapshot_controller.go
```diff
@@ -0,0 +1,246 @@
...
+		}
+	}
+
+	// get replication group
+	cacheOutput, err := cacheSvc.DescribeReplicationGroups(&elasticache.DescribeReplicationGroupsInput{
+		ReplicationGroupId: aws.String(clusterName),
+	})
+
+	// ensure replication group is available
+	if *cacheOutput.ReplicationGroups[0].Status != "available" {
```
aidenkeating: Is it possible to have no elements in the ReplicationGroups list? I'd assume no, but just want to clarify if we need a check here

pkg/providers/aws/provider_redis.go
```diff
@@ -257,7 +276,7 @@ func (p *AWSRedisProvider) TagElasticacheNode(ctx context.Context, cacheSvc elas
 			}
 			_, err = cacheSvc.AddTagsToResource(snapshotInput)
 			if err != nil {
-				msg := "Failed to add tags to AWS Elasticache Snapshot:"
+				msg := "Failed to add tags to aws elasticache snapshot:"
```
aidenkeating: Should this be `failed to add....`

README.md
ciaranRoche: Tbh, I added that note to remind myself to return to this. I will add links to the documentation around snapshotting and any impact it may have

deploy/role.yaml
```diff
@@ -83,9 +83,11 @@ rules:
   - clusteroperators
   - authentications
   - builds
+  verbs:
+  - '*'
 - apiGroups:
   - cloudcredential.openshift.io
+  resources:
+  - credentialsrequests
```
aidenkeating: have we added these new permissions to the cluster service version?

PR #86, opened 2020-01-17
Fix CSV Formatting & Add Product Name to Util Function

ciaranRoche:
## Overview

Currently description and display name is not generated by the Operator SDK

This change adds both for the new crd`s in the csv

## Verify
- Update MakeFile to target your Quay
- Run `make image/push`
- Run `make manifest/push`
- Ensure you set the Quay Application and Repo to publc
- Create a new Operator Source pointing at your Quay
```
apiVersion: operators.coreos.com/v1
kind: OperatorSource
metadata:
  name: integreatly-operators
spec:
  authorizationToken: {}
  displayName: Integreatly Operators
  endpoint: 'https://quay.io/cnr'
  publisher: Integreatly Publisher
  registryNamespace: <<YOUR QUAY NAME>>
  type: appregistry
```
- Create a `Cloud Resource Operator` subscription from OLM
- Ensure the subscription is a success and the operator is installed

### DIFF
--- a/pkg/resources/config.go
+++ b/pkg/resources/config.go
@@ -14,8 +14,8 @@ import (
 const (
 	EnvForceReconcileTimeout = "ENV_FORCE_RECONCILE_TIMEOUT"
 	DefaultTagKeyPrefix      = "integreatly.org/"
-	ErrorReconcileTime 		 = time.Second * 30
-	SuccessReconcileTime	 = time.Second * 60
+	ErrorReconcileTime       = time.Second * 30
+	SuccessReconcileTime     = time.Second * 60
 )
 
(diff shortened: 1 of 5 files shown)


pkg/resources/config.go
```diff
@@ -14,8 +14,8 @@ import (
 const (
 	EnvForceReconcileTimeout = "ENV_FORCE_RECONCILE_TIMEOUT"
 	DefaultTagKeyPrefix      = "integreatly.org/"
-	ErrorReconcileTime 		 = time.Second * 30
```
aidenkeating: can we run a format?

aidenkeating: ignore me

PR #87, opened 2020-01-20
Added custom metrics for AWS RDS and ElastiCache scheduled maintenance

davidkirwan:
## Overview

Jira: https://issues.redhat.com/browse/INTLY-4669

## Verification

- Checkout the branch `adding_metrics` on my fork `github.com/davidkirwan/cloud-resource-operator`
- Run `make cluster/prepare`
- Run `make run`
- metrics will be created, `curl localhost:8383/metrics` and expect to see `cro_aws_elasticache_service_maintenance`, `cro_aws_rds_service_maintenance`, `cro_aws_elasticache_info` and `cro_aws_rds_info` eg:
```
cro_aws_elasticache_service_maintenance{AutoUpdateAfterRecommendedApplyByDate="false",Engine="redis",EstimatedUpdateTime="30 minutes per node",ServiceUpdateDescription="Upgrades to improve the security, reliability, and operational performance of your ElastiCache nodes",ServiceUpdateEndDate="1570456320",ServiceUpdateName="elc-20190604-001",ServiceUpdateRecommendedApplyByDate="1563105599",ServiceUpdateReleaseDate="1560513600",ServiceUpdateSeverity="important",ServiceUpdateStatus="expired",ServiceUpdateType="security-update"} 1.563105599e+09
cro_aws_elasticache_service_maintenance{AutoUpdateAfterRecommendedApplyByDate="false",Engine="redis, memcached",EstimatedUpdateTime="30 minutes per node",ServiceUpdateDescription="Upgrades to improve the security, reliability, and operational performance of your ElastiCache nodes",ServiceUpdateEndDate="1581861599",ServiceUpdateName="elc-20191007-001",ServiceUpdateRecommendedApplyByDate="1573826399",ServiceUpdateReleaseDate="1571234400",ServiceUpdateSeverity="important",ServiceUpdateStatus="available",ServiceUpdateType="security-update"} 1.573826399e+09
```

*Note* maintenance metrics might not show, but this is expected if no maintenance is due :)

For verification purposes you can set the `pkg/resources/custom_metrics.go:L13` constant `sleepytime` to `30` seconds, then look for the following log message:

```
INFO[0030] calling reset on all prometheus gauge vectors  custom_metrics=StartGaugeVector
```


pkg/providers/aws/provider_redis.go
```diff
@@ -99,6 +101,34 @@ func (p *AWSRedisProvider) CreateRedis(ctx context.Context, r *v1alpha1.Redis) (
 	// setup aws elasticache cluster sdk session
 	cacheSvc, stsSvc := createAWSService(stratCfg, providerCreds)
 
+	// Retrieve service maintenance updates, create and export Prometheus metrics
+	output, err := cacheSvc.DescribeServiceUpdates(&elasticache.DescribeServiceUpdatesInput{})
+	if err != nil {
+		logrus.Info(fmt.Sprintf("elasticache serviceupdates error: %s:", err))
```
ciaranRoche: We can return on this error
```
if err != nil {
     errMsg := "elasticache service updates error"
     return nil, croType.StatusMessage(errMsg), errorUtil.Wrap(err, errMsg)```
}

Once we return on this error, we can then remove the `else` statement

File: pkg/providers/aws/provider_redis.go [hunk 2]
```diff
@@ -99,6 +101,34 @@ func (p *AWSRedisProvider) CreateRedis(ctx context.Context, r *v1alpha1.Redis) (
...
+	} else {
+		logrus.Info(fmt.Sprintf("there are elasticache serviceupdates: %s available", len(output.ServiceUpdates)))
+		metricName := "cro_aws_elasticache_service_maintenance"
+		for _, su := range output.ServiceUpdates {
+			metricLabels := map[string]string{}
+
+			metricLabels["AutoUpdateAfterRecommendedApplyByDate"] = strconv.FormatBool(*su.AutoUpdateAfterRecommendedApplyByDate)
```
ciaranRoche: Do you know if any of these labels can be `nil` or are they always included in `DescribeServiceUpdatesOutput`?

pkg/providers/aws/provider_redis.go
```diff
@@ -99,6 +101,34 @@ func (p *AWSRedisProvider) CreateRedis(ctx context.Context, r *v1alpha1.Redis) (
 	// setup aws elasticache cluster sdk session
 	cacheSvc, stsSvc := createAWSService(stratCfg, providerCreds)
 
+	// Retrieve service maintenance updates, create and export Prometheus metrics
+	output, err := cacheSvc.DescribeServiceUpdates(&elasticache.DescribeServiceUpdatesInput{})
+	if err != nil {
+		logrus.Info(fmt.Sprintf("elasticache serviceupdates error: %s:", err))
+	} else {
+		logrus.Info(fmt.Sprintf("there are elasticache serviceupdates: %s available", len(output.ServiceUpdates)))
+		metricName := "cro_aws_elasticache_service_maintenance"
```
ciaranRoche: These metric names, I think we can move them back to `custom_metrics.go` as a const, and just export them. So we can reference them `resources.PostgresMaintenanceMetricName`, at least then we can keep all metric names together

pkg/providers/aws/provider_redis.go
davidkirwan: That I don't know will need to look at the SDK docs

pkg/providers/aws/provider_redis.go
davidkirwan: I'd prefer to stick to the interface set by the: `SetMetric(name string, labels map[string]string, epochTimestamp int64)`

We can go with a const in `pkg/providers/aws/provider_redis.go` though?

ciaranRoche: Yeah either way I am not that pushed, my thinking was trying to keep as much metric related stuff and things out of the providers/controllers to avoid bloat. But since its a var name, i have not strong opinion :)

pkg/providers/aws/provider_redis.go
ciaranRoche: Yeah it is worth just to clarify, to avoid any unwanted behavior

davidkirwan: From the AWS Documentation it does not mention that these results contain optional data, from this we're going to assume they will always be returned.

pkg/providers/aws/provider_postgres.go
```diff
@@ -129,6 +132,32 @@ func (p *AWSPostgresProvider) CreatePostgres(ctx context.Context, pg *v1alpha1.P
...
+	if err != nil {
+		msg := "rds serviceupdates error"
+		return nil, croType.StatusMessage(msg), errorUtil.Wrap(err, msg)
+	}
+
+	logrus.Info(fmt.Sprintf("rds serviceupdates: %d available", len(output.PendingMaintenanceActions)))
+	metricName := CRO_AWS_RDS_SERVICE_MAINTENANCE
+	for _, su := range output.PendingMaintenanceActions {
+		metricLabels := map[string]string{}
+
```
ciaranRoche: I think maybe we can add more labels here, the likes of `clusterID`, `resourceID`, `namespace`, `rdsIdentifier`. 
What do you think? we could create a more meaning full alert

pkg/providers/aws/provider_postgres.go
```diff
@@ -35,8 +37,9 @@ import (
 )
 
 const (
-	postgresProviderName       = "aws-rds"
-	DefaultAwsIdentifierLength = 40
+	CRO_AWS_RDS_SERVICE_MAINTENANCE = "cro_aws_rds_service_maintenance"
```
ciaranRoche: It looks like the best practice in golang, is to use `camelCase`, even for `consts`.

pkg/providers/aws/provider_postgres.go
```diff
@@ -129,6 +132,32 @@ func (p *AWSPostgresProvider) CreatePostgres(ctx context.Context, pg *v1alpha1.P
 	// setup aws RDS instance sdk session
 	rdsSession := createRDSSession(stratCfg, providerCreds)
 
+	// Retrieve service maintenance updates, create and export Prometheus metrics
```
ciaranRoche: What do you think about we abstract this logic to separate functions. These create functions are becoming more and more bloated.

pkg/resources/custom_metrics.go
```diff
@@ -0,0 +1,57 @@
...
+	StartGaugeVector()
+}
+
+// periodic loop thats wiping the vector.. maybe every hour
+func StartGaugeVector() {
+	MetricVecs = map[string]prometheus.GaugeVec{}
+
+	go func() {
+		for {
+			for _, val := range MetricVecs {
```
ciaranRoche: Should we log this, just to inform that we are resetting all metrics.

ciaranRoche:
`make code/fix` will sort the CI issue :)

davidkirwan:
/cc @aidenkeating 
/cc @david-martin 

Needs approval when you get a chance please!

david-martin:
/lgtm

@aidenkeating would you mind approving after review?

austinmartinh:
I might have misread that comment @davidkirwan

PR #90, opened 2020-01-21
🚑 Fix dependency on openshift/api

grdryn:
It looks like the openshift/api project deleted all of their tags, so
now the following error is thrown any time we run e.g. a go mod tidy
in any project that depends on it:

```
go: github.com/operator-framework/operator-sdk@v0.12.1-0.20191112211508-82fc57de5e5b requires
	github.com/operator-framework/operator-lifecycle-manager@v0.0.0-20190605231540-b8a4faf68e36 requires
	github.com/openshift/api@v3.9.1-0.20190424152011-77b8897ec79a+incompatible: invalid pseudo-version: preceding tag (v3.9.0) not found
```

For this change, all I did was set an override in the "replace"
section of the go.mod, and ran `go mod tidy` (which changed a bunch of
other things...).

See openshift/api#456 for more info on the issue.

## Verification

- Clone this branch
- Run `make cluster/prepare`
- Run `make run`
- Ensure...


go.mod
```diff
@@ -22,23 +22,22 @@ require (
...
 	k8s.io/client-go v11.0.0+incompatible
-	k8s.io/klog v1.0.0 // indirect
 	k8s.io/kube-openapi v0.0.0-20190918143330-0270cf2f1c1d
 	sigs.k8s.io/controller-runtime v0.3.0
 )
 
 // Pinned to kubernetes-1.15.4
 replace (
+
+	github.com/openshift/api => github.com/openshift/api v0.0.0-20200117162508-e7ccdda6ba67
```
grdryn: For some reason, even if I change the dependency in the "require" section (L17), it always gets reset by `go mod tidy` :thinking:

davidffrench: that is expected and what the replace is needed for

grdryn: @davidffrench why wouldn't it let us just specify an expected version in the `require` section? I thought the `replace` was more for transitive dependency versions, but that's just based on what I expect it to be for, rather than having read it somewhere.

davidffrench: Maybe it is possible but this is what I have noticed from my own use. A replace is about pinning to a specific version so even if I run `go get <module>@latest`, the require will be updated but the replace is still pinning it. `go mod tidy` can also change the require but I don't know under what circumstances.

HubertStefanski: Austin said he had success in fixing this issue, there's a change to go.mod in this PR https://github.com/integr8ly/cloud-resource-operator/commit/47646dbdff784f7dc09c0cd73dad7880f29892d8

grdryn:
Setting `GOPROXY=https://proxy.golang.org` is a sufficient workaround for this for now

PR #91, opened 2020-01-23
add available metrics

ciaranRoche:
## Overview

In order to make a cleaner alert we decided on returning a value of `1` to represent an `available resource` and a value of `0` to represent an `unavailable resource`

## Verification

- Clone this branch
- Run `make cluster/prepare`
- Run `make run`
- Run `curl localhost:8383/metrics | grep aws `
- Ensure `cro_aws_rds_available` and `cro_aws_elasticache_available` metric and the value represents the current state of the provisioning resource

### DIFF
--- a/pkg/providers/aws/provider_redis.go
+++ b/pkg/providers/aws/provider_redis.go
@@ -33,9 +33,10 @@ import (
 )
 
 const (
-	CROAWSElastiCacheServiceMaintenance = "cro_aws_elasticache_service_maintenance"
-	defaultCroAwsElasticacheInfo        = "cro_aws_elasticache_info"
-	redisProviderName                   = "aws-elasticache"
+	defaultRedisMaintenanceMetricName = "cro_aws_elasticache_service_maintenance"
+	defaultRedisInfoMetricName        = "cro_aws_elasticache_info"
+	defaultRedisAvailMetricName       = "cro_aws_elasticache_avail"
+	redisProviderName                 = "aws-elasticache"
 	// default create params
 	defaultCacheNodeType     = "cache.t2.micro"
 	defaultEngineVersion     = "3.2.10"
@@ -496,14 +497,19 @@ func (p *AWSRedisProvider) buildElasticacheDeleteConfig(ctx context.Context, r v
 	return nil
 }
 
-func buildRedisInfoMetricLables(r *v1alpha1.Redis, cache *elasticache.ReplicationGroup, clusterID string) (map[string]string, error) {
+func buildRedisInfoMetricLables(r *v1alpha1.Redis, cache *elasticache.ReplicationGroup, clusterID string) map[string]string {
+	labels := buildRedisGenericMetricLabels(r, cache, clusterID)
+	labels["status"] = *cache.Status
+	return labels
+}
+
+func buildRedisGenericMetricLabels(r *v1alpha1.Redis, cache *elasticache.ReplicationGroup, clusterID string) map[string]string {
 	labels := map[string]string{}
 	labels["clusterID"] = clusterID
 	labels["resourceID"] = r.Name
 	labels["namespace"] = r.Namespace
 	labels["instanceID"] = *cache.ReplicationGroupId
-	labels["status"] = *cache.Status
-	return labels, nil
+	return labels
 }
 
 func (p *AWSRedisProvider) setRedisInfoMetric(ctx context.Context, cr *v1alpha1.Redis, instance *elasticache.ReplicationGroup) error {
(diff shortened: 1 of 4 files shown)


davidkirwan:
/lgtm

Working good. Once the RDS instance has been provisioned successfully the metric `cro_aws_rds_available` correctly returns 1

davidkirwan:
/cc @aidenkeating 
/cc @david-martin 

Need an approval please guys

david-martin:
@davidkirwan  or @ciaranRoche Would you have an example of the metric with all labels?

When an alert is defined, I'd like to ensure enough info is available in the metric to help identify the problem instance

ciaranRoche:
@david-martin ```cro_aws_rds_info{clusterID="croche-xp7kv",instanceID="crochexp7kvcloudresourceoperatorexa-zgo3",namespace="cloud-resource-operator",resourceID="example-postgres-2"} 1.5797774944308903e+09
```
We have to be careful with what labels we add, as if the label values change dynamically we can end up with multiple metrics for the same resource. 
I was thinking pointing to the resource should be enough, along with the namespace and the CR. It should make it easy to narrow down the problem. And we are guaranteed that these values wont change eliminating the chance of multiple metrics and false positive alerts

**Comment by david-martin** [comment]
> We have to be careful with what labels we add, as if the label values change dynamically we can end up with multiple metrics for the same resource.
> I was thinking pointing to the resource should be enough, along with the namespace and the CR. It should make it easy to narrow down the problem. And we are guaranteed that these values wont change eliminating the chance of multiple metrics and false positive alerts

@ciaranRoche OK, that make sense. 
Which label allows you to 'join' with the `cro_aws_rds_info` & `cro_aws_rds_service_maintenance` metrics?
Also, does `cro_aws_rds_info` provide more labels/info than the new `cro_aws_rds_available` would?

If yes to both, then I think we're in good shape

**Comment by ciaranRoche** [comment]
@david-martin currently the `info` and `available` metrics share [generic labels](https://github.com/integr8ly/cloud-resource-operator/pull/91/files#diff-8fd2a41e86b5db36e6443ed519133f75R590), so `clusterID`, `resourceID`, `namespace` and `instanceID`. The info currently only has an extra label `status` of the instance. But we can easily add any other labels that might be of benefit. 

As for joining `info` and `maintenance` they share a `clusterID`. For RDS we also share `resourceID` on both `info` and `maintenance`

File: pkg/providers/aws/provider_redis.go [hunk 1]
```diff
@@ -514,13 +520,22 @@ func (p *AWSRedisProvider) setRedisInfoMetric(ctx context.Context, cr *v1alpha1.
...
+	}
+
+	// set available metric
+	if *instance.Status != "available" {
+		if err := resources.SetMetric(defaultRedisAvailMetricName, genericLabels, 0); err != nil {
+			return err
+		}
+		return nil
+	}
+	if err := resources.SetMetric(defaultRedisAvailMetricName, genericLabels, 0); err != nil {
```
austinmartinh:
```suggestion
	if err := resources.SetMetric(defaultRedisAvailMetricName, genericLabels, 1); err != nil {
```

ciaranRoche: good catch :+1:

davidkirwan:
```
INFO[0252] setting redis information metric             
INFO[0252] waiting for redis cluster to become available  controller=controller_redis
....
$ curl localhost:8383/metrics | grep cro_aws_elasticache_available
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 35369    0 35369    0     0  5756k      0 --:--:-- --:--:-- --:--:-- 5756k
# HELP cro_aws_elasticache_available 
# TYPE cro_aws_elasticache_available gauge
cro_aws_elasticache_available{clusterID="dkirwan-s7w9x",instanceID="dkirwans7w9xcloudresourceoperatorex-cu3g",namespace="cloud-resource-operator",resourceID="example-redis"} 0
```


```
INFO[0611] found existing rds instance                  
INFO[0611] adding tags to rds instance dkirwans7w9xcloudresourceoperatorex-nm2p 
INFO[0611] Tags were added successfully to the RDS instance dkirwans7w9xcloudresourceoperatorex-nm2p

...
$ curl localhost:8383/metrics | grep cro_aws_elasticache_available
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0# HELP cro_aws_elasticache_available 
# TYPE cro_aws_elasticache_available gauge
cro_aws_elasticache_available{clusterID="dkirwan-s7w9x",instanceID="dkirwans7w9xcloudresourceoperatorex-cu3g",namespace="cloud-resource-operator",resourceID="example-redis"} 1
100 37651    0 37651    0     0  9192k      0 --:--:-- --:--:-- --:--:-- 9192k
```

PR #94, opened 2020-01-24
fixed comment syntax

austinmartinh:
## Overview

Fixed comment syntax to remove linting errors

## Verification

- Clone this branch
- Run `make cluster/prepare`
- Run `make run`
- Ensure...

### DIFF
--- a/pkg/providers/aws/config.go
+++ b/pkg/providers/aws/config.go
@@ -36,7 +36,7 @@ const (
 	sesSMTPEndpointEUWest1 = "email-smtp.eu-west-1.amazonaws.com"
 )
 
-//DefaultConfigMapNamespace get default namespace
+//DefaultConfigMapNamespace is the default namespace that Configmaps will be created in
 var DefaultConfigMapNamespace, _ = k8sutil.GetWatchNamespace()
 
 //go:generate moq -out config_moq.go . ConfigManager
--- a/pkg/providers/aws/config.go
+++ b/pkg/providers/aws/config.go
@@ -36,7 +36,7 @@ const (
 	sesSMTPEndpointEUWest1 = "email-smtp.eu-west-1.amazonaws.com"
 )
 
-//get default namespace
+//DefaultConfigMapNamespace get default namespace
 var DefaultConfigMapNamespace, _ = k8sutil.GetWatchNamespace()
 
 //go:generate moq -out config_moq.go . ConfigManager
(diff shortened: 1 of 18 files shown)


deploy/crds/integreatly_v1alpha1_postgres_cr.yaml
```diff
@@ -2,14 +2,14 @@ apiVersion: integreatly.org/v1alpha1
 kind: Postgres
 metadata:
   # name must be between 1-40 characters
-  name: example-postgres
+  name: example-postgres-man
```
ciaranRoche: I think we can remove these changes to the CR

austinmartinh: My mistake, I did mean to revert them

deploy/crds/integreatly_v1alpha1_redis_cr.yaml
```diff
@@ -3,7 +3,6 @@ kind: Redis
 metadata:
   # name must be between 1-40 characters
   name: example-redis
-  labels:
```
ciaranRoche: Do we want to remove this from the CR?

austinmartinh: No, added it back in there

pkg/providers/aws/config.go
```diff
@@ -36,7 +36,7 @@ const (
 	sesSMTPEndpointEUWest1 = "email-smtp.eu-west-1.amazonaws.com"
 )
 
-//get default namespace
+//DefaultConfigMapNamespace get default namespace
```
grdryn: [The "Commentary" section of Effective Go](https://golang.org/doc/effective_go.html#commentary) says:
> Doc comments work best as complete sentences, which allow a wide variety of automated presentations. The first sentence should be a one-sentence summary that starts with the name being declared. 

I believe that's the practice that this lint rule is trying to enforce, so I think this could be improved by re-wording -- the old comment was really describing the action of calling `k8sutil.GetWatchNamespace()`, but the new one should be describing `DefaultConfigMapNamespace`.

```suggestion
// DefaultConfigMapNamespace is the default namespace that ConfigMaps will be created in.
```

I think there are a couple of other comment changes in this PR that similarly don't roll off the tongue very well, and could benefit from a re-wording.

ciaranRoche:
@austinmartinh have you finished work on this PR?

@grdryn are you happy with the comment changes?

austinmartinh:
@ciaranRoche Finished assuming @grdryn is happy enough with the changes

grdryn:
@ciaranRoche I can't approve, you'll need to do that

PR #97, opened 2020-01-28
Create new Alert (PrometheusRule) AWS resources being available

R-Lawton:
## Overview

Jira: https://issues.redhat.com/browse/INTLY-4668

## Prerequisites:
Have the application monitoring stack deployed
following theses steps https://github.com/integr8ly/application-monitoring-operator

## Verification

- Clone this branch
- Create a docker image from this branch
- Tag and push to Quay
- Setup the service account using the following `make setup/service_account`
- Update the image in the deployment at deploy/operator.yaml
- `oc apply -f deploy/operator.yaml`

### Redis
- create a instance of redis by using the following `make cluster/seed/managed/redis`
- Wait for redis instance to become available
- Check a prometheus rule is created in the cro namespace
- Check that the alert is visible in the prometheus dashboard.

### RDS (Postgres)
- create a instance of postgres by using the following `make cluster/seed/managed/postgres`
- Wait for postgres instance to become available
- Check a prometheus rule is created in the cro namespace
- Check that the alert is visible in the prometheus dashboard.

Screenshot from 2020-01-29 16-11-04
Screenshot from 2020-01-29 16-09-38


pkg/providers/aws/provider_redis.go
```diff
@@ -168,6 +173,21 @@ func (p *AWSRedisProvider) createElasticacheCluster(ctx context.Context, r *v1al
 		return nil, croType.StatusMessage(fmt.Sprintf("createReplicationGroup() in progress, current aws elasticache status is %s", *foundCache.Status)), nil
 	}
 
+	namespace := r.Namespace
```
ciaranRoche: Do we need to declare these variables, as they are only being used once. It might be better to just pass them directly into the the function. 

I know it reads a bit better as is, but I always lean towards brevity in go.

pkg/providers/aws/provider_redis.go
```diff
@@ -168,6 +173,21 @@ func (p *AWSRedisProvider) createElasticacheCluster(ctx context.Context, r *v1al
...
+	resourceID := r.Name
+	instanceID := *foundCache.ReplicationGroupId
+	clusterID, err := resources.GetClusterID(ctx, p.Client)
+	if err != nil {
+		errMsg := "failed to retrieve cluster identifier"
+		return nil, croType.StatusMessage(errMsg), errorUtil.Wrap(err, errMsg)
+	}
+	// Create the PrometheusRule alert to watch for the availability of this ElastiCache instance we are provisioning
+	err = p.CreateElastiCacheAvailabilityAlert(ctx, resourceID, namespace, instanceID, clusterID)
```
ciaranRoche: It might be easier to pass the `Redis` type directly, or are we keeping it as two strings to keep this function generic and re-use for an RDS availability alert

pkg/providers/aws/provider_redis.go
```diff
@@ -587,3 +616,56 @@ func (p *AWSRedisProvider) setRedisServiceMaintenanceMetric(ctx context.Context,
...
+	if err != nil {
+		return err
+	}
+
+	// Unless it already exists, call the kubernetes api and create this PrometheusRule
+	// Replace this with CreateOrUpdate if we can figure it out
+	err = p.Client.Create(ctx, pr)
+	if err != nil {
+		if !kerrors.IsAlreadyExists(err) {
+			msg := fmt.Sprintf("exception calling Create metricName: %s", alertRuleName)
```
ciaranRoche: as we are only using the string once, I think we can return the `Sprintf` directly in the error

pkg/providers/aws/provider_redis.go
```diff
@@ -370,6 +390,15 @@ func (p *AWSRedisProvider) deleteElasticacheCluster(cacheSvc elasticacheiface.El
 		return croType.StatusMessage(fmt.Sprintf("delete detected, deleteReplicationGroup() in progress, current aws elasticache status is %s", *foundCache.Status)), nil
 	}
 
+	namespace := r.Namespace
```
ciaranRoche: Same as above, I would lean toward brevity here and not declare these variables and pass them directly to the functions

Makefile
```diff
@@ -70,16 +70,16 @@ code/audit:
...
-	oc create -f ./deploy/crds/integreatly_v1alpha1_smtpcredentialset_crd.yaml -n $(NAMESPACE)
-	oc create -f ./deploy/crds/integreatly_v1alpha1_redis_crd.yaml -n $(NAMESPACE)
-	oc create -f ./deploy/crds/integreatly_v1alpha1_postgres_crd.yaml -n $(NAMESPACE)
-	oc create -f ./deploy/crds/integreatly_v1alpha1_redissnapshot_crd.yaml -n $(NAMESPACE)
-	oc create -f ./deploy/crds/integreatly_v1alpha1_postgressnapshot_crd.yaml -n $(NAMESPACE)
-	oc create -f ./deploy/service_account.yaml -n $(NAMESPACE)
-	oc create -f ./deploy/role.yaml -n $(NAMESPACE)
-	oc create -f ./deploy/role_binding.yaml -n $(NAMESPACE)
-	oc create -f ./deploy/examples/ -n $(NAMESPACE)
+	oc apply -f ./deploy/crds/integreatly_v1alpha1_blobstorage_crd.yaml -n $(NAMESPACE)
```
ciaranRoche: I like this change :+1: much cleaner if running the target on a not so clean cluster.

pkg/providers/aws/provider_redis.go
davidkirwan: In this case we can just pass the CR yeah, this function is unique to creating the availability alerts for redis.

pkg/providers/aws/provider_postgres.go
```diff
@@ -674,3 +696,55 @@ func (p *PostgresProvider) setPostgresServiceMaintenanceMetric(ctx context.Conte
...
+	}
+	p.Logger.Info(fmt.Sprintf("PrometheusRule: %s reconcilced successfully.", pr.Name))
+	return nil
+}
+
+// DeleteRDSAvailabilityAlert We call this when we delete an ElastiCache instance,
+// it removes the prometheusrule alert which watches for the availability of the instance.
+func (p *PostgresProvider) DeleteRDSAvailabilityAlert(ctx context.Context, namespace string, instanceID string) error {
+	// query the kubernetes api to find the object we're looking for
+	alertRuleName := fmt.Sprintf("cro-aws-rds-%s", instanceID)
```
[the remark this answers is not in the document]
davidkirwan: Yeah this would be the ideal.. will take a look in the morning and see if we can make this more generic, and then move it alongside the metrics exposing functions..

pkg/providers/aws/provider_redis_test.go
```diff
@@ -45,6 +52,19 @@ type mockStsClient struct {
...
+	err := croApis.AddToScheme(scheme)
+	err = openshiftConfv1.AddToScheme(scheme)
+	err = corev1.AddToScheme(scheme)
+	err = cloudCredentialApis.AddToScheme(scheme)
+	err = monitoringv1.AddToScheme(scheme)
+	if err != nil {
+		return nil, err
+	}
+	return scheme, nil
+}
```
dimitraz: Is this scheme any different to the postgres one? If not, we could also just export this test scheme and use it across both files

pkg/providers/aws/provider_postgres.go
```diff
@@ -674,3 +697,59 @@ func (p *PostgresProvider) setPostgresServiceMaintenanceMetric(ctx context.Conte
...
+	ruleName := fmt.Sprintf("availability-rule-%s", instanceID)
+
+	pr := &prometheusv1.PrometheusRule{}
+	selector := client.ObjectKey{
+		Namespace: namespace,
+		Name:      ruleName,
+	}
+
+	if err := p.Client.Get(ctx, selector, pr); err != nil {
+		return errorUtil.Wrap(err, fmt.Sprintf("exception calling DeleteRDSAvailabilityAlert: %s", ruleName))
```
dimitraz: Can also use `errorUtil.Wrapf`

pkg/providers/aws/provider_redis_test.go
davidkirwan: We could refactor to have a scheme that works for all providers, but it might be outside the scope of this particular PR, maybe keep it in the back of our minds to deal with at a later stage when we have time to address all the technical debt :laughing: 

Until then, I'd prefer to have a scheme specific to each provider.

pkg/providers/aws/provider_postgres.go
R-Lawton: That change has been made

dimitraz: :+1:

aidenkeating:
/lgtm

This seems like a fine foundation for future work

@davidkirwan @R-Lawton Can we update the Jira issue with details on the reconcile time and the longest possible time between a service going down and when we would detect it, and ping Craig Robinson and Alan Moran on it to get there input?

aidenkeating:
/approve

Tried out it on your cluster, this looks good. Let's merge it now and iterate upon this work if needed in later PRs.

PR #103, opened 2020-02-04
Targetting exported_namespace in metric alerts

davidkirwan:
Fixed issue with creating servicemonitors in the namespace the operaor is running in
Targetting exported_namespace in metric alerts

## Overview
The servicemonitor is now being created for the operator metrics service when the watch namespace is different to the namespace in which the CRO is running.
The alerts which are created for availability now correctly target `exported_namespace` rather than `namespace`

## Verification

- Clone this branch
- `oc create namespace "namespace-a"`
- `oc create namespace "namespace-b"`
- Run `make cluster/prepare`
- Update the deployment for the CRO, change the:
```
env:
  - name: WATCH_NAMESPACE
    value: "namespace-b"
```
- `oc apply -f deploy/operator.yaml --namespace "namespace-a"`
- Ensure that a servicemonitor is created in the namespace where the CRO is running

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -143,10 +143,15 @@ func main() {
 		log.Info("Could not create metrics Service", "error", err.Error())
 	}
 
+	// Retrieve the namespace the operator is running in
+	operatorNs, err := k8sutil.GetOperatorNamespace()
+	if err != nil {
+		log.Error(err, "Failed to get operator namespace")
+	}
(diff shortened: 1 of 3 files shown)


ciaranRoche:
:eyes:

aidenkeating:
/approve we covered the operator namespace part of this yesterday. it makes sense why this change needs to be made.

exported namespace also makes sense due to the conflict we discussed

PR #111, opened 2020-02-12
fix region reconciliation in aws service providers

aidenkeating:
currently aws regions are not being set in a way which would
benefit rhmi, instead defaulting to eu-west-1 for aws deploys.

this change sets a priority to different ways of retrieving the
aws region in aws providers:
- try from the configmap strategy 'region' key
- try from the 'cluster' infrastructure cr in openshift
- error

this results in there being no more hardcoded region in the
operator, which was used heavily during development. and ensures
that, unless explicitly overridden with a configmap, rhmi services
will attempt to be provisioned in the same region as the openshift
cluster that the operator is running in.

verification:
- run the operator against a non-eu-west-1 region cluster
- ensure services come up as expected

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,6 @@ require (
 	github.com/lib/pq v1.2.0
 	github.com/onsi/ginkgo v1.10.1 // indirect
 	github.com/onsi/gomega v1.7.0 // indirect
-	github.com/opencontainers/go-digest v1.0.0-rc1 // indirect
 	github.com/openshift/api v3.9.1-0.20190424152011-77b8897ec79a+incompatible
 	github.com/openshift/cloud-credential-operator v0.0.0-20190812222907-ec6f38d73a79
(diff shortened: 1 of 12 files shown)


aidenkeating:
/cc @ciaranRoche @dimitraz 

I've made as few changes as possible here.

However, I think there's a need for a major refactor of this work. Especially around:
- [Treating the strategy configs as immutable instead of overriding aspects of them](https://github.com/integr8ly/cloud-resource-operator/blob/040d85719598c6d526deefc83c5dca721c5a957b/pkg/providers/aws/provider_blobstorage.go#L469)
- Moving to an `InfrastructureClient` interface when interacting with the OpenShift infrastructure resource, instead of using standalone functions
- Moving AWS session creation, from a strategy config, to a single location and making it mockable

I'll put these in existing tickets or create new tickets. However I just wanted to make it clear that I'm aware I'm not removing technical debt in this PR.

PR #115, opened 2020-02-19
fix handle modify aws instance

ciaranRoche:
## Overview

Jira: https://issues.redhat.com/browse/INTLY-5467

Currently how we handled the modify of aws resources meant the status of a CR got stuck in `pending` until the next maintenance window. This change ensures a continuous happy path over applying modifications to resources.  

## Verification

- Clone this branch
- Run `make cluster/prepare cluster/seed/managed/redis cluster/seed/managed/postgres`
- Run `make run`
- Once resources are created, via the aws console bump the instance size of both `elasticache` and `rds`. Ensure you select apply immediately. When AWS successfully changes the instance size and CRO reconciles on these changes it will try to modify the instances to their original size. 
- Via the AWS console you can view pending modifications. Ensure these modifications are present 
- Ensure the operator continues to run, note the logs informing of any pending modifications
- Ensure the CR status reflects the AWS instance status

## Note
Currently for `postgres` we check to see if the modifications we are going to apply will be applied in the next maintenance window. If not they are added. Due to limitation in the AWS elasticache SDK we can not do this for Redis. Resulting in the `modify` function being called every reconcile until the maintenance window

### DIFF
--- a/pkg/resources/configmaps.go
+++ b/pkg/resources/configmaps.go
@@ -2,6 +2,8 @@ package resources
 
 import (
 	"context"
+	"fmt"
+	"github.com/sirupsen/logrus"
 
 	errorUtil "github.com/pkg/errors"
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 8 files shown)


pkg/resources/configmaps.go
```diff
@@ -14,6 +16,7 @@ func GetConfigMapOrDefault(ctx context.Context, c client.Client, name types.Name
 	cm := &v1.ConfigMap{}
 	if err := c.Get(ctx, name, cm); err != nil {
 		if errors.IsNotFound(err) {
+			logrus.Debug(fmt.Sprintf("%s config not found in ns ( %s ) falling back to default strategy", name.Name, name.Namespace))
```
dimitraz: `logrus.Debugf`?

ciaranRoche: Ill catch that in the release PR, just to speed up `prow`

dimitraz: Cool np

PR #136, opened 2020-05-13
[INTLY-7222] Create a skeleton controller for metrics reconciliation in CRO

dimitraz:
## Overview

Work in progress 
Jira: https://issues.redhat.com/browse/INTLY-7222

## Verification

- Clone this branch
- Run `make cluster/prepare`
- Run `make run`
- Ensure that the new controller is registered in the operator logs 
- Ensure that the controller is being reconciled every 5 minutes


aidenkeating:
👀

pkg/controller/cloudwatchmetrics/cloudwatchmetrics_controller.go
```diff
@@ -0,0 +1,111 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+var log = logf.Log.WithName("controller_cloudwatchmetrics")
+
+// Set the reconcile duration for this controller.
+// Currently it will be called once every 5 minutes
+const watchDuration = 5
```
aidenkeating, line 23: Can we replace this with seconds, just to allow for more accuracy

pkg/controller/cloudwatchmetrics/cloudwatchmetrics_controller.go
```diff
@@ -0,0 +1,111 @@
+package cloudwatchmetrics
```
aidenkeating, line 1: Can we rename this to cloud resource metrics, or cloud metrics. To allow for it to be provider-independent.

We'll likely end up using this for GCP and Azure too.

pkg/controller/cloudwatchmetrics/cloudwatchmetrics_controller.go
```diff
@@ -0,0 +1,111 @@
...
+	c, err := controller.New("cloudwatchmetrics-controller", mgr, controller.Options{Reconciler: r})
+	if err != nil {
+		return err
+	}
+
+	// Send a generic event to the channel to kick
+	// off the first reconcile
+	events := make(chan event.GenericEvent)
+	go func() {
+		events <- event.GenericEvent{
```
aidenkeating, line 48: Could we document at the top of this file the approach we're taking here.

- That this controller is for reconciling metrics for every resource (currently only redis and postgres)
- That this controller takes a sync the world approach
- That this controller will reconcile every set period of time

To make sure it's clear for future readers of the code

dimitraz: Sure

pkg/controller/cloudwatchmetrics/cloudwatchmetrics_controller.go
dimitraz: Will do

pkg/controller/cloudwatchmetrics/cloudwatchmetrics_controller.go
dimitraz: Ok :+1:

PR #137, opened 2020-05-14
add phase status metric to aws resources

austincunningham:
## Overview

Jira: https://issues.redhat.com/browse/INTLY-6521

## Verification
BYOC required
- Run ```make cluster/prepare```
- Run ```make cluster/seed/managed/postgres cluster/seed/managed/redis cluster/seed/managed/blobstorage```
- Run ```make run/local```
- Wait for the redis and postgres to reconcile 
- create snapshots from the following cr for redis and posgres 
```yaml
kind: RedisSnapshot
metadata:
  name: example
  namespace: cloud-resource-operator
spec: 
  resourceName: example-redis

```
```yaml
kind: PostgresSnapshot
metadata:
  name: example
  namespace: cloud-resource-operator
spec: 
  resourceName: example-postgres
```
- Run ```curl localhost:8383/metrics | grep aws```
- Ensure the following metrics are there 
  - redhat_rhmi_postgres_status_phase 
  - redhat_rhmi_postgres_snapshot_status_phase
  - redhat_rhmi_blobstorage_status_phase
  - redhat_rhmi_redis_status_phase
  - redhat_rhmi_redis_snapshot_status_phase

### DIFF
--- a/pkg/controller/postgressnapshot/postgressnapshot_controller.go
+++ b/pkg/controller/postgressnapshot/postgressnapshot_controller.go
@@ -246,16 +246,7 @@ func (r *ReconcilePostgresSnapshot) createSnapshot(ctx context.Context, rdsSvc r
 	return croType.PhaseInProgress, croType.StatusMessage(msg), nil
 }
 
-func buildPostgresSnapshotInfoMetricLabels(cr *integreatlyv1alpha1.PostgresSnapshot, clusterID, snapshotName string) map[string]string {
-	labels := buildPostgresSnapshotGenericMetricLabels(cr, clusterID, snapshotName)
-	if len(string(cr.Status.Phase)) != 0 {
-		labels["statusPhase"] = string(cr.Status.Phase)
-		return labels
-	}
-	return labels
-}
-
-func buildPostgresSnapshotGenericMetricLabels(cr *integreatlyv1alpha1.PostgresSnapshot, clusterID, snapshotName string) map[string]string {
+func buildPostgresSnapshotStatusMetricLabels(cr *integreatlyv1alpha1.PostgresSnapshot, clusterID, snapshotName string) map[string]string {
 	labels := map[string]string{}
 	labels["clusterID"] = clusterID
 	labels["resourceID"] = cr.Name
@@ -263,7 +254,11 @@ func buildPostgresSnapshotGenericMetricLabels(cr *integreatlyv1alpha1.PostgresSn
 	labels["instanceID"] = snapshotName
 	labels["productName"] = cr.Labels["productName"]
 	labels["strategy"] = postgresProviderName
-	labels["statusPhase"] = ""
+	if len(string(cr.Status.Phase)) != 0 {
+		labels["statusPhase"] = string(cr.Status.Phase)
+		return labels
+	}
+	labels["statusPhase"] = "nil"
 	return labels
 }
 
@@ -279,18 +274,13 @@ func (r *ReconcilePostgresSnapshot) exposePostgresSnapshotMetrics(ctx context.Co
 		return
 	}
 
-	// build metric labels
-	infoLabels := buildPostgresSnapshotInfoMetricLabels(cr, clusterID, snapshotName)
-	// build generic labels
-	genericLabels := buildPostgresSnapshotGenericMetricLabels(cr, clusterID, snapshotName)
-
-	// set status gauge
-	resources.SetMetricCurrentTime(resources.DefaultPostgresSnapshotMetricName, infoLabels)
+	// build status metric labels
+	statusLables := buildPostgresSnapshotStatusMetricLabels(cr, clusterID, snapshotName)
 
 	// set available metric
 	if len(string(cr.Status.Phase)) == 0 || cr.Status.Phase != croType.PhaseComplete {
-		resources.SetMetric(resources.DefaultPostgresSnapshotMetricName, genericLabels, 0)
+		resources.SetMetric(resources.DefaultPostgresSnapshotStatusMetricName, statusLables, 0)
 		return
 	}
-	resources.SetMetric(resources.DefaultPostgresSnapshotMetricName, infoLabels, 1)
+	resources.SetMetric(resources.DefaultPostgresSnapshotStatusMetricName, statusLables, 1)
 }
--- a/pkg/controller/postgressnapshot/postgressnapshot_controller.go
+++ b/pkg/controller/postgressnapshot/postgressnapshot_controller.go
@@ -112,6 +112,9 @@ func (r *ReconcilePostgresSnapshot) Reconcile(request reconcile.Request) (reconc
 		return reconcile.Result{}, err
 	}
 
+	// set info metric
+	defer r.exposePostgresSnapshotMetrics(ctx, instance)
+
 	// check status, if complete return
 	if instance.Status.Phase == croType.PhaseComplete {
 		r.logger.Infof("skipping creation of snapshot for %s as phase is complete", instance.Name)
@@ -194,9 +197,6 @@ func (r *ReconcilePostgresSnapshot) createSnapshot(ctx context.Context, rdsSvc r
 	// update cr with snapshot name
 	snapshot.Status.SnapshotID = snapshotName
 
-	// set info metric
-	defer r.exposePostgresSnapshotMetrics(ctx, snapshot)
-
 	if err = r.client.Status().Update(ctx, snapshot); err != nil {
 		errMsg := fmt.Sprintf("failed to update instance %s in namespace %s", snapshot.Name, snapshot.Namespace)
 		return croType.PhaseFailed, croType.StatusMessage(errMsg), errorUtil.Wrap(err, errMsg)
@@ -246,14 +246,12 @@ func (r *ReconcilePostgresSnapshot) createSnapshot(ctx context.Context, rdsSvc r
 	return croType.PhaseInProgress, croType.StatusMessage(msg), nil
 }
 
-func buildPostgresSnapshotInfoMetricLabels(cr *integreatlyv1alpha1.PostgresSnapshot, clusterID, bucketName string) map[string]string {
-	labels := buildPostgresSnapshotGenericMetricLabels(cr, clusterID, bucketName)
+func buildPostgresSnapshotInfoMetricLabels(cr *integreatlyv1alpha1.PostgresSnapshot, clusterID, snapshotName string) map[string]string {
+	labels := buildPostgresSnapshotGenericMetricLabels(cr, clusterID, snapshotName)
 	if len(string(cr.Status.Phase)) != 0 {
 		labels["statusPhase"] = string(cr.Status.Phase)
 		return labels
 	}
-	// If the status hasn't reconciled using cr.Status.Phase need to return something
-	labels["statusPhase"] = "nil"
 	return labels
 }
 
@@ -265,6 +263,7 @@ func buildPostgresSnapshotGenericMetricLabels(cr *integreatlyv1alpha1.PostgresSn
 	labels["instanceID"] = snapshotName
 	labels["productName"] = cr.Labels["productName"]
 	labels["strategy"] = postgresProviderName
+	labels["statusPhase"] = ""
 	return labels
 }
 
@@ -282,13 +281,15 @@ func (r *ReconcilePostgresSnapshot) exposePostgresSnapshotMetrics(ctx context.Co
 
 	// build metric labels
 	infoLabels := buildPostgresSnapshotInfoMetricLabels(cr, clusterID, snapshotName)
+	// build generic labels
+	genericLabels := buildPostgresSnapshotGenericMetricLabels(cr, clusterID, snapshotName)
 
 	// set status gauge
 	resources.SetMetricCurrentTime(resources.DefaultPostgresSnapshotMetricName, infoLabels)
 
 	// set available metric
 	if len(string(cr.Status.Phase)) == 0 || cr.Status.Phase != croType.PhaseComplete {
-		resources.SetMetric(resources.DefaultPostgresSnapshotMetricName, infoLabels, 0)
+		resources.SetMetric(resources.DefaultPostgresSnapshotMetricName, genericLabels, 0)
 		return
 	}
 	resources.SetMetric(resources.DefaultPostgresSnapshotMetricName, infoLabels, 1)
--- a/pkg/controller/postgressnapshot/postgressnapshot_controller.go
+++ b/pkg/controller/postgressnapshot/postgressnapshot_controller.go
@@ -30,6 +30,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
+const (
+	postgresProviderName = "aws-rds"
+)
+
 // Add creates a new PostgresSnapshot Controller and adds it to the Manager. The Manager will set fields on the Controller
 // and Start it when the Manager is Started.
 func Add(mgr manager.Manager) error {
@@ -189,6 +193,10 @@ func (r *ReconcilePostgresSnapshot) createSnapshot(ctx context.Context, rdsSvc r
 
 	// update cr with snapshot name
 	snapshot.Status.SnapshotID = snapshotName
+
+	// set info metric
+	defer r.exposePostgresSnapshotMetrics(ctx, snapshot)
+
 	if err = r.client.Status().Update(ctx, snapshot); err != nil {
 		errMsg := fmt.Sprintf("failed to update instance %s in namespace %s", snapshot.Name, snapshot.Namespace)
 		return croType.PhaseFailed, croType.StatusMessage(errMsg), errorUtil.Wrap(err, errMsg)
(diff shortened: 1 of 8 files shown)


pkg/providers/aws/provider_blobstorage.go
```diff
@@ -550,3 +554,55 @@ func (p *BlobStorageProvider) getS3BucketConfig(ctx context.Context, bs *v1alpha
 func buildEndUserCredentialsNameFromBucket(b string) string {
 	return fmt.Sprintf("cro-aws-s3-%s-creds", b)
 }
+
+func buildBlobStorageInfoMetricLabels(cr *v1alpha1.BlobStorage, clusterID, bucketName string) map[string]string {
```
dimitraz, line 558: Given that a lot of these functions are being replicated across all of the aws providers and snapshot controllers, we could possibly extract them into some util functions? Just a thought, it might be overkill. Wdyt?

austincunningham: Your right its a little dry at the moment. Could move it all to the metrics util. Currently this is not working as expected as the status are not returning correctly.

austincunningham: After a second look it may be too closely coupled to the individual providers and controllers, so when decoupling a lot of the logic would end up living in the provider and controllers.

pkg/providers/aws/provider_postgres.go
```diff
@@ -763,11 +763,11 @@ func (p *PostgresProvider) configureRDSVpc(ctx context.Context, rdsSvc rdsiface.
...
 func buildPostgresInfoMetricLabels(cr *v1alpha1.Postgres, instance *rds.DBInstance, clusterID, instanceName string) map[string]string {
 	labels := buildPostgresGenericMetricLabels(cr, clusterID, instanceName)
-	if instance != nil {
-		labels["status"] = *instance.DBInstanceStatus
+	if instance != nil && len(string(cr.Status.Phase)) != 0 {
+		labels["statusAWS"] = *instance.DBInstanceStatus
+		labels["statusPhase"] = string(cr.Status.Phase)
 		return labels
 	}
-	labels["status"] = "nil"
```
dimitraz, line 770: Do we not want to keep this to report back that the instance is nil?

pkg/controller/postgressnapshot/postgressnapshot_controller.go
```diff
@@ -237,3 +245,52 @@ func (r *ReconcilePostgresSnapshot) createSnapshot(ctx context.Context, rdsSvc r
...
+
+func buildPostgresSnapshotGenericMetricLabels(cr *integreatlyv1alpha1.PostgresSnapshot, clusterID, snapshotName string) map[string]string {
+	labels := map[string]string{}
+	labels["clusterID"] = clusterID
+	labels["resourceID"] = cr.Name
+	labels["namespace"] = cr.Namespace
+	labels["instanceID"] = snapshotName
+	labels["productName"] = cr.Labels["productName"]
+	labels["strategy"] = postgresProviderName
+	labels["statusPhase"] = ""
```
dimitraz, line 266: Do we need `statusPhase` here in the generic labels? Technically these should always stay generic

pkg/controller/postgressnapshot/postgressnapshot_controller.go
```diff
@@ -237,3 +245,52 @@ func (r *ReconcilePostgresSnapshot) createSnapshot(ctx context.Context, rdsSvc r
...
+	clusterID, err := resources.GetClusterID(ctx, r.client)
+	if err != nil {
+		logrus.Errorf("failed to get cluster id while exposing information metric for %v", snapshotName)
+		return
+	}
+
+	// build metric labels
+	infoLabels := buildPostgresSnapshotInfoMetricLabels(cr, clusterID, snapshotName)
+	// build generic labels
+	genericLabels := buildPostgresSnapshotGenericMetricLabels(cr, clusterID, snapshotName)
```
dimitraz, line 285: Do we need both of these functions? I think these could probably get merged into one since we don't have 2 separate metrics to label

pkg/controller/postgressnapshot/postgressnapshot_controller.go
```diff
@@ -237,3 +245,52 @@ func (r *ReconcilePostgresSnapshot) createSnapshot(ctx context.Context, rdsSvc r
...
+
+	// set status gauge
+	resources.SetMetricCurrentTime(resources.DefaultPostgresSnapshotMetricName, infoLabels)
+
+	// set available metric
+	if len(string(cr.Status.Phase)) == 0 || cr.Status.Phase != croType.PhaseComplete {
+		resources.SetMetric(resources.DefaultPostgresSnapshotMetricName, genericLabels, 0)
+		return
+	}
+	resources.SetMetric(resources.DefaultPostgresSnapshotMetricName, infoLabels, 1)
```
dimitraz, line 295: Because this metric is already being set on line `288` with a timestamp, I don't think we need this block here to set the values again to `0` or `1`

pkg/controller/redissnapshot/redissnapshot_controller.go
```diff
@@ -261,3 +269,53 @@ func (r *ReconcileRedisSnapshot) createSnapshot(ctx context.Context, cacheSvc el
...
+	}
+
+	// build metric labels
+	infoLabels := buildRedisSnapshotInfoMetricLabels(cr, clusterID, snapshotName)
+
+	// build generic labels
+	genericLabels := buildRedisSnapshotGenericMetricLabels(cr, clusterID, snapshotName)
+
+	// set status gauge
+	resources.SetMetricCurrentTime(resources.DefaultRedisSnapshotMetricName, infoLabels)
```
dimitraz, line 313: I guess the same changed can be applied here as with the postgres controller. We can label this metric with a merged map of `infoLabels` and `genericLabels`, and remove lines `315-320` underneath

pkg/providers/aws/provider_postgres.go
```diff
@@ -779,6 +779,8 @@ func buildPostgresGenericMetricLabels(cr *v1alpha1.Postgres, clusterID, instance
 	labels["instanceID"] = instanceName
 	labels["productName"] = cr.Labels["productName"]
 	labels["strategy"] = postgresProviderName
+	labels["statusAWS"] = ""
+	labels["statusPhase"] = ""
```
dimitraz, line 783: These can be removed as the generic labels should never change

pkg/providers/aws/provider_postgres.go
```diff
@@ -806,11 +808,11 @@ func (p *PostgresProvider) exposePostgresMetrics(ctx context.Context, cr *v1alph
 	resources.SetMetricCurrentTime(resources.DefaultPostgresInfoMetricName, infoLabels)
 
 	// set available metric
-	if instance == nil || *instance.DBInstanceStatus != "available" {
+	if len(string(cr.Status.Phase)) == 0 || cr.Status.Phase != croType.PhaseComplete {
```
dimitraz, line 811: Since this is an `available` metric I think we need to keep the check on line `809`. 
We should be checking the instance for this metric, and not the phase of the cr, so I don't think this should have to change

pkg/providers/aws/provider_postgres.go
```diff
@@ -806,11 +808,11 @@ func (p *PostgresProvider) exposePostgresMetrics(ctx context.Context, cr *v1alph
...
 		resources.SetMetric(resources.DefaultPostgresAvailMetricName, genericLabels, 0)
 		return
 	}
-	resources.SetMetric(resources.DefaultPostgresAvailMetricName, genericLabels, 1)
+	resources.SetMetric(resources.DefaultPostgresAvailMetricName, infoLabels, 1)
```
dimitraz, line 815: I don't think this should change, the availablity metric should be left as is with its generic labels

pkg/providers/aws/provider_redis.go
```diff
@@ -620,17 +620,19 @@ func buildRedisGenericMetricLabels(r *v1alpha1.Redis, clusterID, cacheName strin
...
 func buildRedisInfoMetricLables(r *v1alpha1.Redis, group *elasticache.ReplicationGroup, clusterID, cacheName string) map[string]string {
 	labels := buildRedisGenericMetricLabels(r, clusterID, cacheName)
-	if group != nil {
-		labels["status"] = *group.Status
+	if group != nil && len(string(r.Status.Phase)) != 0 {
+		labels["statusAWS"] = *group.Status
+		labels["statusPhase"] = string(r.Status.Phase)
 		return labels
 	}
-	labels["status"] = fmt.Sprintf("%s is nil", cacheName)
```
dimitraz, line 633: Same questions/comments here as with the postgres provider

pkg/controller/postgressnapshot/postgressnapshot_controller.go
austincunningham: yes its needed here. run into issues with cardinality if not present when you attempt to add statusPhase later in buildPostgresSnapshotInfoMetricLabels

pkg/providers/aws/provider_postgres.go
austincunningham: already populated on https://github.com/integr8ly/cloud-resource-operator/pull/137/files#diff-8fd2a41e86b5db36e6443ed519133f75L765

pkg/providers/aws/provider_postgres.go
austincunningham: run in to cardinality issue when building info metrics https://github.com/integr8ly/cloud-resource-operator/pull/137/files#diff-8fd2a41e86b5db36e6443ed519133f75L765

pkg/providers/aws/provider_postgres.go
austincunningham: maybe ? Only really getting a good look a the metrics now that I have them applied in the monitoring stack.

pkg/controller/postgressnapshot/postgressnapshot_controller.go
austincunningham: we have 2 separate metrics

pkg/controller/postgressnapshot/postgressnapshot_controller.go
austincunningham: We do need to handle the status if it hasn't been reconciled and status doesn't exist. maybe should be passing genericLabels on the last call to SetMetrics.

pkg/providers/aws/provider_postgres.go
dimitraz: Oh right, what do you mean by cardinality issue? 

From what I've seen those `genericLabels` functions are to create static labels that do not ever change after the fact, and then only when calling the `infoLabels` functions do we add additional information such as the status

pkg/providers/aws/provider_postgres.go
austincunningham: `available` was implemented but never used in alerts see https://issues.redhat.com/browse/INTLY-6521?focusedCommentId=14095362&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14095362, as this jira was to check for the phase and not the aws Status I decided I would reuse this function.

pkg/providers/aws/provider_postgres.go
austincunningham: cardinality issue applying 7 labels to 6,

pkg/providers/aws/provider_postgres.go
ciaranRoche: the `available` metric is used and alerted on https://github.com/integr8ly/integreatly-operator/blob/master/pkg/resources/metrics.go#L26
so is the `connectivity` metric https://github.com/integr8ly/integreatly-operator/blob/master/pkg/resources/metrics.go#L55
I meant the `info` metric was never used in alerts https://github.com/integr8ly/cloud-resource-operator/blob/master/pkg/providers/aws/provider_postgres.go#L806 

I think for postgres and redis providers all we need to do is update the status to have both `awsStatus` and `CR phase Status` in the info labels https://github.com/integr8ly/cloud-resource-operator/blob/master/pkg/providers/aws/provider_postgres.go#L764

aidenkeating:
@austincunningham I haven't tried this out, but it seems to be coming along well. I'd like to get your opinion on two items:

- Currently we prefix metrics with `cro_`, do you think it'd be much effort to change this to `redhat_rhmi_`? If it seems like it would be anything other than trivial, we can create a follow up ticket

- I've been reading about Kube State Metrics. They expose elements in the `.status` block as separate metrics. So `.status.phase` on a resource called `Redis` would be called `<prefix>_redis_status_phase`. I don't think it's major, but do you think this is something we could do here? Then we'd have a metric `cro_redis_status_phase{redis="myredis", namespace="redhat-rhmi-operator", phase="failed"}=<1 if "complete" else 0>`

It adds some complexity to this task, but will leave us more inline with the Kube State Metrics approach. See `kube_pod_status_phase` in [1] for an example

[1] https://github.com/kubernetes/kube-state-metrics/blob/master/docs/pod-metrics.md

austincunningham:
@aidenkeating 
- metrics name are all vars that can be easily change here https://github.com/integr8ly/cloud-resource-operator/blob/master/pkg/resources/custom_metrics.go#L22L29
- In the middle of a rewrite anyway should be too hard too create a complete separate metric for phase.

aidenkeating:
@austincunningham Awesome, great news, thanks!  So just to have a clear understanding in my head:

- We'll have all older info in the `_info` metric, but the `cro_` prefix will be replaced with `redhat_rhmi_`
- We'll have a new metric for the `.status.phase` element with the format `redhat_rhmi_<resource>_status_phase` with `<resource>`, `namespace` and `phase` labels. With a value set to `1` when `.status.phase == complete` and `0` otherwise.

Technically we only need that new status phase metric for postgres and redis for now. But I'll leave it up to you to decide if adding in the others is simple enough.

pkg/controller/postgressnapshot/postgressnapshot_controller.go
austincunningham: your right only one at this point, and will refactor to status metric.

pkg/controller/postgressnapshot/postgressnapshot_controller.go
austincunningham: shouldn't be setting this metric with timestamp, should only be 1 or 0 . will remove that line.

pkg/controller/redissnapshot/redissnapshot_controller.go
austincunningham: same as postgressnapshot , metric should be 0 or 1, will remove the time stamp.

pkg/controller/postgressnapshot/postgressnapshot_controller.go
austincunningham: no your right don't need it here.

pkg/providers/aws/provider_postgres.go
austincunningham: moving phase status to an new function and restoring this metric to original format.

pkg/providers/aws/provider_postgres.go
austincunningham: moving phase status to its own metric and restoring this to original

pkg/providers/aws/provider_redis.go
austincunningham: restoring this function and moving phase status to new function

pkg/providers/aws/provider_postgres.go
austincunningham: your right.

pkg/providers/aws/provider_postgres.go
austincunningham: Thanks @ciaranRoche see what I was missing. moving phase status to its own function

pkg/providers/aws/provider_redis.go
```diff
@@ -657,15 +665,24 @@ func (p *RedisProvider) exposeRedisMetrics(ctx context.Context, cr *v1alpha1.Red
 	// build generic metrics
 	genericLabels := buildRedisGenericMetricLabels(cr, clusterID, cacheName)
 
+	statusLables := buildRedisStatusMetricsLabels(cr, clusterID, cacheName)
```
dimitraz, line 668: Just a small typo :) `statusLabels`, think a few other files might have this too

pkg/providers/aws/provider_redis.go
```diff
@@ -620,19 +620,27 @@ func buildRedisGenericMetricLabels(r *v1alpha1.Redis, clusterID, cacheName strin
...
 // adds extra information to labels around resource
 func buildRedisInfoMetricLables(r *v1alpha1.Redis, group *elasticache.ReplicationGroup, clusterID, cacheName string) map[string]string {
 	labels := buildRedisGenericMetricLabels(r, clusterID, cacheName)
-	if group != nil && len(string(r.Status.Phase)) != 0 {
-		labels["statusAWS"] = *group.Status
+	if group != nil {
+		labels["status"] = *group.Status
+		return labels
+	}
+	labels["status"] = fmt.Sprintf("%s is nil", cacheName)
```
dimitraz: I noticed that some of these are simply `"nil"` and others are a string like `"x is nil"`, would probably be nicer to keep just one for consistency

austincunningham: agree, as the other labels already have identifiers like cacheName, will remove `x is nil` as metrics are easier to search for in Prometheus when they are consistent.

pkg/providers/aws/provider_redis.go
austincunningham: :+1:

PR #138, opened 2020-05-14
Strategy config map util function

ciaranRoche:
## Overview

Jira: https://issues.redhat.com/browse/INTLY-7203

## Verification
Verification steps outlined in https://github.com/integr8ly/integreatly-operator/pull/727


pkg/apis/integreatly/v1alpha1/zz_generated.openapi.go
```diff
@@ -442,19 +442,19 @@ func schema_pkg_apis_integreatly_v1alpha1_RedisSnapshot(ref common.ReferenceCall
 					},
 					"spec": {
 						SchemaProps: spec.SchemaProps{
-							Ref: ref("github.com/integr8ly/cloud-resource-operator/pkg/apis/integreatly/v1alpha1.RedisSnapshotSpec"),
+							Ref: ref("./pkg/apis/integreatly/v1alpha1.RedisSnapshotSpec"),
```
aidenkeating, line 445: What's the reasoning behind this change, is it just due to a different Operator SDK version?

pkg/public/strategies.go
```diff
@@ -0,0 +1,209 @@
...
+		if err != nil {
+			return errorUtil.Wrapf(err, "failed to reconcile postgres strategy")
+		}
+		redisStrategy, err := reconcileRedisStrategy(awsStratConfig.Data["redis"], tier, backupWindow, maintenanceWindow)
+		if err != nil {
+			return errorUtil.Wrapf(err, "failed to reconcile redis strategy")
+		}
+
+		// setting postgres and redis values to be updated strategies
+		awsStratConfig.Data["postgres"] = postgresStrategy
```
aidenkeating, line 70: Can we set `postgres` and `redis` as constants somewhere?

pkg/public/strategies.go
```diff
@@ -0,0 +1,209 @@
...
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"strings"
+	"time"
+)
+
+// reconciles aws strategy map, adding maintenance and backup window fields
+func reconcileAWSStrategyMap(ctx context.Context, client client.Client, tier, backupApplyOn, maintenanceApplyFrom, namespace string) error {
```
aidenkeating, line 21: Would there be an issue with expecting a time struct of some type here instead of a string for backupApplyOn and maintenanceApplyFrom?

Also, what do you think about changing the names to `backupTimeStart`, `maintenanceTimeStart` or something like that? To make it clear we're accepting the starting time and figuring out the rest

pkg/public/strategies.go
```diff
@@ -0,0 +1,209 @@
+package public
```
aidenkeating, line 1: Can we add to the top of this file the why of these functions?

- We provide a utility to abstract the various strategy map ConfigMaps from the service using CRO (RHMI)
- We accept start times because different cloud providers may have different minimum lengths for backup and maintenance

pkg/public/util.go
```diff
@@ -127,3 +129,14 @@ func ReconcileRedis(ctx context.Context, client client.Client, productName, depl
 
 	return r, nil
 }
+
+// ReconcileStrategyMaps to be used to reconcile strategy maps expected in RHMI installs
+// A single function which can check the infrastructure and provision the correct strategy config map
+func ReconcileStrategyMaps(ctx context.Context, client client.Client, tier, backupWindow, maintenanceWindow, namespace string) error {
```
aidenkeating, line 135: Is there any reason this public function isn't in `strategies.go`? They're both in the same package and it's probably a better home for it

pkg/public/util.go
```diff
@@ -1,12 +1,14 @@
-package resources
+package public
```
aidenkeating, line 1: `client` could be an alternative name to `public`, unless there's an issue with that

pkg/public/strategies.go
```diff
@@ -0,0 +1,209 @@
...
+
+	// unmarshall the create strategy from the raw strategy tier to type create db instance input
+	rdsCreateConfig := &rds.CreateDBInstanceInput{}
+	if err := json.Unmarshal(rawStrategy[tier].CreateStrategy, rdsCreateConfig); err != nil {
+		return "", errorUtil.Wrapf(err, "failed to unmarshal aws rds cluster config")
+	}
+
+	// check create db instance input against expected values
+	// backup window and maintenance window are expected values via RHMIconfig
+	// todo we may want to handle more strategy input, we should add functionality to check every value
```
aidenkeating, line 99: We do similar checks in other areas in CRO, right? Is there any way we can re-use that logic?

pkg/client/strategies.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"context"
+	errorUtil "github.com/pkg/errors"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	// exported tiers to be used by RHMI operator
+	ProductionTier = "production"
+	DevelopmentTier = "development"
```
aidenkeating, line 30: Small thing, but can we reverse these names, it'll help group them as tiers e.g.

```
TierProduction = ...
TierDevelopment = ... 
```

pkg/client/strategies.go
```diff
@@ -0,0 +1,50 @@
...
+
+	postgresStratKey    = "postgres"
+	redisStratKey       = "redis"
+	blobstorageStratKey = "blobstorage"
+)
+
+type StrategyTimeConfig struct {
+	BackupStartTime      string
+	MaintenanceStartTime string
+}
```
aidenkeating, line 40: Why do we take in strings here instead of a time?

pkg/client/strategies_aws.go
```diff
@@ -0,0 +1,266 @@
...
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"strings"
+	"time"
+)
+
+const (
+	awsStratName = "cloud-resources-aws-strategies"
```
aidenkeating, line 36: Can we use https://github.com/integr8ly/cloud-resource-operator/blob/master/pkg/providers/aws/config.go#L27 to avoid them becoming out of sync in the future

pkg/client/strategies_aws.go
```diff
@@ -0,0 +1,266 @@
...
+	return map[string]string{
+		blobstorageStratKey: buildDefaultAWSStratValue(),
+		redisStratKey:       buildDefaultAWSStratValue(),
+		postgresStratKey:    buildDefaultAWSStratValue(),
+	}
+}
+
+// builder function to provide default value used per tier in aws strategy config map
+func buildDefaultAWSStratValue() string {
+	return "{\"development\": { \"region\": \"\", \"createStrategy\": {}, \"deleteStrategy\": {} }, \"production\": { \"region\": \"\", \"createStrategy\": {}, \"deleteStrategy\": {} }}"
```
aidenkeating, line 194: We'll now have this duplicated, once in the provider and once here. If refactoring would take too much time, can we create a follow-up ticket and link it here?

aidenkeating:
/approve

## RHMI Custom Resource

Screenshot 2020-05-19 at 16 36 19

## RDS

Screenshot 2020-05-19 at 16 36 07

Screenshot 2020-05-19 at 16 35 58


## Elasticache

Screenshot 2020-05-19 at 16 37 52

PR #146, opened 2020-06-05
Add new network strategy to AWS provider ConfigMap

ciaranRoche:
## Overview

Jira: https://issues.redhat.com/browse/INTLY-8100

## Verification 
We need to verify two cases : 
1. If RHMI valid subnets exist in cluster vpc
2. If NO RHMI valid subnets exist in cluster vpc

**No Cluster VPC RHMI Subnet**
- Checkout this branch
- Run `make cluster/prepare NAMESPACE=<<your name>>-cloud-resource-operator`
- Run `make cluster/seed/managed/postgres cluster/seed/managed/redis NAMESPACE=<<your name>>-cloud-resource-operator`
- Run `make run/local NAMESPACE=<<your name>>-cloud-resource-operator`
- We should see the following error for both `Redis` and `Postgres` : 
```
# we see this error as CreateNetwork is just a stub
error reconciling elasticache networking: error reconciling elasticache
    networking: failed to create resource network: CreateNetwork stub
```

**Cluster VPC RHMI Subnets** 
- Checkout Master
- Run `make cluster/prepare NAMESPACE=<<your name>>-cloud-resource-operator`
- Run `make cluster/seed/managed/postgres cluster/seed/managed/redis NAMESPACE=<<your name>>-cloud-resource-operator`
- Run `make run/local NAMESPACE=<<your name>>-cloud-resource-operator`
- Wait for resources to provision correctly
- Log into AWS console, navigate to `elasticache`
- Click on `Redis` expand the menu next to the provisioned cluster
- Note the subnetgroup
- Run `aws elasticache describe-cache-subnet-groups --cache-subnet-group-name=<<subnet group>>`
- Note the `subnets`
- Run `aws rds describe-db-instances`
- Note the `subnet group` and `subnets` for your provisioned resource
- Stop operator and checkout this branch
- Run `make run/local NAMESPACE=<<your name>>-cloud-resource-operator`
- Ensure no errors
- Log into AWS console, navigate to `elasticache`
- Click on `Redis` expand the menu next to the provisioned cluster
- Note the subnetgroup
- Run `aws elasticache describe-cache-subnet-groups --cache-subnet-group-name=hcdrhmiclouds27bc5subnetgroup`
- Verify the subnets match the elasticache subnets noted earlier
- Run `aws rds describe-db-instances`
- Verify the subnets match the rds subnets noted earlier


pkg/providers/aws/cluster_network_provider.go
```diff
@@ -0,0 +1,87 @@
...
+IsEnabled checks for valid rhmi subnets within the cluster VPC
+a valid rhmi subnet will contain a tag with the `organizationTag` value
+
+when rhmi subnets are present in a cluster vpc it indicates that the vpc configuration
+was created in a cluster with a cluster version <= 4.4.5
+
+when rhmi subnets are absent in a cluster vpc it indicates that the vpc configuration has not been created
+and we should create a new vpc for all resources to be deployed in and we should peer the
+resource vpc and cluster vpc
+*/
```
aidenkeating, line 64: I think the why is important here, for future readers. The details can be included at the top of the file.

```suggestion
/*
IsEnabled returns true when no subnets created by the cloud resource operator exist in the openshift cluster vpc.

subnets created by the cloud resource operator are identified by having a tag with the name `<organizationTag>/clusterID`. By default, `integreatly.org/clusterID`.

this check allows us to maintain backwards compatibility with openshift clusters that used the cloud resource operator before this standalone vpc provider was added. If this function returns false, we should continue using the backwards compatible approach of bundling resources in with the openshift cluster vpc.
*/
```

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -0,0 +1,87 @@
+package aws
```
aidenkeating, line 1: I think we need to be verbose about why this exists, for future engineers. So they understand why we have a fork in our logic.

```suggestion
// utility to manage a dedicated vpc for the resources created by the cloud resource operator.
//
// this has been added to allow the operator to work with clusters provisioned with openshift tooling >= 4.4.6,
// as they will not allow multi-az cloud resources to be created in single-az openshift clusters due to the single-az
// cluster subnets taking up all networking addresses in the cluster vpc.
//
// any openshift clusters that have used the cloud resource operator before this utility was added will be using the old approach,
// which is bundling cloud resources in with the cluster vpc. backwards compatibility for this approach must be maintained.
// 
// see [1] for more details.
//
// [1] 

package aws
```

deploy/examples/cloud_resources_aws_strategies.yaml
```diff
@@ -9,3 +9,6 @@ data:
     {"development": { "region": "eu-west-1", "createStrategy": {}, "deleteStrategy": {} }}
   postgres: |
     {"development": { "region": "eu-west-1", "createStrategy": {}, "deleteStrategy": {} }}
+  _network: |
```
dimitraz, line 12: What is the reason for the underscore in front of the network name?

pkg/providers/aws/cluster_network_provider_test.go
```diff
@@ -0,0 +1,211 @@
...
+			},
+			fields: fields{
+				logger: logrus.NewEntry(logrus.StandardLogger()),
+			},
+			want:    true,
+			wantErr: false,
+		},
+		{
+
+			//we expect if a single rhmi subnet is found in cluster vpc isEnabled will return true
```
aidenkeating, line 128: Will return `false`?

pkg/providers/aws/provider_postgres.go
```diff
@@ -701,9 +705,50 @@ func buildDefaultRDSSecret(ps *v1alpha1.Postgres) *v1.Secret {
...
+reconcileRDSNetworking configures networking required for cro resources (postgres)
+
+networkManager isEnabled checks for the presence of valid RHMI subnets in the cluster vpc
+when rhmi subnets are present in a cluster vpc it indicates that the vpc configuration
+was created in a cluster with a cluster version <= 4.4.5
+
+when rhmi subnets are absent in a cluster vpc it indicates that the vpc configuration has not been created
+and we should create a new vpc for all resources to be deployed in and we should peer the
+resource vpc and cluster vpc
+*/
```
aidenkeating, line 718: I think if we bring these explanations out to individual providers, we'll be explaining it twice, once for RDS, once for Elasticache. If we just reference back to the new network provider we can keep this shorter and more maintainable.

```suggestion
/*
reconcileRDSNetworking creates networking resources (vpcs, subnets, subnet groups etc) needed to provision an rds instance.

we will use the backwards compatible networking approach on older openshift clusters, see NetworkProvider for more details.
*/
```

deploy/examples/cloud_resources_aws_strategies.yaml
aidenkeating: @dimitraz The idea here is that `_network` doesn't define settings for a resource that users can create or update. It's configurable, but it's used internally in CRO.

It also won't follow the same pattern as other keys in this configmap. We won't really reconcile on this, we'll create the VPC with the provided CIDR and that's pretty much it. If you change it, it'll be ignored. The delete strategy has just been added to allow for extensibility.

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -0,0 +1,87 @@
...
+		logger: logger.WithField("provider", "standalone_network_provider"),
+	}
+}
+
+func (n *NetworkProvider) CreateNetwork() error {
+	/*
+		todo CreateNetwork is called we need to ensure the following
+		we expect _network to exist with a valid cidr
+		the absence of either _network or valid cidr we should return with an informative message
+		re-reconcile until we have a valid _network strat and valid cidr
```
aidenkeating, line 41: Maybe add:

```
- Check if a VPC exists with the integreatly.org/clusterID tag
- If VPC doesn't exist, create a VPC with CIDR block and tag it
```

deploy/examples/cloud_resources_aws_strategies.yaml
```diff
@@ -9,3 +9,6 @@ data:
...
+    {"development": { "region": "eu-west-1", "createStrategy": {}, "deleteStrategy": {} }}
```
aidenkeating, line 13: Can we add in an example CIDR block here, for dev purposes

deploy/examples/cloud_resources_aws_strategies.yaml
dimitraz: Ok that makes sense. Would it be worth documenting this somewhere?

aidenkeating: Good point. @ciaranRoche Mind throwing a note about this into https://github.com/integr8ly/cloud-resource-operator/tree/master/doc ?

ciaranRoche: @aidenkeating keeping note of all documentation, i was thinking of holding off for a few days before capturing documentation, just in case anything changes. Ill add this to my list and follow up after make some more progress

aidenkeating: @ciaranRoche I think it's best to capture documentation when we can do it piece by piece instead of in a rush. Small refactors of documentation aren't that bad compared to documenting an entire solution at the end.

`_network` isn't going to change from being a CreateVPCInput struct.

PR #151, opened 2020-06-17
Create standalone VPC when "_network" block is defined

ciaranRoche:
Co-authored-by: laurafitzgerald <lfitzger@example.com>

## Overview

Jira: https://issues.redhat.com/browse/INTLY-8103

## Verification Consuming _network Strat Map

_Note : verification must be completed in a fresh cluster and aws account. No remnants of bundled cluster networking (cluster subnets, subnet groups, security groups) should exist in cluster vpc_

_Note : the seeded resource will not fully install, as the security group is not implemented and will cause the install to fail,  the standalone networking is provisioned before the error, the DeleteNetwork()  function is provisionally implemented, so removing the CR will clean up the standalone network. There are follow on tickets for both security group and DeleteNetwork() implementation_

- Clone this branch
- Run `make cluster/prepare`
- Run `make run`
- Make note of the `cidrBlock` value in the `_network` strategy map
- Run `aws ec2 describe-vpcs`
- The `cidrBlock` for the vpc name `RHMI Cloud Resource VPC` it should match the `cidrBlock` in the `_network` strategy map
- Make note of the vpc-id
- Run ` aws ec2 describe-subnets --filters "Name=vpc-id,Values=<<noted vpc-id>>"`   
- Ensure two subnets are returned
- Ensure the `cidrBlock` mask for each subnet is one bit greater then the `cidrBlock` mask in the `_network` strategy map
- Update the `cidrBlock` value in the `_network` strategy map     
- Recheck the `cidrBlock` value in the vpc and subnets, these should not change from the original values        

## Verification Cidr Overlap
_Note we can not have both standalone and bundled vpc cidr blocks overlapping_

- Remove CR from previous verification 
- Ensure standalone vpc, subnets and subnet groups have been cleaned
- Run `aws ec2 describe-vpcs`
- Note the `cidrBlock` on the cluster vpc, which can be found by the following tag
```
{
    "Key": "kubernetes.io/cluster/<<cluster id>>",
    "Value": "owned"
}
```
- ensure the `_network` type `cidrBlock` value in the `cloud-resources-aws-strategies` config map overlaps with the `cidrBlock` from the cluster vpc
- Run `make run`
- Ensure the following error :
```
failed to create resource network: standalone vpc creation failed: standalone cidr block 10.0.0.0/26 overlaps with cluster vpc cidr block 10.0.0.0/16, update _network strategy to continue vpc creation
```

### DIFF
--- a/pkg/providers/aws/cluster_network_provider.go
+++ b/pkg/providers/aws/cluster_network_provider.go
@@ -25,6 +25,7 @@ package aws
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
@@ -35,6 +36,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/elasticache/elasticacheiface"
 	"github.com/aws/aws-sdk-go/service/rds"
 	"github.com/aws/aws-sdk-go/service/rds/rdsiface"
+	"github.com/integr8ly/cloud-resource-operator/pkg/providers"
 	"github.com/integr8ly/cloud-resource-operator/pkg/resources"
 	"github.com/sirupsen/logrus"
 	"net"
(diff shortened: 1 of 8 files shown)


ciaranRoche:
**Aws Strategy Override Map**
image

**Created VPC**
image

**Created Subnets**
image

_Updated Strategy Map_
image

_Un Changed VPC_
image

_Un Changed Subnets_
image

pkg/providers/aws/provider_redis.go
```diff
@@ -173,6 +167,11 @@ func (p *RedisProvider) createElasticacheCluster(ctx context.Context, r *v1alpha
 		return nil, croType.StatusMessage(errMsg), errorUtil.Wrapf(err, errMsg)
 	}
 
+	// we handle standalone networking in CreateRedis() for installs on >= 4.4.6 openshift clusters
+	// this check is to ensure backward compatibility with <= 4.4.5 openshift clusters
+	// creating bundled (in cluster vpc) subnets, subnet groups, security groups
+	//
+	// standaloneNetworkExists if no bundled resources are found in the cluster vpc
```
aidenkeating, line 174:
```suggestion
	// standaloneNetworkExists if no bundled subnets (created by this operator) are found in the cluster vpc
```

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -900,3 +911,70 @@ func isValidCIDRRange(CIDR *net.IPNet) bool {
...
+	return nil
+}
+
+// getNetworkProviderConfig return parsed ipNet cidr block
+// a _network resource type strategy, is expected to have the same tier as either postgres or redis resource type
+// ie. for a postgres tier X there should be a corresponding _network tier X
+//
+// the _network strategy config is unmarshalled into a ec2 create vpc input struct
+// from the struct the cidr block is parsed to ensure validity
+func getNetworkProviderConfig(ctx context.Context, configManager ConfigManager, logger *logrus.Entry, tier string) (*net.IPNet, error) {
```
aidenkeating, line 955: Not a big thing, but I feel like the `logger` should be last here, as it's the most "util" of the arguments in the function signature

pkg/providers/aws/cluster_network_provider_test.go
```diff
@@ -32,6 +34,7 @@ const (
 	validCIDRTwentyThree       = "10.0.50.0/23"
 	defaultValidSubnetMaskTwoA = "10.0.50.0/24"
 	defaultValidSubnetMaskTwoB = "10.0.51.0/24"
+	defaultNonOverlappingCidr  = "192.0.0.0/20"
```
aidenkeating, line 37: I think we're going to have to take a look at cleaning up some testing in a separate PR. I understand this variable naming now, but I can imagine future readers asking `non overlapping with what?/which cidr?`

I don't think it's important for this though, just a note for myself

PR #153, opened 2020-06-22
configure default vpc security groups and create required routes in route tables

laurafitzgerald:
## Overview

Jiras: 
https://issues.redhat.com/browse/INTLY-8105
https://issues.redhat.com/browse/INTLY-8106
Relates to/Replaces https://github.com/integr8ly/cloud-resource-operator/pull/152

## Verification
_Note: this should be run on a fresh BYOC cluster with no previous networking configuration_

### Verify Connection 
With this PR we are now provisioning resources in a standalone VPC and **not** the cluster VPC, for verification we need to ensure we can establish a connection between the cluster vpc and standalone vpc
- Checkout branch
- Run `make cluster/prepare NAMESPACE=<your favourite namespace>`
- Build your own image and push to quay using `make image/push IMAGE_ORG=<your quay username>` 
- Ensure the quay repo is public
- Update `deploy/operator.yaml` to use image the image above `	sed -i.bak 's/image:.*/image: quay\.io\/<your quay username>\/cloud-resource-operator:0.16.1/g' deploy/operator.yaml && rm deploy/operator.yaml.bak
`
- Run `make setup/service_account NAMESPACE=<your favourite namespace>`
- Run `make cluster/seed/managed/postgres NAMESPACE=<your favourite namespace>`
- Run `oc apply -f deploy/operator.yaml -n <your favourite namespace>`
- Add a workshop Postgres Instance to use to verify connection
```
kind: Postgres
metadata:
  name: verify-postgres
  labels:
    productName: productName
spec:
  secretRef:
    name: verify-postgres-sec
  tier: development
  type: workshop
```
- Once `example-postgres` has been successful, open the openshift console
- Navigate to <your favourite namespace> and click on pods, and open the `verify-postgres` pod
- Here navigate to the pods terminal
- In the terminal run 
```
# host value can be found in `example-postgres-sec` secret in <your-favourite-namespace>
psql --host=<host> --port=5432 --username=postgres --password --dbname=postgres
# you will be prompted for a password, this password can also be found in `example-postgres-sec`
```
- Verify the connection is successful 

### Verify Resources
The following resources are provisioned when configuring a standalone vpc, please take time to verify all resources are created and are as expected: 
- Standalone VPC (valid cidr range between /16 and /26
- 2 subnets (cidr block mask one bit added to vpc cidr block mask)
- 1 security group
- 1 route table entry added to standalone vpc route table
- 1 route table entry added to cluster vpc route table
- 1 rds subnet group (contains 2 subnets)
- 1 elasticache subnet group (contains 2 subnets)
### Find Resources
- Verify that
  - security group are correctly reconciled
     - Get the standalone vpc details using the following:
     - Run `aws ec2 describe-vpcs --filter "Name=tag-value,Values=RHMI Cloud Resource VPC"`
     - Note the vpc id and cidr block 
     - Get the cluster vpc details using the following:
     - run `aws ec2 describe-vpcs`
     - Search for the vpc with tag and name <cluster-name-randomstring-vpc>` e.g. for cluster name hcd-rhim-clouds the name will be`"hcd-rhmi-clouds-zpxrz-vpc"`
     - Note the vpc id and cidr block as the cluster vpc details
     - Run `aws ec2 describe-security-groups --filter "Name=tag-value,Values=RHMI Cloud Resource Security Group"`
     - Verify that the Security Group is present with the following configuration
         - Description: `security group for cluster <cluster-name>-<random-string>`
         - Group Name: `<cluster-name><randomstring>securitygroup`
         - VpcId: `previously noted standalone vpc id`
         - IpPermissions[0].IpProtocol: `-1`
         - IpPermissions[0].IpRanges[0].CidrIp: `previously noted cluster cidr block`
   - route in any standalone vpc related route table is created 
      - Get the peeringconnection using the following:
      - Run `aws ec2 describe-vpc-peering-connections --filter "Name=tag-value,Values=RHMI Cloud Resource Peering Connection"`
      - Make a note of VpcPeeringConnectionId
      - Get the standalone Route Table using the following:
      - `aws ec2 describe-route-tables --filter "Name=tag-value,Values=RHMI Cloud Resource Route Table"`
      - Verify the following configuration is correct (using i and j for array reference as they can be in different orders):
         - Routes[i].DestinationCidrBlock: `<previously noted cluster vpc cidr bloc>`
         - Routes[i].VpcPeeringConnectionId: `<previously noted VpcPeeringConnectionId>`
         - Routes[j].DestinationCidrBlock: `<previously noted standalone vpc cidr block>`
         - Routes[j].GatewayId: "local"
   - route in cluster route tables is created 
      - get each route table (there should be two):
      - Run `aws ec2 describe-route-tables --filter "Name=vpc-id,Values=<previous-noted-cluster-vpc-id>`
      - For each route table verify the following configuration (using i and j for array reference as they can be in different orders):
         - Routes[i].DestinationCidrBlock: `<previously noted standalone vpc cidr bloc>`
         - Routes[i].VpcPeeringConnectionId: `<previously noted VpcPeeringConnectionId>`
         - Routes[j].DestinationCidrBlock: `<previously noted cluster vpc cidr block>`
         - Routes[j].GatewayId: "local"
   - RDS SubnetGroup Tags will be replaced if they are deleted/updated
      - Navigate in aws to rds -> subnetgroups.
      - Make a note of the value of the tag Name `integreatly.org/clusterID`
      - Edit the subnet via `Edit`
      - Delete the tag via `Manage Tags`
      - Verify it is replaced
      - Edit the tag via `Manage Tag`
      - Verify that it is set back to `integreatly.org/clusterID | <previously-noted-value>`
   - There should also be no errors on subsequent reconciles


pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1110,32 +1110,62 @@ func (n *NetworkProvider) reconcileRDSVpcConfiguration(ctx context.Context, priv
...
+		}
+
+		tags, err := n.RdsApi.ListTagsForResource(&rds.ListTagsForResourceInput{
+			ResourceName: aws.String(*foundSubnet.DBSubnetGroupArn),
+		})
+		if err != nil {
+			return errorUtil.Wrap(err, "error getting subnet group tags")
+		}
+		for _, tag := range tags.TagList {
+			if tag.Key == aws.String("cluster") && tag.Value != aws.String(clusterID) {
```
ciaranRoche, line 1153: I think this `cluster` key is incorrect, we should be using the the `croOwnerTag`, can we update it here and on creation of the subnet group.

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1157,6 +1187,32 @@ func (n *NetworkProvider) reconcileRDSVpcConfiguration(ctx context.Context, priv
...
+	})
+	if err != nil {
+		return errorUtil.Wrap(err, "error updating db subnet group tags")
+	}
+	_, err = n.RdsApi.AddTagsToResource(&rds.AddTagsToResourceInput{
+		ResourceName: aws.String(*foundSubnet.DBSubnetGroupArn),
+		Tags: []*rds.Tag{
+			{
+				Key:   aws.String("cluster"),
+				Value: aws.String(clusterID),
```
ciaranRoche, line 1206: Might be better to pass in `[]*rds.Tag` to keep this function a bit more generic.

pkg/providers/aws/cluster_network_provider.go
ciaranRoche, line 1153: we have a `tagContains` helper utility, might be better to use that here to keep with the same pattern as the other checks 
https://github.com/integr8ly/cloud-resource-operator/blob/master/pkg/providers/aws/tags.go#L26

pkg/providers/aws/tags.go
```diff
@@ -15,14 +16,41 @@ type tag struct {
...
+
+func genericToRdsTag(tag *tag) *rds.Tag {
+	var rdsTag *rds.Tag
+	rdsTag = &rds.Tag{Key: aws.String(tag.key),Value: aws.String(tag.value)}
+	return rdsTag
+}
+
+func rdsTagstoGeneric(rdsTags []*rds.Tag) []*tag {
+	var genericTags []*tag
+	for _, ec2Tag := range rdsTags {
```
ciaranRoche, line 48: i think this is a copy paste error with the naming of `ec2Tag`

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -805,7 +805,7 @@ func (n *NetworkProvider) reconcileStandaloneSecurityGroup(ctx context.Context,
 	}
 	securityGroupTags := ec2TagsToGeneric(standaloneSecGroup.Tags)
 	if !tagsContains(securityGroupTags, tagDisplayName, DefaultRHMISecurityGroupNameTagValue) ||
-		!tagsContains(securityGroupTags, *croOwnerTag.Key, *croOwnerTag.Value) {
+		!tagsContains(securityGroupTags, croOwnerTag.key, croOwnerTag.value) {
```
ciaranRoche, line 808: Why did we remove the pointer?

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -890,7 +890,7 @@ func (n *NetworkProvider) reconcileStandaloneRouteTableTags(ctx context.Context,
 	for _, routeTable := range routeTables {
 		routeTableTags := ec2TagsToGeneric(routeTable.Tags)
 		if !tagsContains(routeTableTags, tagDisplayName, DefaultRHMIVpcRouteTableNameTagValue) ||
-			!tagsContains(routeTableTags, aws.StringValue(croOwnerTag.Key), aws.StringValue(croOwnerTag.Value)) {
+			!tagsContains(routeTableTags, *aws.String(croOwnerTag.key), *aws.String(croOwnerTag.value)) {
```
ciaranRoche, line 893: I dont think this looks right, can we not use `aws.StringValue`?

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -899,13 +899,13 @@ func (n *NetworkProvider) reconcileStandaloneRouteTableTags(ctx context.Context,
...
 						Value: aws.String(DefaultRHMIVpcRouteTableNameTagValue),
-					}, croOwnerTag,
+					}, genericToEc2Tag(croOwnerTag),
 				},
 			})
 			if err != nil {
 				return errorUtil.Wrap(err, "unable to tag route table")
 			}
-			logger.Infof("successfully tagged route table: %s for clusterID: %s", aws.StringValue(routeTable.RouteTableId), aws.StringValue(croOwnerTag.Value))
+			logger.Infof("successfully tagged route table: %s for clusterID: %s", aws.StringValue(routeTable.RouteTableId), aws.String(croOwnerTag.value))
```
ciaranRoche, line 908: I think this should be the `StringValue`

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1087,13 +1087,13 @@ func (n *NetworkProvider) reconcileVPCTags(vpc *ec2.Vpc, clusterIDTag *ec2.Tag)
...
 					Value: aws.String(DefaultRHMIVpcNameTagValue),
-				}, clusterIDTag,
+				}, genericToEc2Tag(clusterIDTag),
 			},
 		})
 		if err != nil {
 			return errorUtil.Wrap(err, "unable to tag vpc")
 		}
-		logger.Infof("successfully tagged vpc: %s for clusterID: %s", *vpc.VpcId, *clusterIDTag.Value)
+		logger.Infof("successfully tagged vpc: %s for clusterID: %s", *vpc.VpcId, clusterIDTag.value)
```
ciaranRoche, line 1096: I think we are missing a pointer here?

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1110,30 +1110,53 @@ func (n *NetworkProvider) reconcileRDSVpcConfiguration(ctx context.Context, priv
...
-		return errorUtil.Wrap(err, "error getting clusterID")
+		return errorUtil.Wrap(err, "failed getting rds subnet group")
+	}
+	if foundSubnet != nil {
+		logger.Infof("subnet group %s found, verifying it is in the expected state", *foundSubnet.DBSubnetGroupName)
+		if foundSubnet.DBSubnetGroupDescription != aws.String(defaultSubnetGroupDesc) {
+			_, err := n.RdsApi.ModifyDBSubnetGroup(&rds.ModifyDBSubnetGroupInput{
+				DBSubnetGroupDescription: aws.String(defaultSubnetGroupDesc),
+				DBSubnetGroupName: aws.String(*foundSubnet.DBSubnetGroupName),
+				SubnetIds: subnetIDs,
```
ciaranRoche, line 1139: Why are we adding the subnets here, if we have only checked the `description`

I think we should split this, and check the subnets are correct first, as these are the most important. 
The description does not add too much value only a means of having a human readable way of identifying what this group is. We should focus on ensuring the subnets are correct

laurafitzgerald: Yeah I had the same thought and started with only the `DBSubnetGroupDescription` but aws will error out with missing parameters unless all three are passed unfortunately.

pkg/providers/aws/cluster_network_provider.go
laurafitzgerald: Tested this one with a print as suggested and works fine without the pointer. It's related to the change to the return type of getCloudResourceOperatorOwnerTag

pkg/providers/aws/cluster_network_provider.go
laurafitzgerald: yep. it's a generic tag now so don't need those, will remove

pkg/providers/aws/cluster_network_provider.go
laurafitzgerald: i'll remove the aws.String bit as we discussed.

pkg/providers/aws/cluster_network_provider_test.go
```diff
@@ -693,7 +694,25 @@ func TestNetworkProvider_CreateNetwork(t *testing.T) {
...
+				RdsApi: buildMockRdsClient(func(rdsClient *mockRdsClient) {
+					rdsClient.subnetGroups = buildRDSSubnetGroup()
+					rdsClient.modifyDBSubnetGroupFn = func(input *rds.ModifyDBSubnetGroupInput) (*rds.ModifyDBSubnetGroupOutput, error) {
+						return &rds.ModifyDBSubnetGroupOutput{}, nil
+					}
+					rdsClient.listTagsForResourceFn = func(input *rds.ListTagsForResourceInput) (*rds.ListTagsForResourceOutput, error) {
+						return &rds.ListTagsForResourceOutput{
+							TagList: []*rds.Tag{
+								{
+									Key: aws.String("something"),
```
laurafitzgerald, line 706: Todo, need to fix this up.

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1157,6 +1176,30 @@ func (n *NetworkProvider) reconcileRDSVpcConfiguration(ctx context.Context, priv
 	return nil
 }
 
+//this function removes tags and reads them to an rds subnet group
+func (n *NetworkProvider) updateSubnetGroupTags(foundSubnet *rds.DBSubnetGroup, croOwnerTag *rds.Tag, logger *logrus.Entry) error {
+	logger.Info()
```
aidenkeating, line 1181: Should this have a log?

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1035,7 +1035,7 @@ func (n *NetworkProvider) reconcileStandaloneVPCSubnets(ctx context.Context, log
 	for _, sub := range subs {
 		logger.Infof("validating subnet %s", *sub.SubnetId)
 		if !tagsContains(ec2TagsToGeneric(sub.Tags), defaultAWSPrivateSubnetTagKey, "1") ||
-			!tagsContains(ec2TagsToGeneric(sub.Tags), aws.StringValue(croClusterOwnerTag.Key), aws.StringValue(croClusterOwnerTag.Value)) ||
+			!tagsContains(ec2TagsToGeneric(sub.Tags), *aws.String(croClusterOwnerTag.key), *aws.String(croClusterOwnerTag.key)) ||
```
aidenkeating, line 1038: `croClusterOwnerTag.key` here is a `string`

`aws#String(string) *string` takes in a `string` and returns a pointer to it

The `*` then dereferences the pointer that `aws#String` returned, and provides back the original `string`

So we end up getting `croClusterOwnerTag.key` back. Unless I've misread, this should be the same as doing:

```suggestion
			!tagsContains(ec2TagsToGeneric(sub.Tags), croClusterOwnerTag.key, croClusterOwnerTag.value) ||
```

If that makes sense?

I've also changed the second `croClusterOwnerTag.key` to `croClusterOwnerTag.value` as I'm assuming that's what was intended?

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1157,6 +1176,30 @@ func (n *NetworkProvider) reconcileRDSVpcConfiguration(ctx context.Context, priv
 	return nil
 }
 
+//this function removes tags and reads them to an rds subnet group
+func (n *NetworkProvider) updateSubnetGroupTags(foundSubnet *rds.DBSubnetGroup, croOwnerTag *rds.Tag, logger *logrus.Entry) error {
+	logger.Info()
+	_, err := n.RdsApi.RemoveTagsFromResource(&rds.RemoveTagsFromResourceInput{
+		ResourceName: aws.String(*foundSubnet.DBSubnetGroupArn),
+		TagKeys: []*string{
+			aws.String(*croOwnerTag.Key),
```
aidenkeating, line 1185: `croOwnerTag.Key` I'm guessing is a pointer to a string, `*string`?

The `*` dereferences it to the string value

`aws#String(string) *string` takes in a string value and returns a pointer.

So if I've understood correctly, this is the same as doing:

```suggestion
			croOwnerTag.Key,
```

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1157,6 +1176,30 @@ func (n *NetworkProvider) reconcileRDSVpcConfiguration(ctx context.Context, priv
...
+		},
+	})
+	if err != nil {
+		return errorUtil.Wrap(err, "error updating db subnet group tags")
+	}
+	_, err = n.RdsApi.AddTagsToResource(&rds.AddTagsToResourceInput{
+		ResourceName: aws.String(*foundSubnet.DBSubnetGroupArn),
```
aidenkeating, line 1192: Same here with the dereferencing, we can do:

```suggestion
		ResourceName: foundSubnet.DBSubnetGroupArn,
```

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1157,6 +1176,30 @@ func (n *NetworkProvider) reconcileRDSVpcConfiguration(ctx context.Context, priv
 	return nil
 }
 
+//this function removes tags and reads them to an rds subnet group
+func (n *NetworkProvider) updateSubnetGroupTags(foundSubnet *rds.DBSubnetGroup, croOwnerTag *rds.Tag, logger *logrus.Entry) error {
+	logger.Info()
+	_, err := n.RdsApi.RemoveTagsFromResource(&rds.RemoveTagsFromResourceInput{
+		ResourceName: aws.String(*foundSubnet.DBSubnetGroupArn),
```
aidenkeating, line 1183:
```suggestion
		ResourceName: foundSubnet.DBSubnetGroupArn
```

pkg/providers/aws/tags.go
```diff
@@ -23,6 +30,26 @@ func ec2TagsToGeneric(ec2Tags []*ec2.Tag) []*tag {
...
+	var ec2Tag *ec2.Tag
+	ec2Tag = &ec2.Tag{Key: aws.String(tag.key), Value: aws.String(tag.value)}
+	return ec2Tag
+}
+
+func genericToRdsTag(tag *tag) *rds.Tag {
+	var rdsTag *rds.Tag
+	rdsTag = &rds.Tag{Key: aws.String(tag.key), Value: aws.String(tag.value)}
+	return rdsTag
+}
```
aidenkeating, line 43: That's pretty cool, doing it both ways 👍  We can return them inline if we want to, like

```suggestion
func genericToEc2Tag(tag *tag) *ec2.Tag {
	return &ec2.Tag{Key: aws.String(tag.key), Value: aws.String(tag.value)}
}

func genericToRdsTag(tag *tag) *rds.Tag {
	return &rds.Tag{Key: aws.String(tag.key), Value: aws.String(tag.value)}
}
```

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1110,30 +1110,52 @@ func (n *NetworkProvider) reconcileRDSVpcConfiguration(ctx context.Context, priv
...
 	if err != nil {
-		return errorUtil.Wrap(err, "error getting clusterID")
+		return errorUtil.Wrap(err, "failed getting rds subnet group")
+	}
+	if foundSubnet != nil {
+		logger.Infof("subnet group %s found, verifying it is in the expected state", *foundSubnet.DBSubnetGroupName)
+		if foundSubnet.DBSubnetGroupDescription != aws.String(defaultSubnetGroupDesc) {
+			_, err := n.RdsApi.ModifyDBSubnetGroup(&rds.ModifyDBSubnetGroupInput{
+				DBSubnetGroupDescription: aws.String(defaultSubnetGroupDesc),
+				DBSubnetGroupName:        aws.String(*foundSubnet.DBSubnetGroupName),
```
aidenkeating, line 1138: This will equate to

```suggestion
				DBSubnetGroupName:        foundSubnet.DBSubnetGroupName,
```

as we're dereferencing and then using `aws#String(string) *string` to reference to a pointer again.

All `aws#String(string) *string` is doing is:

```
// String returns a pointer to the string value passed in.
func String(v string) *string {
	return &v
}
```

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1110,30 +1110,52 @@ func (n *NetworkProvider) reconcileRDSVpcConfiguration(ctx context.Context, priv
...
+				SubnetIds:                subnetIDs,
+			})
+			if err != nil {
+				return errorUtil.Wrap(err, "error updating db subnet group description")
+			}
+		}
+		tags, err := n.RdsApi.ListTagsForResource(&rds.ListTagsForResourceInput{
+			ResourceName: aws.String(*foundSubnet.DBSubnetGroupArn),
```
aidenkeating, line 1146:
```suggestion
			ResourceName: foundSubnet.DBSubnetGroupArn,
```

aidenkeating:
- verified connectivity
- verified resources exist
- verified commands [1]

[1] https://gist.github.com/aidenkeating/26cb620810d72c0b755062e75ecea901

PR #154, opened 2020-06-25
verify last instance of postgres or redis before deleting standalone network

laurafitzgerald:
## Overview

Jira: https://issues.redhat.com/browse/INTLY-8163

Pushing for review initially. Verification steps are included below but they can wait until code review are done in case of changes. 

## Verification

Note: this should be run on a fresh BYOC cluster with no previous networking configuration
Note: You don't need to let the redis and postgres install completely. It's enough to ensure they get deleted correctly on deletion of the cr. (I think 🤔 )

- Checkout this branch
- Run `make cluster/prepare`
- Run `make run`

- Verify the standalone vpc is created by running ` aws ec2 describe-vpcs --filter "Name=tag-value,Values=RHMI Cloud Resource VPC"` - There should be one VPC - note the id of the vpc.
- Check that the redis and postgres resources are being created by running the following:
   - `aws rds describe-db-instances`  - there should be one instance associated with `DBInstanceIdentifier` == `<clustername><randomstring>cloudresourceoper<randomstring>`
   - `aws elasticache describe-cache-clusters` - there should be on instance associated with `CacheClusterId` == `<clustername><randomstring>cloudresourceoper-<randomstring>`

- Run `oc new-project cro-other-project`
- Run `make cluster/seed/managed/postgres NAMESPACE=cro-other-project`
- Check it's created. cro shouldn't respond to that one.

- Delete the redis cr from the namespace cro is running against.
- Wait for the redis instance to be deleted from aws and the cr to be deleted from the cluster. This may take a while.

- Verify that the VPC is still in aws ` aws ec2 describe-vpcs --filter "Name=tag-value,Values=RHMI Cloud Resource VPC"` - There should be one VPC with the same id previously noted.

- Delete the postgres cr from the namespace cro is running against
- Wait for the postgres instance to be deleted from aws and the cr to be deleted from the cluster. This may also take a while.

- Verify that the VPC is deleted in aws. It should ignore the additional postgres in the other namespace.

### DIFF
--- a/pkg/providers/aws/provider_postgres.go
+++ b/pkg/providers/aws/provider_postgres.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/operator-framework/operator-sdk/pkg/k8sutil"
 	"strconv"
 	"time"
 
@@ -12,8 +14,6 @@ import (
 
 	"github.com/integr8ly/cloud-resource-operator/pkg/annotations"
 	croType "github.com/integr8ly/cloud-resource-operator/pkg/apis/integreatly/v1alpha1/types"
-
-	"github.com/aws/aws-sdk-go/aws/awserr"
 
 	"k8s.io/apimachinery/pkg/types"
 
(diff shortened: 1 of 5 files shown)


pkg/providers/aws/provider_postgres.go
```diff
@@ -444,8 +455,44 @@ func (p *PostgresProvider) deleteRDSInstance(ctx context.Context, pg *v1alpha1.P
...
+		if err != nil {
+			msg := "failed to remove deletion protection"
+			return croType.StatusMessage(msg), errorUtil.Wrap(err, msg)
+		}
+
+		return croType.StatusMessage(fmt.Sprintf("deletion protection detected, modifyDBInstance() in progress, current aws rds status is %s", *foundInstance.DBInstanceStatus)), nil
+	}
+
+	// standaloneNetworkExists if no bundled resources are found in the cluster vpc
+	if standaloneNetworkExists && isLastResource {
```
ciaranRoche, line 494: Just a concern, when we `DeleteNetworkConnection` we remove the route table entry in the cluster route table. A valid thing to do IMO, but we require the peering connection to do this. The issue is, if an error happens when `DeleteNetwork` is called, we re-reconcile, we `GetClusterNetworkPeering`, then try `DeleteNetworkPeering` this will return, as the peering connection does not exist, and the `networkPeering` object will be nil, we call `DeleteNetworkConnection` with a nil object we return an error. We never get passed this error and the original error is lost. 

We should look to update the logic in `DeleteNetworkConnection` to handle this flow better. As the route table entry is a nice to have, we can probably do one of a few things. 

1. If the `networkPeering` connection is nil, we can return and not error or try remove the route table entries. 
2. When the peering connection begins to delete the route table entry goes into `blackhole` status, we can check for `blackhole` status and the standalone vpc cidr block. And delete the route table entry based on those fields existing. 
3. Maybe a discussion around on how necessary removing these entries are from the cluster VPC, chances are if RHMI is being deleted the cluster will be nukes and the route table will be deleted with the cluster VPC

What do you think, might be worth a call/discussion about this.

pkg/providers/aws/provider_postgres.go
```diff
@@ -567,6 +581,20 @@ func (p *PostgresProvider) getRDSConfig(ctx context.Context, r *v1alpha1.Postgre
 	return rdsCreateConfig, rdsDeleteConfig, stratCfg, nil
 }
 
+func (p *PostgresProvider) isLastResource(ctx context.Context) (bool, error) {
```
ciaranRoche, line 584: While this may never happen, I am just thinking of the case something outside of CRO's control gets provisioned in the CRO VPC, how would we handle this situation. We would error out and the un-install would never complete. Is this a behaviour we want?
Like above might be worth a discussion :thinking:

pkg/providers/aws/provider_postgres.go
aidenkeating: What's the issue with just deleting the route table entry before the peering connection?

pkg/providers/aws/provider_postgres.go
aidenkeating: Let's namespace the list to the watch namespace. I think this situation is unlikely to happen.

pkg/providers/aws/provider_postgres.go
ciaranRoche: no issue, that would probably be the easiest fix, to alternate the flow of the order we remove items. Remove the connection -> remote the peering -> remove the network.

pkg/providers/aws/provider_postgres.go
ciaranRoche: agree, unlikely to happen. 

Good catch on reducing the list scope to the watch namespace

pkg/providers/aws/provider_postgres.go
aidenkeating: Let's go with that then, unless you see any issues @laurafitzgerald ?

pkg/providers/aws/provider_postgres.go
aidenkeating: @laurafitzgerald Unless there are any objections, we can go with that?

pkg/providers/aws/provider_postgres.go
laurafitzgerald: @aidenkeating @ciaranRoche the flow is now `Remove the connection -> remote the peering -> remove the network.` in case postgres and redis. 👍

pkg/providers/aws/provider_postgres.go
laurafitzgerald: @aidenkeating @ciaranRoche the check now accepts the watch namespace parameter. Updated the verification to include that as a check. 👍

laurafitzgerald:
@psturc this one should be good to verify when you have some time. 🙂

psturc:
@laurafitzgerald 👌  I'm on it

psturc:
👍 lgtm, verified on BYOC OSD cluster (v4.4.9)

PR #155, opened 2020-06-26
feat: deletion of rds & elasticache snapshots

darahayes:
## Overview

Jira: https://issues.redhat.com/browse/INTLY-5484

* Refactors the aws specific logic in the postgres and redis snapshot controllers into new postgres and redis aws providers
* Adds deletion capability for postgres and redis snapshots

## Discussion

I'm going to leave some comments/questions on a couple of things in the code changes.

## Verification

- Clone this branch
- Run `make cluster/prepare`
- Run `make run`

# postgres

* Create a postgres CR to create a postgres instance
* You can immediately create a postgres snapshot that references the postgres instance
	* Ensure that while the postgres instance is still being created, the snapshot CR will have a status message along the lines of "waiting for postgres instance to be available"
* Wait until snapshot is created in AWS
* delete the Snapshot CR.
	* Ensure that the snapshot in AWS is removed
	* Ensure that the snapshot CR is successfully deleted (i.e. the finalizer is removed)

# redis

* Create a redis CR to create a postgres instance
* You can immediately create a redis snapshot that references the redis instance
	* Ensure that while the redis instance is still being created, the snapshot CR will have a status message along the lines of "current replication group status is creating"
* Wait until snapshot is created in AWS
* delete the Snapshot CR.
	* Ensure that the snapshot in AWS is removed
	* Ensure that the snapshot CR is successfully deleted (i.e. the finalizer is removed)


pkg/providers/aws/provider_postgressnapshot.go
```diff
@@ -0,0 +1,244 @@
...
+		// or is already backing up
+		if postgres.Status.Phase == croType.PhaseInProgress {
+			return nil, croType.StatusMessage("waiting for postgres instance to be available"), nil
+		}
+		// postgres instance is being deleted
+		// impossible to create a snapshot
+		if postgres.Status.Phase == croType.PhaseDeleteInProgress {
+			errMsg := "cannot create snapshot when instance deletion is in progress"
+			return nil, croType.StatusMessage(errMsg), errorUtil.New(errMsg)
+		}
```
darahayes, line 133: This is new logic that wasn't originally there.

pkg/providers/aws/provider_postgressnapshot.go
```diff
@@ -0,0 +1,244 @@
...
+	for _, c := range listOutput.DBSnapshots {
+		if *c.DBSnapshotIdentifier == snapshotName {
+			foundSnapshot = c
+			break
+		}
+	}
+	return foundSnapshot, nil
+}
+
+func (r *PostgresSnapshotProvider) getAwsSession(ctx context.Context, postgresCr *v1alpha1.Postgres) (*session.Session, error) {
```
darahayes, line 217: This functionality used to be directly in the create function but after adding the ability to delete, it made sense to make it a bit more reusable. The thing is I also have this function in `provider_redissnapshot.go` and it's pretty much identical except it takes in a Redis CR type as an argument. Question: are we happy with the duplication or should I make it into a more generic function that will work for both providers?

pkg/providers/aws/provider_redissnapshot.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+
+	// find primary cache node
+	var cacheName string
+	for _, i := range cacheOutput.ReplicationGroups[0].NodeGroups[0].NodeGroupMembers {
+		if *i.CurrentRole == "primary" {
+			cacheName = *i.CacheClusterId
+			break
+		}
+	}
```
darahayes, line 130: Question: this is copy pasted from the previous redis controller and I didn't make any changes. I didn't check yet but it looks like there's the potential for nil pointer errors here if AWS returned back an empty list. Should I add some more checks?

pkg/controller/postgressnapshot/postgressnapshot_controller_test.go
```diff
@@ -1,210 +0,0 @@
-package postgressnapshot
```
darahayes, line 1: Deleted this file and the redissnapshot_controller_test.go also because they were only unit testing a create method that doesn't exist in the controller anymore. There are unit tests for the new providers that check their create methods

pkg/providers/aws/provider_postgressnapshot.go
```diff
@@ -0,0 +1,244 @@
...
+	// if snapshot status complete update status
+	if *foundSnapshot.Status == "available" {
+		return &providers.PostgresSnapshotInstance{
+			Name: *foundSnapshot.DBSnapshotIdentifier,
+		}, "snapshot created", nil
+	}
+
+	// creation in progress
+	msg := fmt.Sprintf("current snapshot status : %s", *foundSnapshot.Status)
+	p.logger.Info(msg)
```
ciaranRoche, line 155: Just a small one, we are in the process of updating the logging format in CRO - https://github.com/integr8ly/cloud-resource-operator/blob/master/pkg/providers/aws/provider_postgres.go#L105-L106

Using the `withField` and `action` we can pin point the function that called the log. Like adding the developer docs we are trying to make a concious effort to update these as we go, to avoid large refactors/tech debt 

There is even a utility function for it I just noticed :) https://github.com/integr8ly/cloud-resource-operator/blob/master/pkg/resources/logger.go#L9

pkg/providers/aws/provider_postgressnapshot.go
```diff
@@ -0,0 +1,244 @@
...
+			return nil, croType.StatusMessage("waiting for postgres instance to be available"), nil
+		}
+		// postgres instance is being deleted
+		// impossible to create a snapshot
+		if postgres.Status.Phase == croType.PhaseDeleteInProgress {
+			errMsg := "cannot create snapshot when instance deletion is in progress"
+			return nil, croType.StatusMessage(errMsg), errorUtil.New(errMsg)
+		}
+		p.logger.Info("creating rds snapshot")
+		_, err = rdsSvc.CreateDBSnapshot(&rds.CreateDBSnapshotInput{
```
ciaranRoche, line 135: Should we be tagging these SnapShots on creation?

pkg/providers/aws/provider_redissnapshot.go
ciaranRoche: If we add an err handler here https://github.com/integr8ly/cloud-resource-operator/pull/155/files#diff-c5a0bd9058dfdae2fb6c150c7a880efcR108 when you try to describe a resource by `ReplicationGroupId` if it does not find it, it will return an error. So we wont get down this far. 
If it does find it, it will be the first item in the list of `ReplicationGroups`

pkg/providers/aws/provider_postgressnapshot.go
darahayes: Good point. I don't think the old logic tagged the snapshots either. I could add that here as part of this PR but I wonder would it make more sense to create a separate ticket for it?

pkg/providers/aws/provider_postgressnapshot.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+func (p *PostgresSnapshotProvider) CreatePostgresSnapshot(ctx context.Context, snapshot *v1alpha1.PostgresSnapshot, postgres *v1alpha1.Postgres) (*providers.PostgresSnapshotInstance, croType.StatusMessage, error) {
+	// add finalizer to the snapshot cr
+	if err := resources.CreateFinalizer(ctx, p.client, snapshot, DefaultFinalizer); err != nil {
+		errMsg := fmt.Sprintf("failed to set finalizer for snapshot %s", snapshot.Name)
+		return nil, croType.StatusMessage(errMsg), errorUtil.Wrap(err, errMsg)
+	}
+
+	session, err := p.getAwsSession(ctx, postgres)
```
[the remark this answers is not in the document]
darahayes: Good to know!

pkg/providers/aws/provider_postgressnapshot.go
```diff
@@ -0,0 +1,244 @@
+package aws
```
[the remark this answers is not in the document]
darahayes: For sure. Thanks for the link

pkg/providers/aws/provider_postgressnapshot.go
ciaranRoche: tut tut tut who wrote that old logic :rofl: 

if we can get it in in this PR I think we should as it should be a small change, the `CreateDBSnapshotInput` has a [tags](https://docs.aws.amazon.com/sdk-for-go/api/service/rds/#CreateDBSnapshotInput) field

But of course elasticache does not have a [tags](https://docs.aws.amazon.com/sdk-for-go/api/service/elasticache/#CreateSnapshotInput) field. :thinking: 

Ill leave it up to you, I dont want to hold up this PR if it will take too much time, another follow up ticket might be worth it.

darahayes: Okay I'll definitely get it in there for Postgres and I'll try get it in for Redis 😄

darahayes: I started working on this but then I discovered that snapshots are actually tagged within the respective postgres and redis providers.

Postgres: https://github.com/integr8ly/cloud-resource-operator/blob/master/pkg/providers/aws/provider_postgres.go#L367-L391
Redis https://github.com/integr8ly/cloud-resource-operator/blob/master/pkg/providers/aws/provider_redis.go#L377-L397

@aidenkeating @ciaranRoche do you think I should move that logic into the new snapshot providers?

pkg/controller/postgressnapshot/postgressnapshot_controller.go
```diff
@@ -133,117 +124,60 @@ func (r *ReconcilePostgresSnapshot) Reconcile(request reconcile.Request) (reconc
...
-	// get resource region
-	stratCfg, err := r.ConfigManager.ReadStorageStrategy(ctx, providers.PostgresResourceType, postgresCr.Spec.Tier)
-	if err != nil {
-		if updateErr := resources.UpdateSnapshotPhase(ctx, r.client, instance, croType.PhaseFailed, croType.StatusMessage(err.Error())); updateErr != nil {
-			return reconcile.Result{Requeue: true, RequeueAfter: resources.ErrorReconcileTime}, updateErr
+	if instance.DeletionTimestamp != nil {
+		msg, err := r.provider.DeletePostgresSnapshot(ctx, instance, postgresCr)
+		if err != nil {
+			if updateErr := resources.UpdateSnapshotPhase(ctx, r.client, instance, croType.PhaseFailed, msg.WrapError(err)); updateErr != nil {
+				return reconcile.Result{Requeue: true, RequeueAfter: resources.ErrorReconcileTime}, updateErr
```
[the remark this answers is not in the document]
darahayes: Okay I've done this and pushed the changes but Github is acting a bit weird and I can't see them here in the PR.

pkg/providers/aws/provider_postgressnapshot.go
ciaranRoche: To save this already big PR from getting bigger I would vote to leave it as it is, as long as we can validate that snapshots created via the snapshot provider are tagged on reconcile of the postgres/redis provider. 

That work in the provider to tag resources, from my memory was added to make sure automatic snapshots taken by AWS had CRO tags on them. It was done before we has logic to take manual snapshots. So it might be worth a ticket (technical debt) to revisit this logic and flow if we are using it to tag both manual and automatic snapshots.  

What do you think?

pkg/controller/postgressnapshot/postgressnapshot_controller.go
```diff
@@ -118,40 +118,40 @@ func (r *ReconcilePostgresSnapshot) Reconcile(request reconcile.Request) (reconc
...
-		return reconcile.Result{Requeue: true, RequeueAfter: resources.ErrorReconcileTime}, errorUtil.New(errMsg)
+		return reconcile.Result{Requeue: true, RequeueAfter: r.provider.GetReconcileTime(instance)}, errorUtil.New(errMsg)
 	}
 
 	if instance.DeletionTimestamp != nil {
 		msg, err := r.provider.DeletePostgresSnapshot(ctx, instance, postgresCr)
 		if err != nil {
 			if updateErr := resources.UpdateSnapshotPhase(ctx, r.client, instance, croType.PhaseFailed, msg.WrapError(err)); updateErr != nil {
-				return reconcile.Result{Requeue: true, RequeueAfter: resources.ErrorReconcileTime}, updateErr
+				return reconcile.Result{Requeue: true, RequeueAfter: r.provider.GetReconcileTime(instance)}, updateErr
```
ciaranRoche, line 139: I think on error we do not want to specify a requeue time, as we want to keep reconciling on an error till we get passed it. Looking at the other controllers, we only slow down the reconcile after it has returned with no error, so that we can limit api calls to aws.

Maybe @aidenkeating can confirm this, if I am wrong we will need to stick a ticket in the backlog to update all the controllers requeue times
```suggestion
				return reconcile.Result{}, updateErr
```

pkg/providers/aws/provider_postgressnapshot.go
darahayes: I'm +1 on that.

aidenkeating: A follow up ticket seems fine. If we are tagging the snapshots anyway I wouldn't see doing this as high priority.

pkg/controller/postgressnapshot/postgressnapshot_controller.go
aidenkeating: `RequeueAfter` will be ignored here

aidenkeating: I think we can just remove the `Result` body anywhere we return an non-nil `error`, I don't see this as anything major

darahayes: Thanks for the comments. I'll return an empty result body instead

aidenkeating:
Verifying in `akeating-cro` on MW Collab

PR #158, opened 2020-07-06
deleting bundle network resources on deletion of last rds or elasticache

laurafitzgerald:
## Overview

Jira: https://issues.redhat.com/browse/INTLY-5160

Pushing an early pr for review to verify that the approach is good. 

## Verification

- Install cloud-resource-operator with bundled resources
  - git checkout v0.16.1 
  - `make cluster/prepare NAMESPACE=cloud-resource-operator-<name>`
  - `make cluster/seed/managed/redis NAMESPACE=clourd-resource-operator-<name>`
  - `make cluster/seed/managed/redis NAMESPACE=cloud-resource-operator-<name>`
  - `make run/local NAMESPACE=cloud-resource-operator-<name>`
  - Get the cluster vpc id by running `aws ec2 describe-vpcs` and noting the vpc id with the tag     ```{
                    "Key": "kubernetes.io/cluster/<clustername>-<randomchars>",
                    "Value": "owned"
                }```
  - Wait for the resources to install. The resources of interest are below and will follow naming syntax `<clustername><randomchars>subnetgroup` and `<clustername><randomchars>securitygroup` and an associated vpc noted above.
      - ec2 security group with security group 
      - rds subent group with the name similar to the below
      - elasticache subnet group with the name similar to the below

See images below for examples for cluster mw-colab-byoc - using ui rather than cli as there's problems with the cli for some of the resources.
Screenshot 2020-07-07 at 15 25 10
Screenshot 2020-07-07 at 15 26 47
       - elasticache subnet group with the name similar to the below
Screenshot 2020-07-07 at 15 27 46

- Run the operator with the deletion changes in place - i.e. use this branch
  - `git checkout INTLY-5160`
  - `make run NAMESPACE=cloud-resource-operator-<name>`
  - delete either a postgres or redis
  - wait for the resource and the cr to delete (this will take a few minutes as the postgres is deleted)
  - verify that the network resources (rds subnet group, elasticache group & ec2 security group) are still in place
      - check in the aws console as above
  - delete the remaining either postgres or redis
  - verify that the operator doesn't error on reconcile if any of the resources are removed.
  - verify that the network resources are deleted - verify in the aws console the resources listed above are deleted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,12 @@ Operator is running.
 ## Running the Cloud Resource Operator
 ## Locally
 
+## Supported Openshift Versions
+
+Due to a change in how networking is configured for Openshift >= v4.4.6 the use of cro <= v0.16.1 with these Openshift versions is unsupported.
+Please use >= v0.17.x of CRO for Openshift >= v4.4.6.
+
+
 Prerequisites:
 - `go`
 - `make`
@@ -56,39 +62,6 @@ Clean up the Kubernetes/OpenShift cluster:
 ```shell script
 $ make cluster/clean
 ```
-
-## VPC Peering 
-Currently AWS resources are deployed into a separate Virtual Private Cloud (VPC) than the VPC that the cluster is deployed into. In order for these to communicate, a `peering connection` must be established between the two VPCS. To do this:
-1. Create a new peering connection between the two VPCs.
-  - Go to `VPC` > `Peering Connections`
-  - Click on `Create Peering Connection`
-    - **NOTE**: This is a two-way communication channel so only one needs to be created.
-  - Select the newly created connection, then click `Actions` > `Accept Request` to accept the peering request.
-2. Edit your cluster's VPC route table
-  - Go to `VPC` > `Your VPCs`
-  - Find the VPC your cluster is using and click on its route table under the heading `Main Route Table`
-    - Your cluster VPC name is usually in a format of `<cluster-name>-<uid>-vpc`
-  - Select the cluster route table and click on `Action` > `Edit routes`
-  - Add a new route with the following details:
-    ```
-    Destination: <resource VPC CIDR block>
-    Target: <newly created peering connection>
-    ```
-  - Click on `Save routes`
-3. Edit the resource VPC route table
-  - Go to `VPC` > `Your VPCs`
-  - Find the VPC where the AWS resources are provisioned in and click on its route table under the heading `Main Route Table`
-    - The name of this VPC is usually empty or named as `default`
-  - Select the cluster route table and click on `Action` > `Edit routes`
-  - Add a new route with the following details:
-    ```
-    Destination: <your cluster's VPC CIDR block>
-    Target: <newly created peering connection>
-    ```
-  - Click on `Save routes`
-4. Edit the Security Groups associated with the resource VPC to ensure database and cache traffic can pass between the two VPCs.
-
-The two VPCs should now be able to communicate with each other. 
 
 ## Snapshots
 The cloud resource operator supports the taking of arbitrary snapshots in the AWS provider for both `Postgres` and `Redis`. To take a snapshot you must create a `RedisSnapshot` or `PostgresSnapshot` resource, which should reference the `Redis` or `Postgres` resource you wish to create a snapshot of. The snapshot resource must also exist in the same namespace.
--- a/pkg/providers/aws/cluster_network_provider.go
+++ b/pkg/providers/aws/cluster_network_provider.go
@@ -27,6 +27,10 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net"
+	"reflect"
+	"sort"
+
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/session"
@@ -39,10 +43,7 @@ import (
 	"github.com/integr8ly/cloud-resource-operator/pkg/providers"
 	"github.com/integr8ly/cloud-resource-operator/pkg/resources"
 	"github.com/sirupsen/logrus"
-	"net"
-	"reflect"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sort"
 
 	errorUtil "github.com/pkg/errors"
 )
@@ -98,6 +99,7 @@ type NetworkManager interface {
 	GetClusterNetworkPeering(context.Context) (*NetworkPeering, error)
 	DeleteNetworkPeering(*NetworkPeering) error
 	IsEnabled(context.Context) (bool, error)
+	DeleteBundledCloudResources(context.Context) error
 }
 
 var _ NetworkManager = (*NetworkProvider)(nil)
@@ -683,6 +685,69 @@ func (n *NetworkProvider) IsEnabled(ctx context.Context) (bool, error) {
 	}
 	logger.Infof("found %d bundled vpc subnets in cluster vpc", len(validBundledVPCSubnets))
 	return len(validBundledVPCSubnets) == 0, nil
+}
+
+// DeleteBundledCloudResources returns an error on any error deleting of the following resources
+// * elasticache subnet group
+// * rds subnet group
+// * ec2 security group
+//
+// it has been located under the cluster network provider as it requires 3 different aws sessions
+// (elasticache, rds and ec2) to delete the required resources even though it deals with bundled
+// resources. The majority of the functionality in this file relates to standalone aws vpc and it's
+// resources.
+func (n *NetworkProvider) DeleteBundledCloudResources(ctx context.Context) error {
+	logger := n.Logger.WithField("action", "deleteBundledCloudResources")
+
+	subnetGroupName, err := BuildInfraName(ctx, n.Client, "subnetgroup", DefaultAwsIdentifierLength)
+	if err != nil {
+		return errorUtil.Wrap(err, "error building bundle subnet group resource name on deletion")
+	}
+	logger.Infof("deleting bundled elasticache subnet group %s, if it's not found it's already deleted and will continue", subnetGroupName)
+	_, err = n.ElasticacheApi.DeleteCacheSubnetGroup(&elasticache.DeleteCacheSubnetGroupInput{
+		CacheSubnetGroupName: aws.String(subnetGroupName),
+	})
+	cacheSubnetErr, isAwsErr := err.(awserr.Error)
+	if err != nil && isAwsErr && cacheSubnetErr.Code() != elasticache.ErrCodeCacheSubnetGroupNotFoundFault {
+		return errorUtil.Wrap(err, "error deleting elasticache subnet group")
+	}
+	logger.Infof("deleting bundled rds subnet group %s, if it's not found it's already deleted and will continue", subnetGroupName)
+	_, err = n.RdsApi.DeleteDBSubnetGroup(&rds.DeleteDBSubnetGroupInput{
+		DBSubnetGroupName: aws.String(subnetGroupName),
+	})
+	dbSubnetErr, isAwsErr := err.(awserr.Error)
+	if err != nil && isAwsErr && dbSubnetErr.Code() != rds.ErrCodeDBSubnetGroupNotFoundFault {
+		return errorUtil.Wrap(err, "error deleting rds subnet group")
+	}
+
+	securityGroupName, err := BuildInfraName(ctx, n.Client, "securitygroup", DefaultAwsIdentifierLength)
+	if err != nil {
+		return errorUtil.Wrap(err, "error building bundle security group resource name on deletion")
+	}
+	logger.Infof("Deleting bundled ec2 security group %s, if it's not found it's already deleted and will continue", securityGroupName)
+	// in the case of the security group the Group Id is required in order to delete security groups
+	// not connected with the default vpc. In order to delete it it is required to describe them
+	// all in the account and then find the one with the correct group name and then request deletion
+	// using the group id of the matched security group
+	securityGroup, err := getSecurityGroup(n.Ec2Api, securityGroupName)
+	if err != nil {
+		return errorUtil.Wrap(err, "error getting ec2 security group")
+	}
+	if securityGroup == nil {
+		return nil
+	}
+	vpc, err := getClusterVpc(ctx, n.Client, n.Ec2Api, logger)
+	if err != nil {
+		return errorUtil.Wrap(err, "error getting cluster vpc")
+	}
+	if securityGroup.VpcId != nil && vpc.VpcId != nil && aws.StringValue(securityGroup.VpcId) == aws.StringValue(vpc.VpcId) {
+		if _, err = n.Ec2Api.DeleteSecurityGroup(&ec2.DeleteSecurityGroupInput{
+			GroupId: securityGroup.GroupId,
+		}); err != nil {
+			return errorUtil.Wrap(err, "error deleting bundled security group")
+		}
+	}
+	return nil
 }
 
 func (n *NetworkProvider) getNetworkPeering(ctx context.Context, network *Network) (*ec2.VpcPeeringConnection, error) {
--- a/pkg/providers/aws/cluster_network_provider_moq.go
+++ b/pkg/providers/aws/cluster_network_provider_moq.go
@@ -10,14 +10,15 @@ import (
 )
 
 var (
-	lockNetworkManagerMockCreateNetwork            sync.RWMutex
-	lockNetworkManagerMockCreateNetworkConnection  sync.RWMutex
-	lockNetworkManagerMockCreateNetworkPeering     sync.RWMutex
-	lockNetworkManagerMockDeleteNetwork            sync.RWMutex
-	lockNetworkManagerMockDeleteNetworkConnection  sync.RWMutex
-	lockNetworkManagerMockDeleteNetworkPeering     sync.RWMutex
-	lockNetworkManagerMockGetClusterNetworkPeering sync.RWMutex
-	lockNetworkManagerMockIsEnabled                sync.RWMutex
+	lockNetworkManagerMockCreateNetwork               sync.RWMutex
+	lockNetworkManagerMockCreateNetworkConnection     sync.RWMutex
+	lockNetworkManagerMockCreateNetworkPeering        sync.RWMutex
+	lockNetworkManagerMockDeleteBundledCloudResources sync.RWMutex
+	lockNetworkManagerMockDeleteNetwork               sync.RWMutex
+	lockNetworkManagerMockDeleteNetworkConnection     sync.RWMutex
+	lockNetworkManagerMockDeleteNetworkPeering        sync.RWMutex
+	lockNetworkManagerMockGetClusterNetworkPeering    sync.RWMutex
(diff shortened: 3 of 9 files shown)


pkg/providers/aws/cluster_network_provider.go
```diff
@@ -685,6 +686,63 @@ func (n *NetworkProvider) IsEnabled(ctx context.Context) (bool, error) {
 	return len(validBundledVPCSubnets) == 0, nil
 }
 
+// DeleteBundledCloudResources returns an error on any error deleting of the following resources
```
laurafitzgerald, line 689: I put this function in cluster_network_provider.go because it has access to the three required sessions for deleting the related resources. Initially I was putting this in cluster_vpc.go. I can still move it back there, but it will require some additional code to get the required aws sessions in place to do it all in one go. I've left a note here that it relates to the bundled networking unlike the other code in this file. It might suit. If not just let me know. :)

pkg/providers/aws/cluster_network_provider_moq.go
```diff
@@ -10,14 +10,15 @@ import (
...
 var (
-	lockNetworkManagerMockCreateNetwork            sync.RWMutex
-	lockNetworkManagerMockCreateNetworkConnection  sync.RWMutex
-	lockNetworkManagerMockCreateNetworkPeering     sync.RWMutex
-	lockNetworkManagerMockDeleteNetwork            sync.RWMutex
-	lockNetworkManagerMockDeleteNetworkConnection  sync.RWMutex
-	lockNetworkManagerMockDeleteNetworkPeering     sync.RWMutex
-	lockNetworkManagerMockGetClusterNetworkPeering sync.RWMutex
-	lockNetworkManagerMockIsEnabled                sync.RWMutex
+	lockNetworkManagerMockCreateNetwork               sync.RWMutex
```
laurafitzgerald, line 13: I did this moq change manually just to get things running. Can probably use the automation to ensure there's nothing missing here.

pkg/providers/aws/cluster_network_provider.go
aidenkeating: I think this is fine 👍

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -685,6 +686,63 @@ func (n *NetworkProvider) IsEnabled(ctx context.Context) (bool, error) {
...
+
+	subnetGroupName, err := BuildInfraName(ctx, n.Client, "subnetgroup", DefaultAwsIdentifierLength)
+	if err != nil {
+		return errorUtil.Wrap(err, "error building bundle subnet group resource name on deletion")
+	}
+	logger.Info("deleting bundled elasticcache subnet group %s , if it's not found it's already deleted and will continue", subnetGroupName)
+	_, err = n.ElasticacheApi.DeleteCacheSubnetGroup(&elasticache.DeleteCacheSubnetGroupInput{
+		CacheSubnetGroupName: aws.String(subnetGroupName),
+	})
+	if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() != "CacheSubnetGroupNotFoundFault" {
```
aidenkeating, line 709: I believe there are variables that we can use instead of inline strings for AWS errors e.g. `elasticache.ErrCodeCacheSubnetGroupNotFoundFault`

Not super important, but it'll help us handle if that exact string ever changes.

pkg/providers/aws/cluster_network_provider_moq.go
aidenkeating: Yep I'd probably run `go generate ./...` just to be sure

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -685,6 +686,63 @@ func (n *NetworkProvider) IsEnabled(ctx context.Context) (bool, error) {
...
+	// in the case of the security group the Group Id is required in order to delete security groups
+	// not connected with the default vpc. In order to delete it it is required to describe them
+	// all in the account and then find the one with the correct group name and then request deletion
+	// using the group id of the matched security group
+	securityGroupOutput, err := n.Ec2Api.DescribeSecurityGroups(&ec2.DescribeSecurityGroupsInput{})
+	if err != nil {
+		return errorUtil.Wrap(err, "error getting ec2 security group")
+	}
+	for _, securityGroup := range securityGroupOutput.SecurityGroups {
+		if *securityGroup.GroupName == securityGroupName {
```
ciaranRoche, line 734: Should we be checking that this security group is targeting the bundled vpc? Just to ensure we are deleting the correct security group and potentially avoid a similar issue we seen last Friday

aidenkeating: Realistically that should never happen. But checking it targets the bundled VPC makes sense for dev environments

pkg/providers/aws/cluster_network_provider_test.go
```diff
@@ -733,10 +736,14 @@ func TestNetworkProvider_CreateNetwork(t *testing.T) {
 					}
 				}),
 				ElasticacheApi: buildMockElasticacheClient(func(elasticacheClient *mockElasticacheClient) {
-					elasticacheClient.cacheSubnetGroup = buildElasticacheSubnetGroup()
 					elasticacheClient.modifyCacheSubnetGroupFn = func(input *elasticache.ModifyCacheSubnetGroupInput) (*elasticache.ModifyCacheSubnetGroupOutput, error) {
 						return &elasticache.ModifyCacheSubnetGroupOutput{}, nil
 					}
+					elasticacheClient.describeCacheSubnetGroupsFn = func(input *elasticache.DescribeCacheSubnetGroupsInput) (*elasticache.DescribeCacheSubnetGroupsOutput, error) {
+						return &elasticache.DescribeCacheSubnetGroupsOutput{
+							CacheSubnetGroups: buildElasticacheSubnetGroup(),
```
ciaranRoche, line 744: If we have time, we should remove and refactor `buildElasticacheSubnetGroup` to the new override modifyFn pattern.

pkg/providers/aws/cluster_network_provider_test.go
```diff
@@ -2195,3 +2259,70 @@ func TestNetworkProvider_DeleteNetworkConnection(t *testing.T) {
...
+		ctx context.Context
+	}
+	tests := []struct {
+		name    string
+		fields  fields
+		args    args
+		wantErr bool
+	}{
+		{
+			name: "return nil on successful deletion of bundled vpc resources",
```
ciaranRoche, line 2285: As we rely on `awsError` for `notFoundFault` should we add a test case to capture this logic. 
Also maybe a test case where we call this delete function and SubnetGroups and the SecurityGroup exist.

pkg/providers/aws/provider_redis_test.go
```diff
@@ -470,7 +524,13 @@ func TestAWSRedisProvider_deleteRedisCluster(t *testing.T) {
 				Logger:            testLogger,
 				CredentialManager: &CredentialManagerMock{},
 				ConfigManager:     &ConfigManagerMock{},
-				CacheSvc:          &mockElasticacheClient{replicationGroups: buildReplicationGroupPending()},
+				CacheSvc: buildMockElasticacheClient(func(elasticacheClient *mockElasticacheClient) {
+					elasticacheClient.describeReplicationGroupsFn = func(*elasticache.DescribeReplicationGroupsInput) (*elasticache.DescribeReplicationGroupsOutput, error) {
+						return &elasticache.DescribeReplicationGroupsOutput{
+							ReplicationGroups: buildReplicationGroupPending(),
```
ciaranRoche, line 530: If we have time, i think we should remove/refactor `buildReplicationGroupPending` to the new approach

pkg/providers/aws/provider_redis_test.go
```diff
@@ -489,7 +549,13 @@ func TestAWSRedisProvider_deleteRedisCluster(t *testing.T) {
 				Logger:            testLogger,
 				CredentialManager: &CredentialManagerMock{},
 				ConfigManager:     &ConfigManagerMock{},
-				CacheSvc:          &mockElasticacheClient{replicationGroups: buildReplicationGroupReady()},
+				CacheSvc: buildMockElasticacheClient(func(elasticacheClient *mockElasticacheClient) {
+					elasticacheClient.describeReplicationGroupsFn = func(*elasticache.DescribeReplicationGroupsInput) (*elasticache.DescribeReplicationGroupsOutput, error) {
+						return &elasticache.DescribeReplicationGroupsOutput{
+							ReplicationGroups: buildReplicationGroupReady(),
```
ciaranRoche, line 555: Same as above we can refactor these ReplicationGroup functions into a single function with an override.

pkg/providers/aws/provider_redis_test.go
```diff
@@ -604,7 +696,13 @@ func TestAWSRedisProvider_TagElasticache(t *testing.T) {
 			args: args{
 				ctx:      context.TODO(),
 				r:        buildTestRedisCR(),
-				cacheSvc: &mockElasticacheClient{replicationGroups: buildReplicationGroupReady()},
+				cacheSvc: buildMockElasticacheClient(func(elasticacheClient *mockElasticacheClient) {
+					elasticacheClient.describeCacheClustersFn = func(input *elasticache.DescribeCacheClustersInput) (*elasticache.DescribeCacheClustersOutput, error) {
+						return &elasticache.DescribeCacheClustersOutput{
+							CacheClusters: buildCacheClusterList(),
```
ciaranRoche, line 702: Same as the others, if there is time, we should refactor this `buildCacheClusterList`

pkg/providers/aws/provider_redissnapshot_test.go
```diff
@@ -34,7 +34,9 @@ type elasticacheClientMock struct {
 	DescribeReplicationGroupsFunc func(in1 *elasticache.DescribeReplicationGroupsInput) (*elasticache.DescribeReplicationGroupsOutput, error)
 	CreateSnapshotFunc            func(in1 *elasticache.CreateSnapshotInput) (*elasticache.CreateSnapshotOutput, error)
 	DeleteSnapshotFunc            func(in1 *elasticache.DeleteSnapshotInput) (*elasticache.DeleteSnapshotOutput, error)
-	calls                         struct {
+	DeleteCacheSubnetGroupFn      func(in1 *elasticache.DeleteCacheSubnetGroupInput) (*elasticache.DeleteCacheSubnetGroupOutput, error)
```
ciaranRoche, line 37: I know you called this out, but I think we should only have one elasticacheMock struct, i think it would be valuable to refactor this struct into the other existing struct.

laurafitzgerald:
@ciaranRoche cheers for the review. There's definitely time so I'll implement those changes. Some of them were on my hit list anyway. Just didn't want to refactoring massive amounts here without a feedback loop. Will progress with those changes now. 👍

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -685,6 +687,58 @@ func (n *NetworkProvider) IsEnabled(ctx context.Context) (bool, error) {
...
+	// all in the account and then find the one with the correct group name and then request deletion
+	// using the group id of the matched security group
+	securityGroup, err := getSecurityGroup(n.Ec2Api, securityGroupName)
+	if err != nil {
+		return errorUtil.Wrap(err, "error getting ec2 security group")
+	}
+	if securityGroup != nil {
+		_, err = n.Ec2Api.DeleteSecurityGroup(&ec2.DeleteSecurityGroupInput{
+			GroupId: securityGroup.GroupId,
+		})
```
aidenkeating, line 737: Should we be handling this `err`?

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -685,6 +687,58 @@ func (n *NetworkProvider) IsEnabled(ctx context.Context) (bool, error) {
...
+	if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() != elasticache.ErrCodeCacheSubnetGroupNotFoundFault {
+		return errorUtil.Wrap(err, "error deleting elasticache subnet group")
+	}
+	logger.Infof("deleting bundled rds subnet group %s, if it's not found it's already deleted and will continue", subnetGroupName)
+	_, err = n.RdsApi.DeleteDBSubnetGroup(&rds.DeleteDBSubnetGroupInput{
+		DBSubnetGroupName: aws.String(subnetGroupName),
+	})
+	if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() != rds.ErrCodeDBSubnetGroupNotFoundFault {
+		return errorUtil.Wrap(err, "error deleting rds subnet group")
+	}
```
aidenkeating, line 719: What if the error isn't an AWS err? e.g. `ok` is `false`? Maybe something like

```suggestion
	if awsErr, ok := err.(awserr.Error); !ok || awsErr.Code() != rds.ErrCodeDBSubnetGroupNotFoundFault {
		return errorUtil.Wrap(err, "error deleting rds subnet group")
	}
```

pkg/providers/aws/provider_redis_test.go
```diff
@@ -44,13 +45,16 @@ type mockElasticacheClient struct {
...
 	// to allow for simple overrides in test table declarations
-	modifyCacheSubnetGroupFn func(*elasticache.ModifyCacheSubnetGroupInput) (*elasticache.ModifyCacheSubnetGroupOutput, error)
+	modifyCacheSubnetGroupFn    func(*elasticache.ModifyCacheSubnetGroupInput) (*elasticache.ModifyCacheSubnetGroupOutput, error)
+	deleteCacheSubnetGroupFn    func(*elasticache.DeleteCacheSubnetGroupInput) (*elasticache.DeleteCacheSubnetGroupOutput, error)
+	describeCacheSubnetGroupsFn func(*elasticache.DescribeCacheSubnetGroupsInput) (*elasticache.DescribeCacheSubnetGroupsOutput, error)
+	describeCacheClustersFn     func(*elasticache.DescribeCacheClustersInput) (*elasticache.DescribeCacheClustersOutput, error)
+	describeReplicationGroupsFn func(*elasticache.DescribeReplicationGroupsInput) (*elasticache.DescribeReplicationGroupsOutput, error)
 }
 
 func buildMockElasticacheClient(modifyFn func(*mockElasticacheClient)) *mockElasticacheClient {
```
aidenkeating, line 60: I've noticed a lot of tests return `&elasticache.DescribeCacheSubnetGroupsOutput{}` when overriding `describeCacheSubnetGroupsFn`. That seems like a sane default function for a mock. So maybe we could do something like?

```suggestion
func buildMockElasticacheClient(modifyFn func(*mockElasticacheClient)) *mockElasticacheClient {
	mock := &mockElasticacheClient{
	        describeCacheSubnetGroupsFn: // function that returns &elasticache.DescribeCacheSubnetGroupsOutput{}, nil
	 }
	if modifyFn != nil {
		modifyFn(mock)
	}
	return mock
}
```

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -685,6 +687,60 @@ func (n *NetworkProvider) IsEnabled(ctx context.Context) (bool, error) {
...
+// resources. The majority of the functionality in this file relates to standalone aws vpc and it's
+// resources.
+func (n *NetworkProvider) DeleteBundledCloudResources(ctx context.Context) error {
+	logger := n.Logger.WithField("action", "deleteBundledCloudResources")
+
+	subnetGroupName, err := BuildInfraName(ctx, n.Client, "subnetgroup", DefaultAwsIdentifierLength)
+	if err != nil {
+		return errorUtil.Wrap(err, "error building bundle subnet group resource name on deletion")
+	}
+	logger.Infof("deleting bundled elasticcache subnet group %s, if it's not found it's already deleted and will continue", subnetGroupName)
```
darahayes, line 706: minor nit: typo in the word `elasticache`

pkg/providers/aws/provider_redissnapshot_test.go
darahayes: This was a bit of a mess that I created so If you want to create a separate ticket for it I would be happy to do the refactor 🙂

darahayes: This was a bit of a mess that I created so If you want to create a separate ticket for it I would be happy to do the refactor 🙂

pkg/providers/aws/provider_redissnapshot.go
```diff
@@ -114,7 +114,7 @@ func (p *RedisSnapshotProvider) createRedisSnapshot(ctx context.Context, snapsho
 		ReplicationGroupId: aws.String(clusterName),
 	})
 
-	if cacheOutput == nil {
+	if len(cacheOutput.ReplicationGroups) == 0 {
```
laurafitzgerald, line 117: I'm not sure about this change but based on the return of an empty list it's possible for this check to pass and the following check on the 0th array would be a nil check.

laurafitzgerald: I should say I haven't verified this tho. I can remove it and create a separate JIRA if necesary.

README.md
```diff
@@ -22,6 +22,12 @@ Operator is running.
 ## Running the Cloud Resource Operator
 ## Locally
 
+## Supported Openshift Versions
+
+Due to a change in how networking is configured for Openshift >= v4.4.6 the use of cro <= v0.16.1 with these Openshift versions is unsupported.
+Please use >= v0.17.x of CRO for Openshift >= v4.4.6.
+
+
```
aidenkeating, line 30: 👍

pkg/providers/aws/provider_redissnapshot.go
laurafitzgerald: we also not handling the error here, because there's some more changes needed here I'm going to pull this out and create a seperate JIRA.

darahayes: @laurafitzgerald I think this change would require some testing to check. I believe it's checking if `cacheOutput` is nil because AWS tends to return a `nil` and a `NotFoundError` if you try to find something by ID and it doesn't exist. So if that's the case then the change to `len(cacheOutput.ReplicationGroups)` would actually cause a panic because `cacheOutput` will be nil. 

You can see I had the same conversation with @ciaranRoche not so long ago https://github.com/integr8ly/cloud-resource-operator/pull/155/files#r446083471

This code does cause confusion, if you really wanted to reduce that confusion you could add a proper check on the error object, which would look something like this https://github.com/integr8ly/cloud-resource-operator/pull/155/files#diff-6b187e80c9ce1e3241605d642acb94c0R206-R209

laurafitzgerald: Hi @darahayes yes agreed. I've removed it from this pr altogether so as not to block the current verification. After chatting with @ciaranRoche and your comments above it may not be just this one change. I'll go ahead and create a JIRA for it. Thanks for the extra detail. 👍

laurafitzgerald:
Hey folks. I've pushed a final commit here which should address the final feedback item from @ciaranRoche above. If you have some time to review that would fab. 😄

pkg/providers/aws/provider_redissnapshot.go
laurafitzgerald: JIRA created: https://issues.redhat.com/browse/INTLY-9446 Resolving this discussion. 👍

ciaranRoche:
@laurafitzgerald are we safe to remove [wip] from this?

PR #159, opened 2020-07-09
feat: expose deletion timestamp metric for redis and postgres

darahayes:
## Overview

Jira: <jira-id>

I'm throwing up this PR early to get some feedback. I have some metrics printing but I'm not really sure if it's what we actually want. Take a look at the examples below and let me know. To explain them a little bit. I basically followed the same format that the `kube_pod_deletion_timestamp` metric follows https://github.com/kubernetes/kube-state-metrics/blob/0bfc2981f9c281c78e33052abdc2d621630562b9/internal/store/pod.go#L200-L218

It's a gauge with some labels that identify the resource and then the numerical value is `(float64(cr.DeletionTimestamp.Unix()))`. One of the labels is `statusPhase` which will either be `"deletion in progress"` or it will be `"failed"` if the delete has failed.

Below are examples of the metrics that are printed when a deletion is triggered.

## Postgres Deletion Timestamp Metric Example

Delete in Progress
```
# HELP cro_postgres_deletion_timestamp 
# TYPE cro_postgres_deletion_timestamp gauge
cro_postgres_deletion_timestamp{clusterID="dhayes-mw-colla-gd7xc",instanceID="dhayesmwcollagd7xcdaracloudresource-zjme",namespace="dara-cloud-resources-operator",productName="productName",resourceID="dhayes-postgres",statusPhase="deletion in progress",strategy="aws-rds"} 1.59421868e+09
```

Delete has failed
```
# HELP cro_postgres_deletion_timestamp 
# TYPE cro_postgres_deletion_timestamp gauge
cro_postgres_deletion_timestamp{clusterID="dhayes-mw-colla-gd7xc",instanceID="dhayesmwcollagd7xcdaracloudresource-zjme",namespace="dara-cloud-resources-operator",productName="productName",resourceID="dhayes-postgres",statusPhase="failed",strategy="aws-rds"} 1.59421868e+09
```

### Redis Deletion Timestamp Metric Example

Delete in progress
```
cro_redis_deletion_timestamp{clusterID="dhayes-mw-colla-gd7xc",instanceID="dhayesmwcollagd7xcdaracloudresource-mqsx",namespace="dara-cloud-resources-operator",productName="productName",resourceID="dhayes-redis",statusPhase="deletion in progress",strategy="aws-elasticache"} 1.594308846e+09
```

Delete has failed
```
cro_redis_deletion_timestamp{clusterID="dhayes-mw-colla-gd7xc",instanceID="dhayesmwcollagd7xcdaracloudresource-mqsx",namespace="dara-cloud-resources-operator",productName="productName",resourceID="dhayes-redis",statusPhase="failed",strategy="aws-elasticache"} 1.594308846e+09
```

## Verification

* Run this branch locally
* Create a Redis and/or Postgres instance

Two ways I know of the cause deletes to fail:

* modifying the Describe calls in the delete methods for each respective provider to always return an error. i.e. the `DescribeDbInstances` call.

* If your local instance of CRO is reconciling only one CR, e.g. a postgres instance, and you try to delete it. The local CRO instance will fail to clean it up as it thinks there are no other CRs left and will unsuccessfully try to clean up the VPC and all of its networking. This will fail because there is still another instance of CRO running within the actual cluster that has resources within the VPC.

Verify that the metrics described above are shown in the output when you call the metrics endpoint


pkg/providers/aws/provider_postgres.go
```diff
@@ -965,6 +969,30 @@ func (p *PostgresProvider) exposePostgresMetrics(ctx context.Context, cr *v1alph
...
+
+		// get Cluster Id
+		logrus.Info("setting postgres information metric")
+		clusterID, err := resources.GetClusterID(ctx, p.Client)
+		if err != nil {
+			logrus.Errorf("failed to get cluster id while exposing information metric for %v", instanceName)
+			return
+		}
+
+		labels := buildPostgresStatusMetricsLabels(cr, clusterID, instanceName, cr.Status.Phase)
```
ciaranRoche, line 991: Using the status metric would mean there could be potentially multiple metrics exposed for the one resource as the phase changes. 

Would it be worth a single metric using the build generic label function. 

We can build an alert on the basis of if we have a present `cro_deletion_timestamp` metric and a `cro_phase_failed` metric = 1

darahayes: Yeah I guess so. I'm not really sure to be honest 😅

My thinking was more along the lines of an alert could look for a `cro_deletion_timestamp` metric that included the label `statusPhase="failed"` which is what will happen with this implementation, but maybe I don't understand how to build alerts.

darahayes: Had a call with @david-martin and he mentioned that the query `cro_postgres_deletion_timestamp{statusPhase="failed"}` should work.

ciaranRoche: cool cool, i wasn't 100% sure either :smile_cat:

austincunningham:
verified on Friday

PR #161, opened 2020-07-16
Add aws postgres metric provider

ciaranRoche:
## Overview

Jira: https://issues.redhat.com/browse/INTLY-7224

## Verification
**Note :** This work contains redis metric provider, which is just boilerplate and will be handled in https://issues.redhat.com/browse/INTLY-7740

- run `make cluster/prepare`
- run `make cluster/seed/managed/postgres`
- run `make run`
- wait a few minutes for the instance to begin provisioning before cloud watch metrics are exposed
- verify the following debug statements are logged : 
```
DEBU[0008] {free_storage_space_average map[clusterID:hcd-rhmi-clouds-6pdvc instanceID:hcdrhmiclouds6pdvccloudresourceoper-eu6t namespace:cloud-resource-operator productName:productName resourceID:example-postgres strategy:aws-rds] 2.0676653056e+10}  controller=controller_cloudmetrics
DEBU[0008] {free_storage_space_maximum map[clusterID:hcd-rhmi-clouds-6pdvc instanceID:hcdrhmiclouds6pdvccloudresourceoper-eu6t namespace:cloud-resource-operator productName:productName resourceID:example-postgres strategy:aws-rds] 2.0710719488e+10}  controller=controller_cloudmetrics
DEBU[0008] {free_storage_space_minimum map[clusterID:hcd-rhmi-clouds-6pdvc instanceID:hcdrhmiclouds6pdvccloudresourceoper-eu6t namespace:cloud-resource-operator productName:productName resourceID:example-postgres strategy:aws-rds] 2.0659613696e+10}  controller=controller_cloudmetrics

```


pkg/controller/cloudmetrics/cloudmetrics_controller.go
```diff
@@ -77,47 +84,98 @@ var _ reconcile.Reconciler = &ReconcileCloudMetrics{}
...
+	}
+
+	// Check the CR for existing Strategy
+	strategyToUse := stratMap.Postgres
+	if postgres.Status.Strategy != "" {
+		strategyToUse = postgres.Status.Strategy
+		if strategyToUse != stratMap.Postgres {
+			r.logger.Infof("strategy and provider already set, changing of cloud-resource-config config maps not allowed in existing installation. the existing strategy is '%s' , cloud-resource-config is now set to '%s'. operator will continue to use existing strategy", strategyToUse, stratMap.Postgres)
+		}
+	}
```
aidenkeating, line 146: For metrics do we need to do this? Is checking the CR not enough? If the strategy isn't set in the CR we could just skip it

pkg/controller/cloudmetrics/cloudmetrics_controller.go
```diff
@@ -77,47 +84,98 @@ var _ reconcile.Reconciler = &ReconcileCloudMetrics{}
...
+			continue
+		}
+
+		// only scrape and expose metrics in an annotation is present
+		// this is to ensure the resource exists prior to calling the cloud watch api
+		// to reduce api calls, if alerts are present they will fire if metric is absent catching any issues to a resource not existing
+		if !annotations.Has(postgres, aws.ResourceIdentifierAnnotation) {
+			r.logger.Infof("scraping rds cloudwatch metrics, could not find rds resource identifier for postgres %s", postgres.Name)
+			continue
+		}
```
aidenkeating, line 160: This seems like something that should be down at the AWS level? We don't set that annotation in the OpenShift provider, there's no guarantee we'll do it for GCP. We're also mentioning RDS at the controller level in the log

pkg/providers/types.go
```diff
@@ -48,6 +48,32 @@ type RedisSnapshotInstance struct {
...
+	Timestamp  *time.Time
+}
+
+type ScrapeMetricsInput struct {
+	Context    context.Context
+	Name       string
+	Namespace  string
+	Tier       string
+	ResourceID string
+}
```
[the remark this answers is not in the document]
ciaranRoche: yeah I like this idea, we originally had the annotation field down in provider level, but moved it up to allow for this generic struct. Fetching the CR at the provider level is a nicer approach. 

What do you think about removing the use of the annotation all together and just generate the resource id? :thinking:

pkg/controller/cloudmetrics/cloudmetrics_controller.go
```diff
@@ -77,47 +83,72 @@ var _ reconcile.Reconciler = &ReconcileCloudMetrics{}
...
+func (r *ReconcileCloudMetrics) reconcilePostgresMetrics(ctx context.Context, postgres *integreatlyv1alpha1.Postgres) error {
+	for _, p := range r.metricProviderList {
+		// only scrape metrics on supported strategies
+		if !p.SupportsStrategy(postgres.Status.Strategy) {
+			continue
+		}
+
+		// all postgres metric providers inherit the same interface
+		// scrapeMetrics returns scraped metrics output which contains a list of GenericCloudMetrics
+		scrapedMetricsOutput, err := p.ScrapeMetrics(ctx, &providers.ScrapeMetricsInput{
```
darahayes, line 140: Question/Observation:

Do we want the `MetricsProvider` interface to be as generic as it is here? We're passing down the name + namespace of the resource to make another API call to fetch the resource within the provider when we already have it. Would it not be a bit easier to deal with if we had separate interfaces like `PostgresMetricProvider`, `RedisMetricProvider` and maybe `BlobStorageMetricProvider` and then the implementations for those would be something like `AWSPostgresMetricProvider`, `AWSRedisMetricProvider` and so on for different cloud providers in the future?

I see the `Postgres`, `Redis` and `BlobStorage` CRs as the top level interfaces we deal with and then we have providers for each of those. To me it would kinda make sense that at a similar level you'd have `PostgresMetricProvider`, `RedisMetricProvider` and so on. It would mean that in future you could define what is expected at a high level in terms of metrics for a Postgres or for a Redis and you could ensure that it's the same across all cloud providers.

pkg/controller/cloudmetrics/cloudmetrics_controller.go
```diff
@@ -77,47 +83,72 @@ var _ reconcile.Reconciler = &ReconcileCloudMetrics{}
...
 	return reconcile.Result{
-		RequeueAfter: watchDuration * time.Second,
+		RequeueAfter: watchDuration,
 	}, nil
 }
+
+func (r *ReconcileCloudMetrics) reconcilePostgresMetrics(ctx context.Context, postgres *integreatlyv1alpha1.Postgres) error {
+	for _, p := range r.metricProviderList {
+		// only scrape metrics on supported strategies
+		if !p.SupportsStrategy(postgres.Status.Strategy) {
```
darahayes, line 134: In relation to my previous comment, I think this isn't going to work if the `r.metricProviderList` contains metric providers for both AWS redis and postgres instances. The AWS Redis Metrics Provider is going to return true for `SupportsStrategy` also.

If you think of the `MetricProvider` interface, it has a matrix of ways it can be implemented. It can be either a Postgres or a Redis and for both of those it can be either OpenShift, AWS, Azure etc. I think it might make more sense to declare interfaces like `PostgresMetricProvider` and `RedisMetricProvider` and then this controller would have two lists like so:

```
type ReconcileCloudMetrics struct {
	client client.Client		client             client.Client
	scheme *runtime.Scheme		scheme             *runtime.Scheme
	logger *logrus.Entry		logger             *logrus.Entry
	postgresMetricProviderList []providers.PostgresMetricsProvider
    redusMetricProviderList    []providers.RedisMetricsProvider
}
```

pkg/providers/aws/provider_postgres_metrics.go
```diff
@@ -0,0 +1,211 @@
...
+// ScrapeMetrics returns scraped metrics to metric controller
+func (p PostgresMetricsProvider) ScrapeMetrics(ctx context.Context, postgresInput *providers.ScrapeMetricsInput) (*providers.ScrapeMetricsOutput, error) {
+	logger := resources.NewActionLoggerWithFields(p.Logger, map[string]interface{}{
+		resources.LoggingKeyAction: "ScrapeMetrics",
+		"Resource":                 postgresInput.Name,
+	})
+	logger.Infof("reconciling postgres metrics %s", postgresInput.Name)
+
+	postgres := &v1alpha1.Postgres{}
+	err := p.Client.Get(ctx, types.NamespacedName{Name: postgresInput.Name, Namespace: postgresInput.Namespace}, postgres)
```
darahayes, line 88: Relates to my previous comment. This API call wouldn't be needed if we were implementing an interface like `PostgresMetricsProvider` rather than just `MetricsProvider` because the `ScrapeMetrics` function could accept the Postgres CR

pkg/controller/cloudmetrics/cloudmetrics_controller.go
aidenkeating: @darahayes Good catch, though the status block of a custom resource has a provider flag. We can just use that instead of the strategy itself. I don't really have a major opinion either way on the multiple list/interface approach.

pkg/providers/aws/provider_postgres_metrics.go
```diff
@@ -0,0 +1,211 @@
...
+	{
+		PromethuesMetricName: "cro_postgres_free_storage_average_minimum",
+		ProviderMetricName:   "FreeStorageSpace",
+		Statistic:            "Minimum",
+	},
+}
+
+var _ providers.MetricsProvider = (*PostgresMetricsProvider)(nil)
+
+type PostgresMetricsProvider struct {
```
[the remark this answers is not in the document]
aidenkeating: In Go it's generally recommended not to use the name of the package at the start of the struct. I think as long as the struct is scoped to the aws package, there shouldn't be much confusion. Again, no major opinion either away.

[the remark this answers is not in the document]
darahayes: Good point. I'm still relatively new to Go (again) so I am just throwing stuff out there 😅

pkg/controller/cloudmetrics/cloudmetrics_controller.go
aidenkeating: Thinking about this some more, I think the suggestion above is the approach I'd also take if I were writing it. If there's no objections from yourselves @laurafitzgerald @austincunningham @ciaranRoche I think it makes sense to make that change in this PR?

laurafitzgerald:
@ciaranRoche et al. The last commit is just a pitstop commit as I'm finishing up for the day and wanted to push what I had done in a spare slot of time. It can be dropped if needed. Just wanted to understand the comments from Dara above so added the redis piece to see what would happen with the interfaces. I see the problem with the interface in that for the RedisMetricsProvider you wont be implementing the function `ScrapeRDSMetrics` and vice versa so you can't have each of those implement the generic `MetricsProvider` as we were doing. I also refactored the ScrapeRDSMetrics call to pass the cr rather than the input struct we were using. Hack, drop or refactor that commit as you want. 👍

pkg/controller/cloudmetrics/cloudmetrics_controller.go
```diff
@@ -77,47 +78,95 @@ var _ reconcile.Reconciler = &ReconcileCloudMetrics{}
...
+	redisProviderList    []providers.RedisMetricsProvider
 }
 
 // Reconcile reads all redis and postgres crs periodically and reconcile metrics for these
 // resources.
 // The Controller will requeue the Request every 5 minutes constantly when RequeueAfter is set
 func (r *ReconcileCloudMetrics) Reconcile(request reconcile.Request) (reconcile.Result, error) {
-	r.logger.Info("Reconciling CloudMetrics")
+	r.logger.Info("reconciling CloudMetrics")
+	ctx := context.TODO()
```
aidenkeating, line 93: Let's use `context.Background()`, it's effectively the same but shows the intention that we aren't going to change it any time soon

pkg/controller/cloudmetrics/cloudmetrics_controller.go
```diff
@@ -77,47 +78,95 @@ var _ reconcile.Reconciler = &ReconcileCloudMetrics{}
...
+		// scrapeMetrics returns scraped metrics output which contains a list of GenericCloudMetrics
+		scrapedMetricsOutput, err := p.ScrapeRDSMetrics(ctx, postgres)
+		if err != nil {
+			r.logger.Errorf("failed to scrape metrics for postgres %v", err)
+			continue
+		}
+
+		for _, metricData := range scrapedMetricsOutput.Metrics {
+			r.logger.Debug(*metricData)
+		}
```
aidenkeating, line 147: I'm guessing we're planning to do something with these metrics (e.g. return them out of here) in a future ticket?

pkg/providers/aws/provider_postgres_metrics.go
```diff
@@ -0,0 +1,203 @@
...
+	{
+		PromethuesMetricName: "cro_postgres_free_storage_average_maximum",
+		ProviderMetricName:   "FreeStorageSpace",
+		Statistic:            "Maximum",
+	},
+	{
+		PromethuesMetricName: "cro_postgres_free_storage_average_minimum",
+		ProviderMetricName:   "FreeStorageSpace",
+		Statistic:            "Minimum",
+	},
```
aidenkeating, line 53: What's the reasoning behind the min and max statistics? Is there an additional cost for this?

pkg/controller/cloudmetrics/cloudmetrics_controller.go
austincunningham: :+1:

pkg/controller/cloudmetrics/cloudmetrics_controller.go
austincunningham: at this point this is mainly here for verification, but I do see future uses for it.

pkg/providers/aws/provider_postgres_metrics.go
laurafitzgerald: After discussion remove the min and the max.

ciaranRoche:
Verified -> 
image

PR #163, opened 2020-07-20
INTLY-7740 add redis scrape on DatabaseMemoryUsagePercentage

austincunningham:
## Overview

Jira: https://issues.redhat.com/browse/INTLY-7740
https://issues.redhat.com/browse/INTLY-7227
https://issues.redhat.com/browse/INTLY-7228

## Verification

- Clone this branch
- Run `make cluster/prepare`
- Run `make run`
- Wait for Redis to provision 
- Ensure the following metrics are present using curl
```bach
 curl http://localhost:8383/metrics | grep _average  
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
# HELP cro_redis_cpu_utilization_average The percentage of CPU utilization. Units: Percent
# TYPE cro_redis_cpu_utilization_average gauge
cro_redis_cpu_utilization_average{clusterID="mw-collab-byoc-v8926",instanceID="mwcollabbyocv8926cloudresourceopera-mnas-001",namespace="cloud-resource-operator",productName="productName",resourceID="example-redis",strategy="aws-elasticache"} 1
cro_redis_cpu_utilization_average{clusterID="mw-collab-byoc-v8926",instanceID="mwcollabbyocv8926cloudresourceopera-mnas-002",namespace="cloud-resource-operator",productName="productName",resourceID="example-redis",strategy="aws-elasticache"} 1
# HELP cro_redis_freeable_memory_average The amount of available random access memory. Units: Bytes
# TYPE cro_redis_freeable_memory_average gauge
cro_redis_freeable_memory_average{clusterID="mw-collab-byoc-v8926",instanceID="mwcollabbyocv8926cloudresourceopera-mnas-001",namespace="cloud-resource-operator",productName="productName",resourceID="example-redis",strategy="aws-elasticache"} 6.366068736e+08
cro_redis_freeable_memory_average{clusterID="mw-collab-byoc-v8926",instanceID="mwcollabbyocv8926cloudresourceopera-mnas-002",namespace="cloud-resource-operator",productName="productName",resourceID="example-redis",strategy="aws-elasticache"} 6.29112832e+08
# HELP cro_redis_memory_usage_percentage_average The percentage of redis used memory. Units: Bytes
# TYPE cro_redis_memory_usage_percentage_average gauge
cro_redis_memory_usage_percentage_average{clusterID="mw-collab-byoc-v8926",instanceID="mwcollabbyocv8926cloudresourceopera-mnas-001",namespace="cloud-resource-operator",productName="productName",resourceID="example-redis",strategy="aws-elasticache"} 1.3509599367777505
cro_redis_memory_usage_percentage_average{clusterID="mw-collab-byoc-v8926",instanceID="mwcollabbyocv8926cloudresourceopera-mnas-002",namespace="cloud-resource-operator",productName="productName",resourceID="example-redis",strategy="aws-elasticache"} 1.3540788491566977
100 36832    0 36832    0     0  35.1M      0 --:--:-- --:--:-- --:--:-- 35.1M
```

### DIFF
--- a/pkg/controller/cloudmetrics/cloudmetrics_controller.go
+++ b/pkg/controller/cloudmetrics/cloudmetrics_controller.go
@@ -29,6 +29,9 @@ const (
 	postgresCPUUtilizationAverage = "cro_postgres_cpu_utilization_average"
 	postgresFreeableMemoryAverage = "cro_postgres_freeable_memory_average"
 
+	redisMemoryUsagePercentageAverage = "cro_redis_memory_usage_percentage_average"
+	redisFreeableMemoryAverage        = "cro_redis_freeable_memory_average"
+
 	labelClusterIDKey   = "clusterID"
 	labelResourceIDKey  = "resourceID"
 	labelNamespaceKey   = "namespace"
@@ -109,7 +112,41 @@ var postgresGaugeMetrics = []CroGaugeMetric{
 
 // redisGaugeMetrics stores a mapping between an exposed (redis) prometheus metric and multiple cloud provider specific metric
 // to add any addition metrics simply add to this mapping and it will be scraped and exposed
-var redisGaugeMetrics []CroGaugeMetric
+var redisGaugeMetrics = []CroGaugeMetric{
+	{
+		Name: redisMemoryUsagePercentageAverage,
+		GaugeVec: prometheus.NewGaugeVec(
+			prometheus.GaugeOpts{
+				Name: redisMemoryUsagePercentageAverage,
+				Help: "The percentage of redis used memory. Units: Bytes",
+			},
+			labels),
+		ProviderType: map[string]providers.CloudProviderMetricType{
+			providers.AWSDeploymentStrategy: {
+				PromethuesMetricName: redisMemoryUsagePercentageAverage,
+				//calculated on used_memory/maxmemory from Redis INFO http://redis.io/commands/info
+				ProviderMetricName: "DatabaseMemoryUsagePercentage",
+				Statistic:          cloudwatch.StatisticAverage,
+			},
+		},
+	},
+	{
+		Name: redisFreeableMemoryAverage,
+		GaugeVec: prometheus.NewGaugeVec(
+			prometheus.GaugeOpts{
+				Name: redisFreeableMemoryAverage,
+				Help: "The amount of available random access memory. Units: Bytes",
+			},
+			labels),
+		ProviderType: map[string]providers.CloudProviderMetricType{
+			providers.AWSDeploymentStrategy: {
+				PromethuesMetricName: redisFreeableMemoryAverage,
+				ProviderMetricName:   "FreeableMemory",
+				Statistic:            cloudwatch.StatisticAverage,
+			},
+		},
+	},
+}
 
 // Add creates a new CloudMetrics Controller and adds it to the Manager. The Manager will set fields on the Controller
 // and Start it when the Manager is Started.
@@ -206,16 +243,41 @@ func (r *ReconcileCloudMetrics) Reconcile(request reconcile.Request) (reconcile.
 			if !p.SupportsStrategy(redis.Status.Strategy) {
 				continue
 			}
+			var redisMetricTypes []providers.CloudProviderMetricType
+			for _, gaugeMetric := range redisGaugeMetrics {
+				for provider, metricType := range gaugeMetric.ProviderType {
+					if provider == redis.Status.Strategy {
+						redisMetricTypes = append(redisMetricTypes, metricType)
+						continue
+					}
+				}
+			}
 
 			// all redis scrapedMetric providers inherit the same interface
 			// scrapeMetrics returns scraped metrics output which contains a list of GenericCloudMetrics
-			scrapedMetricsOutput, err := p.ScrapeRedisMetrics(ctx, &redis)
+			scrapedMetricsOutput, err := p.ScrapeRedisMetrics(ctx, &redis, redisMetricTypes)
 			if err != nil {
 				r.logger.Errorf("failed to scrape metrics for redis %v", err)
 				continue
 			}
 
 			scrapedMetrics = append(scrapedMetrics, scrapedMetricsOutput.Metrics...)
+		}
+	}
+	// for each scraped metric value we check redisGaugeMetrics for a match and set the value and labels
+	for _, scrapedMetric := range scrapedMetrics {
+		for _, croMetric := range redisGaugeMetrics {
+			if scrapedMetric.Name == croMetric.Name {
+				croMetric.GaugeVec.WithLabelValues(
+					scrapedMetric.Labels[labelClusterIDKey],
+					scrapedMetric.Labels[labelResourceIDKey],
+					scrapedMetric.Labels[labelNamespaceKey],
+					scrapedMetric.Labels[labelInstanceIDKey],
+					scrapedMetric.Labels[labelProductNameKey],
+					scrapedMetric.Labels[labelStrategyKey]).Set(scrapedMetric.Value)
+				r.logger.Infof("successfully set redis metric value for %s", croMetric.Name)
+				continue
+			}
 		}
 	}
 
@@ -268,7 +330,7 @@ func (r *ReconcileCloudMetrics) Reconcile(request reconcile.Request) (reconcile.
 					scrapedMetric.Labels[labelInstanceIDKey],
 					scrapedMetric.Labels[labelProductNameKey],
 					scrapedMetric.Labels[labelStrategyKey]).Set(scrapedMetric.Value)
-				r.logger.Infof("successfully set metric value for %s", croMetric.Name)
+				r.logger.Infof("successfully set postgres metric value for %s", croMetric.Name)
 				continue
 			}
 		}
--- a/pkg/controller/cloudmetrics/cloudmetrics_controller.go
+++ b/pkg/controller/cloudmetrics/cloudmetrics_controller.go
@@ -31,6 +31,7 @@ const (
 
 	redisMemoryUsagePercentageAverage = "cro_redis_memory_usage_percentage_average"
 	redisFreeableMemoryAverage        = "cro_redis_freeable_memory_average"
+	redisCPUUtilizationAverage = "cro_redis_cpu_utilization_average"
 
 	labelClusterIDKey   = "clusterID"
 	labelResourceIDKey  = "resourceID"
@@ -142,6 +143,22 @@ var redisGaugeMetrics = []CroGaugeMetric{
 			providers.AWSDeploymentStrategy: {
 				PromethuesMetricName: redisFreeableMemoryAverage,
 				ProviderMetricName:   "FreeableMemory",
+				Statistic:            cloudwatch.StatisticAverage,
+			},
+		},
+	},
+	{
+		Name: redisCPUUtilizationAverage,
+		GaugeVec: prometheus.NewGaugeVec(
+			prometheus.GaugeOpts{
+				Name: redisCPUUtilizationAverage,
+				Help: "The percentage of CPU utilization. Units: Percent",
+			},
+			labels),
+		ProviderType: map[string]providers.CloudProviderMetricType{
+			providers.AWSDeploymentStrategy: {
+				PromethuesMetricName: redisCPUUtilizationAverage,
+				ProviderMetricName:   "CPUUtilization",
 				Statistic:            cloudwatch.StatisticAverage,
 			},
 		},
@@ -265,21 +282,7 @@ func (r *ReconcileCloudMetrics) Reconcile(request reconcile.Request) (reconcile.
 		}
 	}
 	// for each scraped metric value we check redisGaugeMetrics for a match and set the value and labels
-	for _, scrapedMetric := range scrapedMetrics {
-		for _, croMetric := range redisGaugeMetrics {
-			if scrapedMetric.Name == croMetric.Name {
-				croMetric.GaugeVec.WithLabelValues(
-					scrapedMetric.Labels[labelClusterIDKey],
-					scrapedMetric.Labels[labelResourceIDKey],
-					scrapedMetric.Labels[labelNamespaceKey],
-					scrapedMetric.Labels[labelInstanceIDKey],
-					scrapedMetric.Labels[labelProductNameKey],
-					scrapedMetric.Labels[labelStrategyKey]).Set(scrapedMetric.Value)
-				r.logger.Infof("successfully set redis metric value for %s", croMetric.Name)
-				continue
-			}
-		}
-	}
+	r.setGaugeMetrics(redisGaugeMetrics, scrapedMetrics)
 
 	// Fetch all postgres crs
 	postgresInstances := &integreatlyv1alpha1.PostgresList{}
@@ -320,21 +323,7 @@ func (r *ReconcileCloudMetrics) Reconcile(request reconcile.Request) (reconcile.
 	}
 
 	// for each scraped metric value we check postgresGaugeMetrics for a match and set the value and labels
-	for _, scrapedMetric := range scrapedMetrics {
-		for _, croMetric := range postgresGaugeMetrics {
-			if scrapedMetric.Name == croMetric.Name {
-				croMetric.GaugeVec.WithLabelValues(
-					scrapedMetric.Labels[labelClusterIDKey],
-					scrapedMetric.Labels[labelResourceIDKey],
-					scrapedMetric.Labels[labelNamespaceKey],
-					scrapedMetric.Labels[labelInstanceIDKey],
-					scrapedMetric.Labels[labelProductNameKey],
-					scrapedMetric.Labels[labelStrategyKey]).Set(scrapedMetric.Value)
-				r.logger.Infof("successfully set postgres metric value for %s", croMetric.Name)
-				continue
-			}
-		}
-	}
+	r.setGaugeMetrics(postgresGaugeMetrics, scrapedMetrics)
 
 	// we want full control over when we scrape metrics
 	// to allow for this we only have a single requeue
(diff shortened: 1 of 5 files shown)


pkg/providers/aws/provider_redis_metrics.go
```diff
@@ -154,6 +150,33 @@ func (r *RedisMetricsProvider) scrapeRedisCloudWatchMetricData(ctx context.Conte
 }
 
 func buildRedisMetricDataQuery(resourceID string) []*cloudwatch.MetricDataQuery {
-	//todo build this out
-	return nil
+	var metricDataQueries []*cloudwatch.MetricDataQuery
+	for _, metricType := range redisCloudWatchMetricTypes {
+		// Metrics are returned per node for ElastiCache
+		// Poll through the two nodes for ElastiCache ( would like a better way to do this as this is not robust)
```
laurafitzgerald, line 156: You could use [DescribeReplicationGroups](https://docs.aws.amazon.com/sdk-for-go/api/service/elasticache/#ElastiCache.DescribeReplicationGroups) which would give you [ReplicationGroup](https://docs.aws.amazon.com/sdk-for-go/api/service/elasticache/#DescribeReplicationGroupsOutput) from the output where you can get a node group with the two nodes which will contain the id. T'would save you building them and you could use those for the loop instead of the for loop you have there?

austincunningham: Thanks for the suggestion . The Dimension is `CacheClusterId` but that is returned by DescribeReplicationGroup https://docs.aws.amazon.com/AmazonElastiCache/latest/APIReference/API_DescribeReplicationGroups.html so better to get it directly.

```bash
 aws cloudwatch list-metrics --region=eu-west-1 | jq -c '.Metrics[]' |grep DatabaseMemoryUsagePercentage
{"Namespace":"AWS/ElastiCache","MetricName":"DatabaseMemoryUsagePercentage","Dimensions":[{"Name":"CacheClusterId","Value":"mwcolabbyoc8zbc5redhatrhmioperatort-5q6x-001"}]}
{"Namespace":"AWS/ElastiCache","MetricName":"DatabaseMemoryUsagePercentage","Dimensions":[{"Name":"CacheClusterId","Value":"mwcolabbyoc8zbc5redhatrhmioperatort-5q6x-002"}]}
```

pkg/controller/cloudmetrics/cloudmetrics_controller.go
```diff
@@ -109,7 +111,25 @@ var postgresGaugeMetrics = []CroGaugeMetric{
...
+		Name: redisMemoryUsagePercentage,
+		GaugeVec: prometheus.NewGaugeVec(
+			prometheus.GaugeOpts{
+				Name: redisMemoryUsagePercentage,
+				Help: "The percentage of redis used memory. Units: Bytes",
+			},
+			labels),
+		ProviderType: map[string]providers.CloudProviderMetricType{
+			providers.AWSDeploymentStrategy: {
+				PromethuesMetricName: redisMemoryUsagePercentage,
```
ciaranRoche, line 125: Should we also be exposing `FreeableMemory` not just `MemoryUsage`

pkg/providers/aws/provider_redis_metrics.go
```diff
@@ -153,7 +155,42 @@ func (r *RedisMetricsProvider) scrapeRedisCloudWatchMetricData(ctx context.Conte
...
+			}
+		}
+		// poll MemberCluster array for CacheClusterId
+		for _, cacheClusterId := range foundCache.MemberClusters{
+			//cast string pointer to string
+			cacheClusterIdString := "" + *cacheClusterId
+			metricDataQueries = append(metricDataQueries, &cloudwatch.MetricDataQuery{
+				// id needs to be unique, and is built from the metric name and type, and the CacheClusterId number
+				// the metric name is converted from camel case to snake case to allow it to be easily reused when exposing the metrics
+				Id: aws.String(metricType.PromethuesMetricName +"_"+ cacheClusterIdString[len(cacheClusterIdString)-3:]),
```
ciaranRoche, line 177: We want to only return a single metric ID based on `PromethuesMetricName`, I am thinking we can either look to expose the `cacheClusterID` in the label `resourceID` or we only look to monitor the primary node. I think this would be worth a discussion, as it will somewhat change/diverge from the behavior of the Postgres Provider

pkg/controller/cloudmetrics/cloudmetrics_controller.go
austincunningham: no reason why we couldn't

pkg/providers/aws/provider_redis_metrics.go
```diff
@@ -14,24 +15,19 @@ import (
...
+)
+
+//var redisCloudWatchMetricTypes = []providers.CloudProviderMetricType{
+//	{
+//		PromethuesMetricName: "cro_redis_database_memory_usage_percentage",
+//		//calculated on used_memory/maxmemory from Redis INFO http://redis.io/commands/info
+//		ProviderMetricName: "DatabaseMemoryUsagePercentage",
+//		Statistic:          "Average",
+//	},
+//}
```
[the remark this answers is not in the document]
austincunningham: :+1:

pkg/providers/aws/provider_redis_metrics.go
austincunningham: yea lets discuss it

pkg/providers/aws/provider_redis_metrics.go
```diff
@@ -81,8 +77,14 @@ func (r *RedisMetricsProvider) ScrapeRedisMetrics(ctx context.Context, redis *v1
 		return nil, errorUtil.Wrap(err, "failed to create aws session to scrape elasticache cloud watch metrics")
 	}
 
+	// Get redis instances foundCache
+	replicationGroups, err := getReplicationGroups(elasticache.New(sess))
```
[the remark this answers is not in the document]
austincunningham: polling in the buildRedisMetricDataQuery https://github.com/integr8ly/cloud-resource-operator/pull/163/files#diff-f2f7dfe227dab87b138b0fc131532c47R163-R169 , moq the replication group in the tests as couldn't figure how to moq the session .

pkg/providers/aws/provider_redis_metrics.go
```diff
@@ -102,58 +90,100 @@ func (r *RedisMetricsProvider) scrapeRedisCloudWatchMetricData(ctx context.Conte
...
+	for _, c := range replicationGroups {
+		if *c.ReplicationGroupId == resourceID {
+			foundCache = c
+			break
+		}
 	}
+	// poll MemberCluster array for CacheClusterId
+	for _, cacheClusterId := range foundCache.MemberClusters {
+		//cast string pointer to string
+		cacheClusterIdString := "" + *cacheClusterId
```
ciaranRoche, line 111: why can't we just use the `cacheClusterId` here, and just pass it directly into `buildRedisMetricDataQuery`

I think something like `aws.StringValue(cacheClusterId)` should work here

pkg/providers/aws/provider_redis_metrics.go
```diff
@@ -102,58 +90,100 @@ func (r *RedisMetricsProvider) scrapeRedisCloudWatchMetricData(ctx context.Conte
...
+		metricOutput, err := cloudWatchApi.GetMetricData(&cloudwatch.GetMetricDataInput{
+			// build metric data query array from `metricType`
+			MetricDataQueries: buildRedisMetricDataQuery(cacheClusterIdString, metricTypes),
+			// metrics gathered from start time to end time
+			StartTime: aws.Time(time.Now().Add(-resources.GetMetricReconcileTimeOrDefault(resources.MetricsWatchDuration))),
+			EndTime:   aws.Time(time.Now()),
+		})
+		if err != nil {
+			return nil, errorUtil.Wrap(err, "error getting metric for elasticache")
```
ciaranRoche, line 120: Since we are looping here through the two cache clusters, do we want to return if one of them errors. 

It might be worth logging the error and running with a `continue` instead of a return. So that we ensure we expose all metrics regardless if one of them fails.

pkg/providers/aws/provider_redis_metrics.go
```diff
@@ -102,58 +90,100 @@ func (r *RedisMetricsProvider) scrapeRedisCloudWatchMetricData(ctx context.Conte
...
-	// parse the returned data from the cloudwatch to a GenericCloudMetric
-	var metrics []*providers.GenericCloudMetric
-	for _, metricData := range metricOutput.MetricDataResults {
-		// status code complete ensures all metrics have been successful
-		if *metricData.StatusCode != cloudwatch.StatusCodeComplete {
-			continue
+		// get cluster if for use in metric labels
+		clusterID, err := resources.GetClusterID(ctx, r.Client)
+		if err != nil {
+			return nil, errorUtil.Wrap(err, "error getting clusterID")
```
ciaranRoche, line 126: It might be worth declaring this before the loop, to avoid calling it twice.

pkg/providers/aws/provider_redis_metrics.go
```diff
@@ -102,58 +90,100 @@ func (r *RedisMetricsProvider) scrapeRedisCloudWatchMetricData(ctx context.Conte
...
-					"instanceID":  resourceID,
-					"productName": redis.Labels["productName"],
-					"strategy":    redisProviderName,
-				},
-				Value: *value,
-			})
+
+		// ensure metric data results are not nil
+		if metricOutput.MetricDataResults == nil {
+			return nil, errorUtil.New("no metric data returned from elasticache cloudwatch")
```
ciaranRoche, line 131: Same as above,
```suggestion
			logger.Error("no metric data returned from elasticache cloudwatch")
			continue
```

pkg/providers/aws/provider_redis_metrics.go
```diff
@@ -102,58 +90,100 @@ func (r *RedisMetricsProvider) scrapeRedisCloudWatchMetricData(ctx context.Conte
...
+	// Metrics are returned per node for ElastiCache
+	var foundCache *elasticache.ReplicationGroup
+	for _, c := range replicationGroups {
+		if *c.ReplicationGroupId == resourceID {
+			foundCache = c
+			break
+		}
 	}
+	// poll MemberCluster array for CacheClusterId
+	for _, cacheClusterId := range foundCache.MemberClusters {
```
ciaranRoche, line 109: Would it be worth doing a nil check on founcCache, just so we can return if we don't find the replication group. We can also log something, so we can trace back if there is an issue of no exposed metrics.

pkg/providers/aws/provider_redis_metrics.go
```diff
@@ -102,58 +90,100 @@ func (r *RedisMetricsProvider) scrapeRedisCloudWatchMetricData(ctx context.Conte
...
 
-func buildRedisMetricDataQuery(resourceID string) []*cloudwatch.MetricDataQuery {
-	//todo build this out
-	return nil
+func buildRedisMetricDataQuery(cacheClusterId string, metricTypes []providers.CloudProviderMetricType) []*cloudwatch.MetricDataQuery {
+	var metricDataQueries []*cloudwatch.MetricDataQuery
+	for _, metricType := range metricTypes {
+		metricDataQueries = append(metricDataQueries, &cloudwatch.MetricDataQuery{
+			// id needs to be unique, and is built from the metric name and type, and the CacheClusterId number
+			// the metric name is converted from camel case to snake case to allow it to be easily reused when exposing the metrics
```
ciaranRoche, line 170: I think we can remove this comment, it was left over from the original approach

pkg/providers/aws/provider_redis_metrics_test.go
```diff
@@ -0,0 +1,248 @@
...
+	"strategy":    "aws-elasticache",
+}
+var testRedisMetricLabels2 = map[string]string{
+	"clusterID":   "test",
+	"instanceID":  "test-002",
+	"namespace":   "test",
+	"productName": "",
+	"resourceID":  "test",
+	"strategy":    "aws-elasticache",
+}
```
ciaranRoche, line 67: Would it be worth adding these to a single build mock labels function, so that we can easily override them inline in the test :thinking:

pkg/providers/aws/provider_redis_metrics_test.go
```diff
@@ -0,0 +1,248 @@
...
+				}),
+				elastiCacheApi: moq_aws.BuildMockElastiCacheClient(func(watchClient *moq_aws.MockElastiCacheClient){
+					watchClient.DescribeReplicationGroupsFn = func(input *elasticache.DescribeReplicationGroupsInput)(*elasticache.DescribeReplicationGroupsOutput, error) {
+						return &elasticache.DescribeReplicationGroupsOutput{
+							ReplicationGroups: buildReplicationGroupReadyCacheClusterId(),
+						}, nil
+					}
+				}),
+				redis:             buildTestRedisCR(),
+				//replicationGroups: buildReplicationGroupReadyCacheClusterId(),
```
ciaranRoche, line 222: I think we can remove this `comment`

pkg/controller/cloudmetrics/cloudmetrics_controller.go
```diff
@@ -218,6 +281,22 @@ func (r *ReconcileCloudMetrics) Reconcile(request reconcile.Request) (reconcile.
...
+					scrapedMetric.Labels[labelResourceIDKey],
+					scrapedMetric.Labels[labelNamespaceKey],
+					scrapedMetric.Labels[labelInstanceIDKey],
+					scrapedMetric.Labels[labelProductNameKey],
+					scrapedMetric.Labels[labelStrategyKey]).Set(scrapedMetric.Value)
+				r.logger.Infof("successfully set redis metric value for %s", croMetric.Name)
+				continue
+			}
+		}
+	}
```
ciaranRoche, line 299: Would it be worth adding a helper function here, as this logic is repeated for `postgresGaugeMetrics`

pkg/providers/aws/provider_redis_metrics_test.go
ciaranRoche, line 67: Would it be worth, making these into a build mock labels func, so we can pass and override the value in line in the test cases

pkg/providers/aws/provider_redis_metrics_test.go
```diff
@@ -0,0 +1,248 @@
...
```
ciaranRoche, line 180: I think we can remove this comment

pkg/controller/cloudmetrics/cloudmetrics_controller.go
austincunningham: it might

pkg/providers/aws/provider_redis_metrics.go
austincunningham: good point

pkg/providers/aws/provider_redis_metrics.go
austincunningham: :+1:

pkg/providers/aws/provider_redis_metrics.go
austincunningham: :+1:

pkg/providers/aws/provider_redis_metrics.go
austincunningham: agree

pkg/providers/aws/provider_redis_metrics_test.go
austincunningham: agree

pkg/providers/aws/provider_redis_metrics_test.go
austincunningham: agree

pkg/providers/aws/provider_redis_metrics.go
austincunningham: sure why not

pkg/providers/aws/provider_redis_metrics.go
austincunningham: we can , I just have a dislike for pointers

pkg/providers/aws/provider_redis_metrics_test.go
austincunningham: like this idea, will go with it.

pkg/controller/cloudmetrics/cloudmetrics_controller.go
```diff
@@ -293,3 +344,22 @@ func registerGaugeVectorMetrics(logger *logrus.Entry) {
 		customMetrics.Registry.MustRegister(metric.GaugeVec)
 	}
 }
+
+// func labelScrapedMetrics add labels to scraped metrics
+func (r *ReconcileCloudMetrics) labelScrapeMetrics( gaugeMetrics []CroGaugeMetric, scrapedMetrics []*providers.GenericCloudMetric) {
```
ciaranRoche, line 349: small nit pick on this, could we rename this to something like `setGaugeMetrics` as this is what we are doing here, setting a value on an exposed metric with labels.

pkg/providers/aws/provider_redis_metrics_test.go
```diff
@@ -0,0 +1,255 @@
...
+	"strategy":    "aws-elasticache",
+}
+var testRedisMetricLabels2 = map[string]string{
+	"clusterID":   "test",
+	"instanceID":  "test-002",
+	"namespace":   "test",
+	"productName": "",
+	"resourceID":  "test",
+	"strategy":    "aws-elasticache",
+}
```
ciaranRoche, line 68: I think we can remove these two var.

pkg/controller/cloudmetrics/cloudmetrics_controller.go
austincunningham: sure

pkg/providers/aws/provider_redis_metrics_test.go
austincunningham: yep

ciaranRoche:
/override ci/prow/e2e

last change was an update to the func name, no need to run e2e again

PR #167, opened 2020-07-23
remove ability to update allocated storage via strategy map

ciaranRoche:
add ability to update max allocated storage via strategy map

Currently, aws will autoscale as needed from allocated storage up to max allocated storage value. 
We need to remove the ability to allow cro to update the allocated storage value, as we want to rely 
on aws auto scaling. 
We do wish to allow cro the ability to update max allocated storage value

## Overview

Jira: https://issues.redhat.com/browse/INTLY-7231

## Verification

- Clone this branch
- Run `make cluster/prepare`
- Run `make cluster/seed/managed/postgres
- Run `make run`
- Update `"AllocatedStorage":50` in the create strategy for postgres in the cloud-resources-aws-strategies config map
- Update `"MaxAllocatedStorage":500` in the create strategy for postgres in the cloud-resources-aws-strategies config map
- Ensure allocated storage is not updated (ensure it is not added to pending modifications) 
- Ensure MaxAllocatedStorage is updated to 500 GiB


pkg/providers/aws/provider_postgres.go
```diff
@@ -696,11 +696,6 @@ func verifyPendingModification(mi *rds.ModifyDBInstanceInput, pm *rds.PendingMod
 			pendingModifications = false
 		}
 	}
-	if mi.AllocatedStorage != nil && pm.AllocatedStorage != nil {
-		if *mi.AllocatedStorage == *pm.AllocatedStorage {
-			pendingModifications = false
-		}
-	}
```
aidenkeating, line 703: Should we be replacing this with MaxAllocatedStorage also?

ciaranRoche: No, this is to check for pending modifications, such that we do not continue calling modify instance until the next maintenance window hits. Id the modification is in a pending state. 

max allocation storage is modified instantly

aidenkeating: Ok, I see that documented here https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Overview.DBInstance.Modifying.html under `Settings for your DB Instances`

aidenkeating:
/approve

Verified against MW Collab

PR #168, opened 2020-07-27
add exposing max and allocated storage metrics for postgres

ciaranRoche:
## Overview
cloud watch only provides us with free storage space, we need to expose more metrics to allow for more accurate alerting
as `predict_linear` is hard to predict on non-linear growth & results in false positives
we should follow the approach AWS take to auto scaling, and alert when free storage space is less than 10%

Co-Authored-By: laurafitzgerald <lfitzger@redhat.com>

## Verification

- Clone this branch
- Run `make cluster/prepare`
- Run `make run`
- Run `curl localhost:8383/metrics`
- Ensure the following metrics are exposed
image

### DIFF
--- a/pkg/providers/aws/provider_postgres.go
+++ b/pkg/providers/aws/provider_postgres.go
@@ -974,6 +974,14 @@ func (p *PostgresProvider) exposePostgresMetrics(ctx context.Context, cr *v1alph
 	} else {
 		resources.SetMetric(resources.DefaultPostgresAvailMetricName, genericLabels, 1)
 	}
+
+	// cloud watch only provides us with free storage space, we need to expose more metrics to allow for more accurate alerting
+	// as `predict_linear` is hard to predict on non-linear growth & results in false positives
+	// we should follow the approach AWS take to auto scaling, and alert when free storage space is less than 10%
+	if instance != nil && instance.AllocatedStorage != nil {
+		// convert allocated storage to bytes and expose as a metric
+		resources.SetMetric(resources.DefaultPostgresAllocatedStorageMetricName, genericLabels, float64(*instance.AllocatedStorage*resources.BytesInGibiBytes))
+	}
 }
 
 // set metrics about the postgres instance being deleted
--- a/pkg/resources/custom_metrics.go
+++ b/pkg/resources/custom_metrics.go
@@ -18,22 +18,25 @@ import (
 )
 
 const (
-	sleepytime                              = 3600
-	DefaultPostgresMaintenanceMetricName    = "cro_postgres_service_maintenance"
-	DefaultPostgresInfoMetricName           = "cro_postgres_info"
-	DefaultPostgresAvailMetricName          = "cro_postgres_available"
-	DefaultPostgresConnectionMetricName     = "cro_postgres_connection"
-	DefaultPostgresStatusMetricName         = "cro_postgres_status_phase"
-	DefaultPostgresDeletionMetricName       = "cro_postgres_deletion_timestamp"
-	DefaultPostgresSnapshotStatusMetricName = "cro_postgres_snapshot_status_phase"
-	DefaultRedisMaintenanceMetricName       = "cro_redis_service_maintenance"
-	DefaultRedisInfoMetricName              = "cro_redis_info"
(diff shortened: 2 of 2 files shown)


pkg/providers/aws/provider_postgres.go
```diff
@@ -974,6 +974,16 @@ func (p *PostgresProvider) exposePostgresMetrics(ctx context.Context, cr *v1alph
 	} else {
 		resources.SetMetric(resources.DefaultPostgresAvailMetricName, genericLabels, 1)
 	}
+
+	// cloud watch only provides us with free storage space, we need to expose more metrics to allow for more accurate alerting
+	// as `predict_linear` is hard to predict on non-linear growth & results in false positives
+	// we should follow the approach AWS take to auto scaling, and alert when free storage space is less than 10%
+	if instance != nil && instance.MaxAllocatedStorage != nil && instance.AllocatedStorage != nil{
```
aidenkeating, line 981: Might need a `go fmt` here? I'm surprised that isn't in our CI

pkg/providers/aws/provider_postgres.go
```diff
@@ -974,6 +974,16 @@ func (p *PostgresProvider) exposePostgresMetrics(ctx context.Context, cr *v1alph
...
+		// convert max allocated storage to bytes and expose as a metric
+		resources.SetMetric(resources.DefaultPostgresMaxAllocatedStorageMetricName, genericLabels, float64(*instance.MaxAllocatedStorage)*1e+9)
```
aidenkeating, line 983: Just to double check `MaxAllocatedStorage` is in GiB, right?  Not GB. Making sure this calculation is accurate

pkg/resources/custom_metrics.go
```diff
@@ -18,22 +18,24 @@ import (
...
+	sleepytime                                   = 3600
+	DefaultPostgresMaintenanceMetricName         = "cro_postgres_service_maintenance"
+	DefaultPostgresInfoMetricName                = "cro_postgres_info"
+	DefaultPostgresAvailMetricName               = "cro_postgres_available"
+	DefaultPostgresConnectionMetricName          = "cro_postgres_connection"
+	DefaultPostgresStatusMetricName              = "cro_postgres_status_phase"
+	DefaultPostgresDeletionMetricName            = "cro_postgres_deletion_timestamp"
+	DefaultPostgresSnapshotStatusMetricName      = "cro_postgres_snapshot_status_phase"
+	DefaultPostgresMaxAllocatedStorageMetricName = "cro_postgres_max_allocated_storage"
+	DefaultPostgresAllocatedStorageMetricName    = "cro_postgres_allocated_storage"
```
aidenkeating, line 30: This probably isn't going to scale well across clouds, but if it's something we need, let's go with it and create a ticket in the GCP Epic to be aware of these two metrics and how to satisfy them in GCP land

pkg/providers/aws/provider_postgres.go
laurafitzgerald: @aidenkeating we noticed last week it's not working as expected. There's a ticket to fix it: https://issues.redhat.com/browse/INTLY-9497

aidenkeating: Oof 👍

pkg/resources/custom_metrics.go
ciaranRoche: good catch, will remove to have a single metric for `current_allocated_storage` that is what we will be alerting off of anyway :+1:

pkg/providers/aws/provider_postgres.go
ciaranRoche: Yeah just changing it to GiB after confirming Gibibytes 

https://docs.aws.amazon.com/sdk-for-go/api/service/rds/#DBInstance

aidenkeating: Is this correct then? Gibibyte is `1024^3` bytes, right? Looks like we're doing `1000^3`

austincunningham:
Seeing **cro_postgres_current_allocated_storage** rather that **cro_postgres_allocated_storage** 
image

presume the verification steps are out of date, if so this is verified

PR #171, opened 2020-08-06
aws supported types script 

ciaranRoche:
## Overview

Jira: https://issues.redhat.com/browse/INTLY-9737

### Verification
- run `./supported_types.sh t3.micro `

### DIFF
--- a/hack/aws/README.md
+++ b/hack/aws/README.md
@@ -1,9 +1,17 @@
 # AWS Helpers
 
 ### supported_types
-Used to check every Availability Zone to see that specific instance types are supported.
+Used to check every Availability Zone to see that specific instance types are supported. To match Availability Zones from the scripts result, to its region please follow [AWS documentation](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-available-regions).
 
 We should check for both current instance types used by CRO and their legacy equivalent. 
+
+#### Prerequisites
+
+- [aws-cli](https://aws.amazon.com/cli/) 
+
+Ensure you have installed and configured the `aws-cli` to the account you wish to check.
+
+_NOTE: This script only has the ability to check regions which the account has opted into_
 #### Usage
 ``` 
 ./supported_types.sh <<instance type to check>>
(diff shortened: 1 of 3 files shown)


hack/aws/supported_types.sh
```diff
@@ -0,0 +1,72 @@
...
+
+  for offering in $(jq -r '.[]["Location"]' <<<"$INSTANCE_OFFERINGS")
+  do
+    OFFERING_ZONE_NAMES=("${OFFERING_ZONE_NAMES[@]}" "$offering")
+  done
+}
+
+# check for regions which do not support offering
+check_supported_offerings() {
+for zone in $ZONE_NAMES
```
aidenkeating:
```suggestion
for zone in "${ZONE_NAMES[@]}"
```

required for macOS.

hack/aws/supported_types.sh
```diff
@@ -0,0 +1,72 @@
...
+}
+
+if [ -z "${INSTANCE_TYPE}" ]; then
+  echo "required to pass instance type, eg t3.micro"
+  exit 1
+fi
+
+echo "checking supported offerings for $INSTANCE_TYPE"
+get_regions
+for region in $REGION_NAMES
```
aidenkeating, line 72:
```suggestion
for region in "${REGION_NAMES}"
```

required for macOS.

hack/aws/README.md
```diff
@@ -0,0 +1,10 @@
+# AWS Helpers
+
+### supported_types
+Used to check every Availability Zone to see that specific instance types are supported.
+
+We should check for both current instance types used by CRO and their legacy equivalent.
```
aidenkeating, line 6: The user needs to ensure they have the `aws` CLI setup and configured against the account they want to check.

Maybe we should also include a link to [1] so they can understand what that availability zone is related to e.g. eu-north-1 means Stockholm. 

Also, might be good to note that this will only include opt-in regions like South Africa if the AWS account has already opted into them.

[1] https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html#concepts-available-regions

PR #173, opened 2020-08-10
ensure duplicate azs are never used

aidenkeating:
currently it is possible that duplicate azs in a region are used
to attempt to create a subnet group, under two conditions.

- aws returns a duplicate set of subnet groups
- a legacy subnet exists

ensure that duplicate azs are never used in subnet creation by
filtering out any duplicates.

verification:
- run 'make run'
- create a redis and postgres instance
- ensure both instances are created successfully in aws


laurafitzgerald:
Verification:
Against `mw-collab-byoc`. I left things in place in case anyone wants to take a look but redis and postgres installed successfully.
Screenshot 2020-08-10 at 17 45 17
Screenshot 2020-08-10 at 17 51 22

The subnet groups can also be viewed in aws.

PR #180, opened 2020-08-19
do not specify cache instances when listing availabile instance types

aidenkeating:
verification:
- install a redis aws resource in stockholm
- override the strategy configmap to override it to cache.t2.micro
- ensure it's not overridden and the redis cr is in a successful
  state


aidenkeating:
@ciaranRoche @austincunningham Would you mind taking a look? I think this would be a good thing to get in for RHMI 2.7.0. It was never picked back to master

austincunningham:
Changes look good  but how will it effect a scale up on the strategy configmap ?

ciaranRoche (approved):
Ability to scale instance
image
Ignoring unsupported types
image

PR #182, opened 2020-08-24
remove hardcoded regions from the aws strategy configmap example

aidenkeating:
currently the example aws strategy configmap is created in dev make
targets. this means that the hardcoded region of eu-west-1 is used
even if the cluster isn't in eu-west-1 which can cause issues with
vpc setup.

remove the hardcoded region but keep the explicit keys to let users
know the possible keys that can be used in the strategy configmap.

verification:
- in cluster in a non-eu-west-1 region
- run make cluster/prepare NAMESPACE=<your-namespace>
- ensure the region isn't specified in the aws strategy configmap
- create a postgres, redis and blobstorage and ensure each are
  created in the region of the cluster

## additional notes

it appears that custom resources (at least postgres and redis) for
aws can no longer specify a specific region. we either need to
consider this a bug or work to remove the region key entirely from
the strategy configmap.

### DIFF
--- a/deploy/examples/cloud_resources_aws_strategies.yaml
+++ b/deploy/examples/cloud_resources_aws_strategies.yaml
@@ -4,11 +4,11 @@ metadata:
   name: cloud-resources-aws-strategies
 data:
   blobstorage: |
-    {"development": { "region": "eu-west-1", "createStrategy": {}, "deleteStrategy": {} }}
+    {"development": { "region": "", "createStrategy": {}, "deleteStrategy": {} }}
   redis: |
-    {"development": { "region": "eu-west-1", "createStrategy": {}, "deleteStrategy": {} }}
+    {"development": { "region": "", "createStrategy": {}, "deleteStrategy": {} }}
   postgres: |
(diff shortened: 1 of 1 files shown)


aidenkeating:
@ciaranRoche @darahayes I'd be interested to get your points of view on this. I'd be tempted to consider this a bug. 

Also, happy to fully remove the key from the example ConfigMap if needed, I find it just makes things more clear about what's possible through the ConfigMap, even though it is broken at the moment.

ciaranRoche:
@aidenkeating this is an interesting one, and definitely a bug. I would vote to leave the key, as I think this is a nice feature. We will need to revisit the networking, and the workflow of how we provision the CRO VPC and the resources. It might be a nice piece of work to bring into [INTLY-6581](https://issues.redhat.com/browse/INTLY-6581)
I would like to sync up with you on this and an approach we could take to solve it.

ciaranRoche (approved):
/approve 

Adding an approve as I think this PR solves the issue intended.

ciaranRoche:
Verified 
image

PR #184, opened 2020-08-26
fix: don't reconcile redis/postgres versions if current version is higher than specified version

darahayes:
## Overview

Jira: https://issues.redhat.com/browse/INTLY-9686

This fixes a bug where if a Postgres or Redis instance is upgraded to a higher version number from the AWS console, then CRO would try to modify the instance back to the version specified in the aws strategy config. When the instance version is higher than what is specified in the config, then CRO tries to 'upgrade' to a lower version number which returns an error from the AWS api and puts the CR into a bad state.

The solution was either to

1. allow the request to be made to AWS and catch that specific error and do nothing
2. check the versions and only try to apply the update where appropriate

Although the 1st option would be easier I went for the second option to rule out a scenario where we're constantly making an unnecessary request to AWS. The solution involves bringing in a new dependency for parsing the version numbers. https://github.com/hashicorp/go-version is a small and well maintained library.

## Verification

- Clone this branch
- Run `make cluster/prepare`
- Run `make run`

* Create a Postgres CR
* Log into AWS and modify the postgres instance to have a newer engine version number. Make sure to click the 'Apply Immediately' checkbox
* Wait for the instance to be upgraded and become available
* check the status of the postgres CR and verify that it's available and there are no errors. (previously there would have been an error from CRO trying to 'upgrade' to an older version number which is not allowed by AWS)
* Repeat the above steps for a Redis CR

## Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,6 +13,7 @@ require (
 	github.com/go-openapi/spec v0.19.3
 	github.com/google/uuid v1.1.1
 	github.com/gregjones/httpcache v0.0.0-20190203031600-7a902570cb17 // indirect
+	github.com/hashicorp/go-version v1.2.1
 	github.com/lib/pq v1.3.0
 	github.com/mattn/go-colorable v0.1.7 // indirect
 	github.com/mattn/go-runewidth v0.0.9 // indirect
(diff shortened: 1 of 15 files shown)


psturc:
👀

psturc:
/lgtm
verified on my cluster

PR #186, opened 2020-09-01
INTLY-9552:  add a check for valid status to ModifyDBInstance

austincunningham:
## Overview
This is to cover a edge case where the configmap `cloud_resource_aws_strategies.yaml` createstrategy is updated but the rds instance status is something other that **available** from the time the configmap is updated to the end of the maintenance window

Jira: https://issues.redhat.com/browse/INTLY-9552

## Verification

- Clone this branch
- Run `make cluster/prepare`
- Run `make run`
- wait for postgres cr status to be 
```yaml
status:
  message: 'successfully created and tagged, aws rds status is available'
```
- Set the status to `Backing-up` by taking a [snapshot](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_CreateSnapshot.html)
- update the configmap `cloud_resource_aws_strategies.yaml` in the cro operator namespace and change the `postgres.development.createstrategy{}`
e.g.
```json
"createStrategy": {
      "DBInstanceClass": "db.t3.micro",
    }
```
- confirm that the pending modification update 
```bash
aws rds describe-db-instances --db-instance-identifier=<DBInstanceIdentifier> | jq -r '.DBInstances[0].PendingModifiedValues'
# e.g. output
{
  "DBInstanceClass": "db.t3.micro"
}
```

## Checklist

### DIFF
--- a/pkg/providers/aws/provider_postgres.go
+++ b/pkg/providers/aws/provider_postgres.go
@@ -664,7 +664,11 @@ func buildRDSUpdateStrategy(rdsConfig *rds.CreateDBInstanceInput, foundConfig *r
 		mi.DeletionProtection = rdsConfig.DeletionProtection
 		updateFound = true
 	}
-	if *rdsConfig.Port != *foundConfig.Endpoint.Port {
+	config := *foundConfig
+	endpoint := config.Endpoint
+	if endpoint == nil {
+		// nil pointer check as *foundConfig.Endpoint doesn't exist on first run
+	} else if *rdsConfig.Port != *foundConfig.Endpoint.Port {
 		mi.DBPortNumber = rdsConfig.Port
 		updateFound = true
 	}
--- a/pkg/providers/aws/provider_postgres.go
+++ b/pkg/providers/aws/provider_postgres.go
@@ -300,12 +300,6 @@ func (p *PostgresProvider) createRDSInstance(ctx context.Context, cr *v1alpha1.P
 		return nil, "started rds provision", nil
 	}
 
-	// check rds instance phase
-	if *foundInstance.DBInstanceStatus != "available" {
-		logger.Infof("found instance %s current status %s", *foundInstance.DBInstanceIdentifier, *foundInstance.DBInstanceStatus)
-		return nil, croType.StatusMessage(fmt.Sprintf("createRDSInstance() in progress, current aws rds resource status is %s", *foundInstance.DBInstanceStatus)), nil
-	}
-
 	// check if found instance and user strategy differs, and modify instance
 	logger.Infof("found existing rds instance: %s", *foundInstance.DBInstanceIdentifier)
 	mi, err := buildRDSUpdateStrategy(rdsCfg, foundInstance)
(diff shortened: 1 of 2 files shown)


pkg/providers/aws/provider_postgres.go
```diff
@@ -657,7 +664,11 @@ func buildRDSUpdateStrategy(rdsConfig *rds.CreateDBInstanceInput, foundConfig *r
 		mi.DeletionProtection = rdsConfig.DeletionProtection
 		updateFound = true
 	}
-	if *rdsConfig.Port != *foundConfig.Endpoint.Port {
+	config := *foundConfig
+	endpoint := config.Endpoint
+	if endpoint == nil {
+		// nil pointer check as *foundConfig.Endpoint doesn't exist on first run
+	} else if *rdsConfig.Port != *foundConfig.Endpoint.Port {
```
ciaranRoche, line 671: I think we can shorten this a bit.
```suggestion
	if *foundConfig.Endpoint != nil && *rdsConfig.Port != *foundConfig.Endpoint.Port {
```

pkg/providers/aws/provider_postgres.go
```diff
@@ -319,12 +313,25 @@ func (p *PostgresProvider) createRDSInstance(ctx context.Context, cr *v1alpha1.P
 	}
 	if mi != nil {
 		if _, err = rdsSvc.ModifyDBInstance(mi); err != nil {
+			// AWS handles invalid state with the ErrCodeInvalidDBInstanceStateFault
+			if aerr, ok := err.(awserr.Error); ok {
+				if aerr.Code() == rds.ErrCodeInvalidDBInstanceStateFault {
```
ciaranRoche, line 318: Just one comment on this, if there is an `ErrCodeInvalidDBInstanceStateFault` we would end up instantly reconciling, hitting this error. Each time calling `ModifyDBInstance`, building up excess write API calls. 

Going by the ticket we have a list of `safe` states, can we check on those before calling `ModifyDBInstance`, similarly to how we expose metrics with `rdsInstanceStatusIsHealthy`

ciaranRoche (commented):
Looks good, just a comment and suggestion inline.

pkg/providers/aws/provider_postgres.go
austincunningham: Unfortunately we don't have a list of safe states. We have a list of all states.

pkg/providers/aws/provider_postgres.go
austincunningham: compiler doesn't like it type issue `Cannot convert 'nil' to type 'Endpoint'` , may be able to shorten a little.

pkg/providers/aws/provider_postgres.go
austincunningham: Checking what the InvalidDBInstanceStateFault checks for I found https://github.com/aws/aws-sdk-go/blob/5d8a3d6c71e822edf1e5b7037989283ddd7b1c8e/models/apis/docdb/2014-10-31/docs-2.json#L823-L824 which is the same as our original check. closing this pr

PR #190, opened 2020-10-06
fixes a change in the aws rds payload

jjaferson:
## Overview

It looks like there was a change in RDS and the `AutoAppliedAfterDate` is not always present in the payload from the `describe-pending-maintenance-actions` request anymore, so if an instance has a pending maintenance, CRO operator will panic trying to get the timestamp in a variable that wasn't returned.  

*Example of the payload sent back from the AWS CLI*
```
{
            "ResourceIdentifier": "instance-id",
            "PendingMaintenanceActionDetails": [
                {
                    "Action": "db-upgrade",
                    "OptInStatus": "next-maintenance",
                    "CurrentApplyDate": "2020-10-08T02:00:00Z",
                    "Description": "Automatic minor version upgrade to postgres 10.13"
                }
            ]
        },
```

Jira: https://issues.redhat.com/browse/INTLY-10206

## Verification

- Install the operator against a cluster that has an RDS instance and a pending maintenance available

### DIFF
--- a/pkg/providers/aws/provider_postgres.go
+++ b/pkg/providers/aws/provider_postgres.go
@@ -1089,11 +1089,19 @@ func (p *PostgresProvider) setPostgresServiceMaintenanceMetric(ctx context.Conte
 
 		for _, pma := range su.PendingMaintenanceActionDetails {
 
-			metricLabels["AutoAppliedAfterDate"] = strconv.FormatInt((*pma.AutoAppliedAfterDate).Unix(), 10)
-			metricLabels["CurrentApplyDate"] = strconv.FormatInt((*pma.CurrentApplyDate).Unix(), 10)
-			metricLabels["Description"] = *pma.Description
+			metricEpochTimestamp := time.Now().Unix()
 
-			metricEpochTimestamp := (*pma.AutoAppliedAfterDate).Unix()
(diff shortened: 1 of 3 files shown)


aidenkeating (approved):
Looks good, does this need an additional verification from me? Or has that already been covered?

Also a release will need to be cut, there's documentation in the README

pkg/providers/aws/provider_postgres.go
```diff
@@ -1089,11 +1089,19 @@ func (p *PostgresProvider) setPostgresServiceMaintenanceMetric(ctx context.Conte
 
 		for _, pma := range su.PendingMaintenanceActionDetails {
 
-			metricLabels["AutoAppliedAfterDate"] = strconv.FormatInt((*pma.AutoAppliedAfterDate).Unix(), 10)
-			metricLabels["CurrentApplyDate"] = strconv.FormatInt((*pma.CurrentApplyDate).Unix(), 10)
-			metricLabels["Description"] = *pma.Description
+			metricEpochTimestamp := time.Now().Unix()
+
+			if !pma.AutoAppliedAfterDate.IsZero() {
```
aidenkeating, line 1094: This `pma.AutoAppliedAfterDate` guarenteed to not be `nil`?

PR #201, opened 2020-12-15
adding logic to provide a default non overlapping cidr block for the standanlone network

laurafitzgerald:
adding logic to provide a default non overlapping cidr block for the standanlone network

## Overview

Jira: https://issues.redhat.com/browse/MGDAPI-582

## Verification

- Clone this branch
- Run `make cluster/prepare`
- Run `make run`

### Verify that the provided cidr in the _network block is used.

Check the config map that was created when preparing the cluster

`oc get configmaps cloud-resources-aws-strategies -n cloud-resource-operator -o json | jq '.data'`

```
{
  "_network": "{\"development\": { \"region\": \"\", \"createStrategy\": { \"CidrBlock\": \"10.1.0.0/16\" }, \"deleteStrategy\": {} }}\n",
  ........
}
```

Create a postgres cr
Run `make cluster/seed/managed/postgres`

Verify that the standalone vpc is created using the default

`aws ec2 describe-vpcs --filter "Name=tag-value,Values=RHMI Cloud Resource VPC"`

Expected Output (with extra values stripped out)
```
{
    "Vpcs": [
        {
            "CidrBlock": "10.1.0.0/16",
........
            "CidrBlockAssociationSet": [
                {
                    "CidrBlock": "10.1.0.0/16",
          ........
                {
                    "Key": "Name",
                    "Value": "RHMI Cloud Resource VPC"
                }
            ]
        }
    ]
}
```

### Verify that the default cidr block is created

Delete the postgres cr
`oc delete postgres example-postgres -n cloud-resource-operator`
Verify that the vpc is deleted

`aws ec2 describe-vpcs --filter "Name=tag-value,Values=RHMI Cloud Resource VPC"`

Expected Output
```
{
    "Vpcs": []
}

```
Patch the config map to have no default

```
oc patch configMap cloud-resources-aws-strategies -n cloud-resource-operator --type='json' -p '[{"op": "add", "path": "/data/_network", "value":"{ \"development\": { \"createStrategy\": { \"CidrBlock\": \"''\" } } }"}]'
```

Create a postgres cr
Run `make cluster/seed/managed/postgres`

Verify that the default cidr block has been generated and the vpc is created.

`aws ec2 describe-vpcs --filter "Name=tag-value,Values=RHMI Cloud Resource VPC"`

The expected cidr block will depend on the cluster cidr block. The options begin at 10.0.0.0/26. If you cluster machine cidr is _not_ that range then the vpc cidr would be 10.0.0.0/26. If not it will be 10.1.0.0/26.

Check the 

```{
    "Vpcs": [
        {
            "CidrBlock": "10.0.0.0/26",
     ......
            "CidrBlockAssociationSet": [
                {
                   ....
                    "CidrBlock": "10.0.0.0/26",
                    "CidrBlockState": {
                        "State": "associated"
                    }
                }
            ],
.....
            "Tags": [
                {
                    "Key": "Name",
                    "Value": "RHMI Cloud Resource VPC"
                },
      .......
}
```

## Checklist

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,6 @@ require (
 	github.com/elazarl/goproxy/ext v0.0.0-20190421051319-9d40249d3c2f // indirect
 	github.com/ghodss/yaml v1.0.1-0.20180820084758-c7ce16629ff4 // indirect
 	github.com/go-openapi/spec v0.19.3
-	github.com/gomodule/redigo v2.0.0+incompatible // indirect
 	github.com/google/uuid v1.1.1
 	github.com/gregjones/httpcache v0.0.0-20190203031600-7a902570cb17 // indirect
 	github.com/hashicorp/go-version v1.2.1
--- a/pkg/providers/aws/cluster_network_provider.go
+++ b/pkg/providers/aws/cluster_network_provider.go
@@ -63,6 +63,8 @@ const (
 	// see https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeVpcPeeringConnections.html
 	filterVpcPeeringRequesterId = "requester-vpc-info.vpc-id"
 	filterVpcPeeringAccepterId  = "accepter-vpc-info.vpc-id"
+	// default cidr values
+	defaultCIDRMask = "26"
 )
 
 // wrapper for ec2 vpcs, to allow for extensibility
@@ -153,12 +155,21 @@ func (n *NetworkProvider) CreateNetwork(ctx context.Context, vpcCidrBlock *net.I
 		//By default, `integreatly.org/clusterID`.
 		//
 		//NOTE - Once a VPC is created we do not want to update it. To avoid changing cidr block
+		clusterVPC, err := getClusterVpc(ctx, n.Client, n.Ec2Api, n.Logger)
+		if err != nil {
+			return nil, errorUtil.Wrap(err, "failed to get cluster vpc")
+		}
+
+		_, clusterVPCCidr, err := net.ParseCIDR(*clusterVPC.CidrBlock)
+		if err != nil {
+			return nil, errorUtil.Wrap(err, "failed to parse cluster vpc cidr block")
+		}
 
 		// standalone vpc cidr block can not overlap with existing cluster vpc cidr block
 		// issue arises when trying to peer both vpcs with invalid vpc error - `overlapping CIDR range`
 		// we need to ensure both cidr ranges do not intersect before creating the standalone vpc
 		// as the creation of a standalone vpc is designed to be a one shot pass and not to be updated after the fact
-		if err := validateStandaloneCidrBlock(ctx, n.Client, n.Ec2Api, logger, vpcCidrBlock); err != nil {
+		if err := validateStandaloneCidrBlock(vpcCidrBlock, clusterVPCCidr); err != nil {
 			return nil, errorUtil.Wrap(err, "vpc validation failure")
 		}
 		logger.Infof("cidr %s is valid 👍", vpcCidrBlock.String())
@@ -1491,13 +1502,15 @@ func getElasticacheSubnetByGroup(elasticacheApi elasticacheiface.ElastiCacheAPI,
 	return nil, nil
 }
 
-// getNetworkProviderConfig return parsed ipNet cidr block
+// ReconcileNetworkProviderConfig return parsed ipNet cidr block
 // a _network resource type strategy, is expected to have the same tier as either postgres or redis resource type
 // ie. for a postgres tier X there should be a corresponding _network tier X
 //
 // the _network strategy config is unmarshalled into a ec2 create vpc input struct
 // from the struct the cidr block is parsed to ensure validity
-func getNetworkProviderConfig(ctx context.Context, configManager ConfigManager, tier string, logger *logrus.Entry) (*net.IPNet, error) {
+// if there is no entry for cidrblock in the _network block a sensible default which doesn't overlap with the cluster vpc
+// if cro is unable to find a valid non-overlapping cidr block it will return an error
+func (n *NetworkProvider) ReconcileNetworkProviderConfig(ctx context.Context, configManager ConfigManager, tier string, logger *logrus.Entry) (*net.IPNet, error) {
 	logger.Infof("fetching _network strategy config for tier %s", tier)
 
 	stratCfg, err := configManager.ReadStorageStrategy(ctx, providers.NetworkResourceType, tier)
@@ -1510,18 +1523,88 @@ func getNetworkProviderConfig(ctx context.Context, configManager ConfigManager, 
 		return nil, errorUtil.Wrap(err, "failed to unmarshal aws vpc create config")
 	}
 
-	if vpcCreateConfig.CidrBlock == nil {
-		return nil, errorUtil.New(fmt.Sprintf("CidrBlock in _network strategy tier %s can not be nil", tier))
+	// if the config map is found and the _network block contains an entry, that is returned for use in the network creation
+	if vpcCreateConfig.CidrBlock != nil && *vpcCreateConfig.CidrBlock != "" {
+		_, vpcCidr, err := net.ParseCIDR(*vpcCreateConfig.CidrBlock)
+		if err != nil {
+			return nil, errorUtil.Wrap(err, "failed to parse cidr block from _network strategy")
+		}
+		logger.Infof("found vpc cidr block %s in network strategy tier %s", vpcCidr.String(), tier)
+		return vpcCidr, nil
 	}
 
-	_, vpcCidr, err := net.ParseCIDR(*vpcCreateConfig.CidrBlock)
+	//if vpcCreateConfig.CidrBlock is nil or an empty string we can go ahead and set a sensible non overlapping default with a mask size of /26
+	defaultCIDR, err := n.getNonOverlappingDefaultCIDR(ctx)
 	if err != nil {
-		return nil, errorUtil.Wrap(err, "failed to parse cidr block from _network strategy")
+		return nil, errorUtil.Wrap(err, "failed to generate default CIDR")
 	}
-	cidrMask, _ := vpcCidr.Mask.Size()
 
-	logger.Infof("found vpc cidr block %s/%d in network strategy tier %s", vpcCidr.IP.String(), cidrMask, tier)
-	return vpcCidr, nil
+	// a default cidr is generated and updated to the config map, that is returned for use in the network creation
+	return defaultCIDR, nil
+}
+
+// getNonOverlappingDefaultCIDR returns a non overlapping cidr block based on the OpenShift vpc cidr block
+// the default mask is /26
+// for other masks the user is required to provide their own via config
+func (n *NetworkProvider) getNonOverlappingDefaultCIDR(ctx context.Context) (*net.IPNet, error) {
+	clusterVpc, err := getClusterVpc(ctx, n.Client, n.Ec2Api, n.Logger)
+	if err != nil {
+		return nil, errorUtil.Wrap(err, "failed to get cluster vpc for cidr block")
+	}
+	//parse the cidr to an ipnet cidr in order to manipulate the ip
+	_, clusterNet, err := net.ParseCIDR(aws.StringValue(clusterVpc.CidrBlock))
+	if err != nil {
+		return nil, errorUtil.Wrap(err, "error parsing cluster cidr block")
+	}
+
+	// this map is used to loop through the available options for a vpc cidr range in aws
+	// See aws docs https://docs.aws.amazon.com/vpc/latest/userguide/VPC_Subnets.html#vpc-sizing-ipv4
+	cidrRanges := map[string]string{
+		"10.255.255.255/8":  fmt.Sprintf("10.0.0.0/%s", defaultCIDRMask),
+		"172.31.255.255/12": fmt.Sprintf("172.16.0.0/%s", defaultCIDRMask),
+	}
+
+	// in this loop we loop through the available cidr ranges for vpcs in aws
+	// in each range the ip of the cidr block is incremented only in the A and B range
+	// the reason for this is that cro allows a maximum mask size of /16
+	// the default cidr mask is set by defaultCIDRMask env and is currently 26
+	//
+	// the current logic checks that the cidr block does not overlap with the cluster machine cidr range
+	// there is a follow on JIRA to check overlap with pod and service cidr blocks -> https://issues.redhat.com/MDGAPI-1000
+	for cidrRange, potentialDefault := range cidrRanges {
+		_, cidrRangeNet, err := net.ParseCIDR(cidrRange)
+		if err != nil {
+			fmt.Println("error parsing cidr range for default cidr block", err)
+		}
+
+		// our potential default
+		potentialDefaultIP, _, err := net.ParseCIDR(potentialDefault)
+		if err != nil {
+			fmt.Println("error parsing potential default cidr range for default cidr block", err)
+		}
+		// loop for as long as potential default is a valid CIDR in range
+		for cidrRangeNet.Contains(potentialDefaultIP) {
+
+			// create CIDR adding the default /26 mask
+			_, defaultNet, err := net.ParseCIDR(fmt.Sprintf("%s/26", potentialDefaultIP.String()))
+			if err != nil {
+				continue
+			}
+			// increment potential IP
+			potentialDefaultIP = incrementIPForDefaultCIDR(potentialDefaultIP)
+
+			// check for overlap
+			// if it overlaps continue with the loop
+			if clusterNet.Contains(defaultNet.IP) || defaultNet.Contains(clusterNet.IP) {
+				continue
+			}
+			// return the first available option that does not overlap
+			return defaultNet, nil
+		}
+	}
+	// if the loop finishes, it means that we have gone through all available cidr blocks in the available ranges in aws
+	// return an error that cro was unable to find an option
+	return nil, errorUtil.New("could not find a default cidr block")
 }
 
 //subnetExists is a helper function for checking if a subnet exists with a specific cidr block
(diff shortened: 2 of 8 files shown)


pkg/providers/aws/config.go
```diff
@@ -79,6 +80,38 @@ func NewDefaultConfigMapConfigManager(client client.Client) *ConfigMapConfigMana
 	return NewConfigMapConfigManager(DefaultConfigMapName, DefaultConfigMapNamespace, client)
 }
 
+func (m *ConfigMapConfigManager) ReconcileNetworkStrategy(ctx context.Context, c client.Client, cidr string) error {
+	cm, err := resources.GetConfigMapOrDefault(ctx, m.client, types.NamespacedName{Name: m.configMapName, Namespace: m.configMapNamespace}, BuildDefaultConfigMap(m.configMapName, m.configMapNamespace))
+	if err != nil {
+		return errorUtil.Wrapf(err, "failed to get aws strategy config map %s in namespace %s", m.configMapName, m.configMapNamespace)
+	}
+	var network struct {
+		Production struct {
```
laurafitzgerald, line 89: Need to improve this so it works for all tiers.

pkg/providers/aws/provider_postgres.go
```diff
@@ -210,6 +210,7 @@ func (p *PostgresProvider) CreatePostgres(ctx context.Context, pg *v1alpha1.Post
 		}
 		logger.Infof("created security group %s", aws.StringValue(securityGroup.StandaloneSecurityGroup.GroupName))
 	}
+	return nil, "don't create the postgres", errorUtil.New("dont create the postgres")
```
laurafitzgerald, line 213: This will be removed, it's just to allow testing of the network creation without creating a postgres.

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1510,18 +1515,61 @@ func getNetworkProviderConfig(ctx context.Context, configManager ConfigManager,
...
+// getNonOverlappingDefaultCIDR returns a non overlapping cidr block based on the OpenShift vpc cidr block
+// the default mask is /26
+// for other masks the user is required to provide their own via config
+func getNonOverlappingDefaultCIDR(ctx context.Context, client client.Client, ec2Api ec2iface.EC2API, logger *logrus.Entry) (*net.IPNet, error) {
+	clusterVpc, err := getClusterVpc(ctx, client, ec2Api, logger)
+	if err != nil {
+		errMsg := "failed to get cluster vpc for cidr block"
+		return nil, errorUtil.Wrap(err, errMsg)
+	}
+	clusterCIDR := *clusterVpc.CidrBlock
```
ciaranRoche, line 1555: might be worth adding a nil check here. Alternatively you can use `aws.StringValue(pointer_to_my_string)` it will return an empty string if the value is nil. Avoiding the nil pointer, but will still need to check the value.

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1510,18 +1515,61 @@ func getNetworkProviderConfig(ctx context.Context, configManager ConfigManager,
...
+	// increment the second byte of the ip by 4 bytes or by 1, e.g. if the ip is 10.0.0.0 then we want 10.1.0.0
+	// as the maximum cro allowed mask size is /16, we are using /26 for the default but the following default
+	// will apply in the case of /16 to /26
+	ip[1] = ip[1] + byte(1)
+
+	_, defaultCIDR, err := net.ParseCIDR(fmt.Sprintf("%s/%s", ip, defaultCIDRMask))
+	if err != nil {
+		return nil, errorUtil.Wrap(err, "error parsing standalone default cidr block")
+	}
+	return defaultCIDR, nil
```
ciaranRoche, line 1572: probably should have some sort of loop here, to validate the `defaultCIDR` is valid and does not overlap with the `clusterCIDR`

pkg/providers/aws/config.go
ciaranRoche: do we need this? We can use the configManager.ReadStorageStrategy(ctx, providers.NetworkResourceType, tier) to get the the data from the config map

pkg/providers/aws/provider_postgres.go
ciaranRoche: I was expecting a less serious error message :rofl:

pkg/providers/aws/config.go
laurafitzgerald: Resolving this one as I've removed the logic to update the config map as discussed 👍

pkg/providers/aws/provider_postgres.go
laurafitzgerald: How about if I add a booooop? 😝

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1510,18 +1515,67 @@ func getNetworkProviderConfig(ctx context.Context, configManager ConfigManager,
 		return nil, errorUtil.Wrap(err, "failed to unmarshal aws vpc create config")
 	}
 
-	if vpcCreateConfig.CidrBlock == nil {
-		return nil, errorUtil.New(fmt.Sprintf("CidrBlock in _network strategy tier %s can not be nil", tier))
+	// if the config map is found and the _network block contains an entry, that is returned for use in the network creation
+	if vpcCreateConfig != nil && vpcCreateConfig.CidrBlock != nil && *vpcCreateConfig.CidrBlock != "" {
```
ciaranRoche, line 1519: I think we can get rid of the first nil check, as we dont dereference the `vpcCreateConfig` object. We will catch the nil pointer in `vpcCreateConfig.CidrBlock != nil`

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1510,18 +1515,67 @@ func getNetworkProviderConfig(ctx context.Context, configManager ConfigManager,
...
+		_, vpcCidr, err := net.ParseCIDR(*vpcCreateConfig.CidrBlock)
+		if err != nil {
+			return nil, errorUtil.Wrap(err, "failed to parse cidr block from _network strategy")
+		}
+		cidrMask, _ := vpcCidr.Mask.Size()
```
ciaranRoche, line 1524: I think we can remove this declaration, it is only used in the log

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1510,18 +1515,67 @@ func getNetworkProviderConfig(ctx context.Context, configManager ConfigManager,
...
+
+		logger.Infof("found vpc cidr block %s/%d in network strategy tier %s", vpcCidr.IP.String(), cidrMask, tier)
```
ciaranRoche, line 1526: This log format can be shortened to : 
```
logger.Infof("found vpc cidr block %s in network strategy tier %s", vpcCidr.String(), tier)
```

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1510,18 +1515,67 @@ func getNetworkProviderConfig(ctx context.Context, configManager ConfigManager,
...
+		}
+
+		_, defaultCIDR, err := net.ParseCIDR(fmt.Sprintf("%s/%s", ip, defaultCIDRMask))
+		if err != nil {
+			return nil, errorUtil.Wrap(err, "error parsing standalone default cidr block")
+		}
+		if isValidCIDRRange(defaultCIDR) {
+			return defaultCIDR, nil
+		}
+	}
```
ciaranRoche, line 1578: I think it would good to have an escape here, just in the off chance that there is an overlap. As we add to this logic to check for an overlap on the Machine, Pod and Service CIDR ranges. It would be good to error if we can not find a valid CIDR.

pkg/providers/aws/provider_postgres.go
ciaranRoche: That would be too many o`s, would have to close the MR

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1510,18 +1515,67 @@ func getNetworkProviderConfig(ctx context.Context, configManager ConfigManager,
...
-	logger.Infof("found vpc cidr block %s/%d in network strategy tier %s", vpcCidr.IP.String(), cidrMask, tier)
-	return vpcCidr, nil
+	// a default cidr is generated and updated to the config map, that is returned for use in the network creation
+	return defaultCIDR, nil
+}
+
+// getNonOverlappingDefaultCIDR returns a non overlapping cidr block based on the OpenShift vpc cidr block
+// the default mask is /26
+// for other masks the user is required to provide their own via config
+func getNonOverlappingDefaultCIDR(ctx context.Context, client client.Client, ec2Api ec2iface.EC2API, logger *logrus.Entry) (*net.IPNet, error) {
```
ciaranRoche, line 1543: looking at the params, it might be worth changing this to target `NetworkProvider`

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1510,18 +1515,67 @@ func getNetworkProviderConfig(ctx context.Context, configManager ConfigManager,
...
+	clusterVpc, err := getClusterVpc(ctx, client, ec2Api, logger)
+	if err != nil {
+		errMsg := "failed to get cluster vpc for cidr block"
+		return nil, errorUtil.Wrap(err, errMsg)
+	}
+
+	clusterCIDR := aws.StringValue(clusterVpc.CidrBlock)
```
ciaranRoche, line 1550: small nit pick, we probably could just pass `aws.StringValue(clusterVpc.CidrBlock)` into `net.ParseCIDR()`

ciaranRoche (commented):
Looks good, mostly some nitpicks inline.

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1600,27 +1605,17 @@ func isValidCIDRRange(CIDR *net.IPNet) bool {
...
 // 		* has a cidr range between \16 and \26
 // 		* does not overlap with cluster vpc cidr block
-func validateStandaloneCidrBlock(ctx context.Context, client client.Client, ec2Api ec2iface.EC2API, logger *logrus.Entry, vpcCidrBlock *net.IPNet) error {
+func validateStandaloneCidrBlock(validateCIDR *net.IPNet, clusterCIDR *net.IPNet) error {
 	// validate has a cidr range between \16 and \26
-	if !isValidCIDRRange(vpcCidrBlock) {
-		return errorUtil.New(fmt.Sprintf("%s is out of range, block sizes must be between `/16` and `/26`, please update `_network` strategy", vpcCidrBlock.String()))
-	}
-
-	clusterVPC, err := getClusterVpc(ctx, client, ec2Api, logger)
```
laurafitzgerald, line 1555: I'm moving this out because we now have a loop where this function could be called more than once.

go.mod
```diff
@@ -11,7 +11,6 @@ require (
 	github.com/elazarl/goproxy/ext v0.0.0-20190421051319-9d40249d3c2f // indirect
 	github.com/ghodss/yaml v1.0.1-0.20180820084758-c7ce16629ff4 // indirect
 	github.com/go-openapi/spec v0.19.3
-	github.com/gomodule/redigo v2.0.0+incompatible // indirect
```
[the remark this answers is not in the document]
laurafitzgerald: make vendor/fix is removing it. I tried adding it back in but it removed it again. It does seem strange as there's no change to dependencies in this change. 🤷

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1510,18 +1525,62 @@ func getNetworkProviderConfig(ctx context.Context, configManager ConfigManager,
...
+		} else {
+			ip[1] = byte(0)
+			ip[0] = ip[0] + byte(1)
+		}
+
+		_, defaultCIDR, err := net.ParseCIDR(fmt.Sprintf("%s/%s", ip, defaultCIDRMask))
+		if err != nil {
+			return nil, errorUtil.Wrap(err, "error parsing standalone default cidr block")
+		}
+		if err = validateStandaloneCidrBlock(defaultCIDR, clusterCidr); err == nil {
```
laurafitzgerald, line 1580: This was what I meant to use originally. 🤦‍♀️

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -101,6 +103,7 @@ type NetworkManager interface {
 	DeleteNetworkPeering(*NetworkPeering) error
 	IsEnabled(context.Context) (bool, error)
 	DeleteBundledCloudResources(context.Context) error
+	getNonOverlappingDefaultCIDR(context.Context) (*net.IPNet, error)
```
ciaranRoche, line 106: We don't export this, so I think we can remove it from the interface.

pkg/providers/aws/cluster_network_provider.go
ciaranRoche: I don't think this will work, as the first none valid CIDR will return. I was thinking something like this : 

```
func (n *NetworkProvider) getNonOverlappingDefaultCIDR(ctx context.Context) (*net.IPNet, error) {
	// get cluster vpc
	clusterVpc, err := getClusterVpc(ctx, n.Client, n.Ec2Api, n.Logger)
	if err != nil {
		errMsg := "failed to get cluster vpc for cidr block"
		return nil, errorUtil.Wrap(err, errMsg)
	}

	// parse CIDR subnet
	clusterCidrIp, clusterCidr, err := net.ParseCIDR(aws.StringValue(clusterVpc.CidrBlock))
	if err != nil {
		return nil, errorUtil.Wrap(err, "error parsing cluster cidr block")
	}

	// find non overlapping default
	for i := 0; clusterCidr.Contains(incrementIP(clusterCidrIp, i)); i++ {

		// increment potential default
		nextNetwork := incrementIP(clusterCidrIp, i)

		// parse potential default to type CIDR
		_, defaultCIDR, err := net.ParseCIDR(nextNetwork.String())
		if err != nil {
			n.Logger.Errorf("error parsing default cidr '%s' skipping setting default : %v", nextNetwork.String(), err)
			continue
		}

		// validate the potential cidr
		if clusterCidr.Contains(defaultCIDR.IP) || defaultCIDR.Contains(clusterCidr.IP) {
			continue
		}

		return defaultCIDR, nil
	}
	// if loop ends return error
	return nil, errors.New("could not find default CIDR")
}
```
note, I didnt run this so it might not even compile :see_no_evil: , just we should check for every possible potential default, if none is found we should error

R-Lawton:
verified that this functionality works on a osd cluster

pkg/providers/aws/cluster_network_provider.go
ciaranRoche: @laurafitzgerald I think this can be removed

laurafitzgerald: 👍  done

ciaranRoche (commented):
NP : I think we could squash the commits

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -63,6 +63,8 @@ const (
 	// see https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeVpcPeeringConnections.html
 	filterVpcPeeringRequesterId = "requester-vpc-info.vpc-id"
 	filterVpcPeeringAccepterId  = "accepter-vpc-info.vpc-id"
+	// default cidr values
+	defaultCIDRMask = "26"
```
MStokluska, line 67: @laurafitzgerald not sure what I'm missing but I can't see this being used anywhere

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -153,12 +155,21 @@ func (n *NetworkProvider) CreateNetwork(ctx context.Context, vpcCidrBlock *net.I
 		//By default, `integreatly.org/clusterID`.
 		//
 		//NOTE - Once a VPC is created we do not want to update it. To avoid changing cidr block
+		clusterVPC, err := getClusterVpc(ctx, n.Client, n.Ec2Api, n.Logger)
+		if err != nil {
+			return nil, errorUtil.Wrap(err, "failed to get cluster vpc")
```
MStokluska, line 160: hey @laurafitzgerald, just more of a question, I see that the getClusterVpc can return different errors, not entirely sure how does the errorUtil.Wrap work, but from my initial understanding, if this error is met we are going to return "failed to get cluster vpc" - if so, maybe it would be better to return the error from getClusterVpc function, which can be different, for example - fail to get clusterId/getting vpcs/ vpc not found?

MStokluska: ah, nvm! I think I just need to look into .Wrap stuff when I get a minute! It looks ok now! :)

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1510,18 +1523,90 @@ func getNetworkProviderConfig(ctx context.Context, configManager ConfigManager,
...
+	return defaultCIDR, nil
+}
+
+// getNonOverlappingDefaultCIDR returns a non overlapping cidr block based on the OpenShift vpc cidr block
+// the default mask is /26
+// for other masks the user is required to provide their own via config
+func (n *NetworkProvider) getNonOverlappingDefaultCIDR(ctx context.Context) (*net.IPNet, error) {
+	clusterVpc, err := getClusterVpc(ctx, n.Client, n.Ec2Api, n.Logger)
+	if err != nil {
+		errMsg := "failed to get cluster vpc for cidr block"
```
MStokluska, line 1552: just a question, is there a reason why we are not keeping it the same as other .Wrap logging?

ciaranRoche: Its left over from a copy and paste, in other parts of cro, we return a status along with the error. So its easier to declare the `errMsg` and return it in both the status and errorUtil.Wrap. 

Example : https://github.com/integr8ly/cloud-resource-operator/blob/master/pkg/providers/aws/provider_postgres.go#L145

pkg/providers/aws/cluster_network_provider.go
laurafitzgerald: good spot, should be used lower down. will update.

pkg/providers/aws/cluster_network_provider.go
laurafitzgerald: I have to make another change so I'll clean this up. 👍

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1510,18 +1523,90 @@ func getNetworkProviderConfig(ctx context.Context, configManager ConfigManager,
...
+		}
+
+		// our potential default
+		potentialDefaultIP, _, err := net.ParseCIDR(potentialDefault)
+		if err != nil {
+			fmt.Println("error ", err)
+		}
+		// loop for as long as potential default is a valid CIDR in range
+		for cidrRangeNet.Contains(potentialDefaultIP) {
+			fmt.Println("inside the loop")
```
MStokluska, line 1588: I think this could go?

laurafitzgerald: 🤦‍♀️ yep

ciaranRoche: I liked it, can we add this back in, plzzzz.

ciaranRoche:
Overrode Prow, as this functionality is not testing in Prow.

PR #204, opened 2021-01-22
Mgdapi 1000 adding to existing non overlapping function to include Service CIDR and POD CIDR

R-Lawton:
## Overview

[Jira: ](https://issues.redhat.com/browse/MGDAPI-1000?devStatusDetailDialog=pullrequest)

## Verification

- Clone this branch
- Run `make cluster/prepare`
- Run `make run`
### Verify that the CRO VPC des not overlap with clusterVPC, service CIDR and pod CIDR
- Patch the config map to have no default to emulate the scenario were the user doesnt input there own CIDR block

`oc patch configMap cloud-resources-aws-strategies -n cloud-resource-operator --type='json' -p '[{"op": "add", "path": "/data/_network", "value":"{ \"development\": { \"createStrategy\": { \"CidrBlock\": \"''\" } } }"}]'`

- Create a postgres cr
`Run make cluster/seed/managed/postgres`

- Verify that the default cidr block has been generated and the vpc is created.

`aws ec2 describe-vpcs --filter "Name=tag-value,Values=RHMI Cloud Resource VPC"`

The expected cidr block will depend on what the clusters service CIDR, Pod CIDr and machiene cidr block. For a cluster that has the following values:` Machine CIDR: 10.4.0.0/16	Service CIDR  10.5.0.0/16 Pod CIDR10.0.0.0/14` you should get a `CRO vpc of 10.6.0.0/26`

Check the

```
  "Vpcs": [
        {
            "CidrBlock": "10.6.0.0/26",
            "DhcpOptionsId": "dopt-4ed78528",
            "State": "available",
            "VpcId": "vpc-09876ef4a72624c6b",
            "OwnerId": "408612754352",
            "InstanceTenancy": "default",
            "CidrBlockAssociationSet": [
                {
                    "AssociationId": "vpc-cidr-assoc-0ebed1fe0c89c2485",
                    "CidrBlock": "10.6.0.0/26",
                    "CidrBlockState": {
                        "State": "associated"
                    }
                }
            ],
            "IsDefault": false,
            "Tags": [
                {
                    "Key": "Name",
                    "Value": "RHMI Cloud Resource VPC"
                },
                {
                    "Key": "integreatly.org/clusterID",
                    "Value": "rhmi-clouds-rl-5dwfx"
                }
            ]
        },
```


pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1563,6 +1564,17 @@ func (n *NetworkProvider) getNonOverlappingDefaultCIDR(ctx context.Context) (*ne
 		"10.255.255.255/8":  fmt.Sprintf("10.0.0.0/%s", defaultCIDRMask),
 		"172.31.255.255/12": fmt.Sprintf("172.16.0.0/%s", defaultCIDRMask),
 	}
+	podCIDR, err := n.getPodCIDR(ctx, logger)
+	logger.Info("podCidr: ", podCIDR)
```
ciaranRoche, line 1568: NP, is this log necessary as an `info`

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1607,6 +1619,38 @@ func (n *NetworkProvider) getNonOverlappingDefaultCIDR(ctx context.Context) (*ne
...
+	networkConf := &v12.Network{}
+	err := n.Client.Get(ctx, client.ObjectKey{Name: "cluster"}, networkConf)
+	if err != nil {
+		return nil, errorUtil.Wrap(err, "failed to get network kind")
+	}
+	var podCIDR *net.IPNet
+	for _, entry := range networkConf.Spec.ClusterNetwork {
+		_, podCIDR, err = net.ParseCIDR(entry.CIDR)
+		if err != nil {
+			panic(err)
```
ciaranRoche, line 1632: Not a good idea to panic, best to return the error and let it be handled at the controller level. 
```suggestion
			return nil, errorUtil.Wrap(err, "error parsing pod cidr")
```

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1607,6 +1619,38 @@ func (n *NetworkProvider) getNonOverlappingDefaultCIDR(ctx context.Context) (*ne
...
+		}
+	}
+	return podCIDR, err
+}
+func (n *NetworkProvider) getServiceCIDR(ctx context.Context, logger *logrus.Entry) (*net.IPNet, error) {
+	networkConf := &v12.Network{}
+	err := n.Client.Get(ctx, client.ObjectKey{Name: "cluster"}, networkConf)
+	if err != nil {
+		return nil, errorUtil.Wrap(err, "failed to get network cr")
+	}
```
ciaranRoche, line 1642: We are fetching this twice for both `getPodCIDR` and `getServiceCIDR`, might be worth fetching one, and passing the object to each function as a parameter.

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1595,7 +1607,7 @@ func (n *NetworkProvider) getNonOverlappingDefaultCIDR(ctx context.Context) (*ne
 
 			// check for overlap
 			// if it overlaps continue with the loop
-			if clusterNet.Contains(defaultNet.IP) || defaultNet.Contains(clusterNet.IP) {
+			if clusterNet.Contains(defaultNet.IP) || defaultNet.Contains(clusterNet.IP) || podCIDR.Contains(defaultNet.IP) || defaultNet.Contains(podCIDR.IP) || serviceCIDR.Contains(defaultNet.IP) || defaultNet.Contains(serviceCIDR.IP) {
```
ciaranRoche, line 1610: At a glance this is hard to read, i think we should break it out into separate blocks. 
```
	if foundClusterNet {
		if clusterNet.Contains(foo) || foo.Contains(clusterNet) {
			continue
		}
	}
	
	if foundPodNet {
		if podNet.Contains(foo) || foo.Contains(podNet) {
			continue
		}
	}
	
	if foundServiceNet {
		if serviceNet.Contains(foo) || foo.Contains(serviceNet) {
			continue
		}
	}
}
```
Bit more verbose but I think it reads better.

pkg/providers/aws/provider_postgres_test.go
```diff
@@ -445,6 +445,25 @@ func buildTestInfra() *v12.Infrastructure {
...
 
+func buildTestNetwork() *v12.Network {
+	return &v12.Network{
+		ObjectMeta: controllerruntime.ObjectMeta{
+			Name: "cluster",
+		},
+		Spec: v12.NetworkSpec{
+			ClusterNetwork: []v12.ClusterNetworkEntry{
+				{
+					CIDR:       "10.0.0.0/14",
```
ciaranRoche, line 456: Can we add a test were no CIDR's are returned

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1607,6 +1619,38 @@ func (n *NetworkProvider) getNonOverlappingDefaultCIDR(ctx context.Context) (*ne
 	return nil, errorUtil.New("could not find a default cidr block")
 }
 
+func (n *NetworkProvider) getPodCIDR(ctx context.Context, logger *logrus.Entry) (*net.IPNet, error) {
```
ciaranRoche, line 1622: this takes in `logger` but never uses it.

pkg/providers/aws/cluster_network_provider.go
R-Lawton: no sorry that was for testing while the the pr is in wip

pkg/providers/aws/cluster_network_provider.go
R-Lawton: yup was meant to change that before i un draft the pr good catch

pkg/providers/aws/cluster_network_provider.go
R-Lawton: good idea

pkg/providers/aws/cluster_network_provider.go
R-Lawton: what do you think about something like this 
```
     // checking clusterVpc (machine) cidr
if clusterNet.Contains(defaultNet.IP) || defaultNet.Contains(clusterNet.IP) ||
    //checking pod cidr overlap
    podCIDR.Contains(defaultNet.IP) || defaultNet.Contains(podCIDR.IP) ||
    //checking service cidr overlap
    serviceCIDR.Contains(defaultNet.IP) || defaultNet.Contains(serviceCIDR.IP) {
    continue
}
```
Easier to read but less need for the verboseness of the extra if statements

ciaranRoche: i think we need the `found` check. In the case that the `Net` object is nil. Due to how the getters are, we could return a nill object. 
if we run the `.Contains` on a nill object we will get a panic.

R-Lawton: ahh ok cool

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1547,12 +1547,15 @@ func (n *NetworkProvider) ReconcileNetworkProviderConfig(ctx context.Context, co
 // the default mask is /26
 // for other masks the user is required to provide their own via config
 func (n *NetworkProvider) getNonOverlappingDefaultCIDR(ctx context.Context) (*net.IPNet, error) {
+	foundClusterNet := true
 	clusterVpc, err := getClusterVpc(ctx, n.Client, n.Ec2Api, n.Logger)
 	if err != nil {
+		foundClusterNet = false
```
ciaranRoche, line 1553: We don't need that here in this case, `getClusterVpc` will return an error if no VPC is found, so we will never get beyond this point if now VPC is found.

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1592,21 +1611,73 @@ func (n *NetworkProvider) getNonOverlappingDefaultCIDR(ctx context.Context) (*ne
...
 			return defaultNet, nil
 		}
 	}
+
 	// if the loop finishes, it means that we have gone through all available cidr blocks in the available ranges in aws
 	// return an error that cro was unable to find an option
 	return nil, errorUtil.New("could not find a default cidr block")
 }
 
+func (n *NetworkProvider) getPodCIDR(networkConf *v12.Network) (*net.IPNet, bool, error) {
```
ciaranRoche, line 1647: we don't need to target `NetworkProvider` for these functions, as we don't use it.

ciaranRoche (commented):
Looks good, can we add a test case, to ensure the error `could not find a default cidr block` is returned if there is an overlap with every network

MStokluska:
I have verified with Rachel on her cluster, it works as expected!

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1592,21 +1609,71 @@ func (n *NetworkProvider) getNonOverlappingDefaultCIDR(ctx context.Context) (*ne
...
 			return defaultNet, nil
 		}
 	}
+
 	// if the loop finishes, it means that we have gone through all available cidr blocks in the available ranges in aws
 	// return an error that cro was unable to find an option
 	return nil, errorUtil.New("could not find a default cidr block")
 }
 
+func getPodCIDR(networkConf *v12.Network) (*net.IPNet, bool, error) {
```
ciaranRoche, line 1643: Nitpick - its self explanatory but, a small comment on each one of these getters (getPodCIDR and getServiceCIDR)

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1592,21 +1609,71 @@ func (n *NetworkProvider) getNonOverlappingDefaultCIDR(ctx context.Context) (*ne
...
+	var podCIDR *net.IPNet
```
ciaranRoche, line 1644: Nitpick - the var name, `podCIDR` is actually an `IPNet` object. So maybe something like `podNet` would make more sense here.

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1592,21 +1609,71 @@ func (n *NetworkProvider) getNonOverlappingDefaultCIDR(ctx context.Context) (*ne
 			}
 			// increment potential IP
 			potentialDefaultIP = incrementIPForDefaultCIDR(potentialDefaultIP)
-
-			// check for overlap
-			// if it overlaps continue with the loop
+			//check for overlap
+			// if it overlaps continue
+			//no need for boolean check as if cluster vpc fails it wont get this far
```
ciaranRoche, line 1614: Nitpick - I think we can remove this comment, the checks are self explanatory and called out in the comment above the loop

ciaranRoche (commented):
looks good, just some nit picks :+1:

ciaranRoche (commented):
/lgtm

Just one more thing, can we squash those commits into one logical commit.

PR #208, opened 2021-03-31
Go vesion bumped to 1.16.2

MStokluska:
## Overview
Go version used was updated to 1.16.2

## Verification
- Confirm that e2e tests were successful
- Install operator on the byoc cluster using make commands and run the set of tests by using the `make cluster/seed commands`


.travis.yml
```diff
@@ -4,10 +4,10 @@ sudo: required
 dist: bionic
 
 go:
-  - 1.13.x
+  - 1.16.2
 
 env:
-  - GO111MODULE=on
+  - GO111MODULE=off
```
austincunningham, line 10: Asking out of curiosity, why is this changed to off?

MStokluska: The `make test/unit/ci` was kept failing. A solution to it was to either remove the steps from the before_install or change GO111MODULE to off.

PR #210, opened 2021-04-22
[WIP] Add ability to turn off "block public acccess" and open GETObject to …

briangallagher:
…public

## Overview

Jira: <jira-id>

## Verification

- Clone this branch
- Run `make cluster/prepare`
- Run `make run`
- Ensure...

## Checklist

### DIFF
--- a/apis/integreatly/v1alpha1/types/types.go
+++ b/apis/integreatly/v1alpha1/types/types.go
@@ -27,6 +27,8 @@ type ResourceTypeSpec struct {
 	Tier       string     `json:"tier"`
 	SkipCreate bool       `json:"skipCreate,omitempty"`
 	SecretRef  *SecretRef `json:"secretRef"`
+	RemoveBlockOnPublicAccess  bool `json:"removeBlockOnPublicAccess,omitempty"`
+	AllowPublicGetObjectAccess  bool `json:"allowPublicGetObjectAccess,omitempty"`
 }
 
 type StatusPhase string
--- a/pkg/providers/aws/credentials.go
+++ b/pkg/providers/aws/credentials.go
@@ -42,6 +42,7 @@ var (
 				"s3:PutBucketTagging",
 				"s3:PutBucketPublicAccessBlock",
 				"s3:PutEncryptionConfiguration",
+				"s3:PutBucketPolicy",
 				"ec2:DescribeVpcs",
 				"ec2:DescribeSubnets",
 				"ec2:DescribeSecurityGroups",
--- a/pkg/providers/aws/provider_blobstorage.go
+++ b/pkg/providers/aws/provider_blobstorage.go
@@ -409,7 +409,7 @@ func (p *BlobStorageProvider) reconcileBucketCreate(ctx context.Context, bs *v1a
 	defer p.exposeBlobStorageMetrics(ctx, bs)
 
 	if foundBucket != nil {
-		if err = reconcileS3BucketSettings(aws.StringValue(foundBucket.Name), s3svc); err != nil {
+		if err = reconcileS3BucketSettings(aws.StringValue(foundBucket.Name), s3svc, bs); err != nil {
 			errMsg := fmt.Sprintf("failed to set s3 bucket settings %s", *foundBucket.Name)
 			return croType.StatusMessage(errMsg), errorUtil.Wrapf(err, errMsg)
 		}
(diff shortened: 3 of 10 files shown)


apis/integreatly/v1alpha1/types/types.go
```diff
@@ -27,6 +27,8 @@ type ResourceTypeSpec struct {
 	Tier       string     `json:"tier"`
 	SkipCreate bool       `json:"skipCreate,omitempty"`
 	SecretRef  *SecretRef `json:"secretRef"`
+	RemoveBlockOnPublicAccess  bool `json:"removeBlockOnPublicAccess,omitempty"`
+	AllowPublicGetObjectAccess  bool `json:"allowPublicGetObjectAccess,omitempty"`
```
ciaranRoche, line 31: normally we use the strategy config pass configuration to resources in cro, this pattern goes against how we handle all other resources. I don't know the longevity of cro, considering things like crossplane. I dont have a strong opinion, only If cro is staying around for the long term I would think it is best use a strat config to get this configuration. 

Maybe @pmccarthy  can give some indication here, is it worth updating to match other resources, or is it okay as is? I am conscious of time to refactor.

pkg/providers/aws/credentials.go
```diff
@@ -42,6 +42,7 @@ var (
 				"s3:PutBucketTagging",
 				"s3:PutBucketPublicAccessBlock",
 				"s3:PutEncryptionConfiguration",
+				"s3:PutBucketPolicy",
```
ciaranRoche, line 45: When adding new permissions to this list, you need to update this document too https://docs.google.com/document/d/11UwspEIuOvowWoILaI_3orQdBdDX_geCx9YABUmL3Kc/edit?usp=sharing

pkg/providers/aws/provider_blobstorage.go
```diff
@@ -493,9 +506,42 @@ func reconcileS3BucketSettings(bucket string, s3svc s3iface.S3API) error {
...
 
+func getAllowPublicGetObjectAccess(bucketName string) ([]byte, error) {
+
+	s3Policy := map[string]interface{}{
+		"Version": "2012-10-17",
+		"Statement": []map[string]interface{}{
+			{
+				"Sid":    "Stmt1618957960008",
+				"Effect": "Allow",
+				"Principal": "*",
```
ciaranRoche, line 535: Do we really need to add get access for everyone, can we not limit this to the cluster? 

Maybe we can add a condition to this to limit it to the cluster VPC at the very least https://docs.aws.amazon.com/AmazonS3/latest/userguide/access-control-block-public-access.html#access-control-block-public-access-policy-status

pkg/providers/aws/provider_blobstorage.go
```diff
@@ -465,14 +465,27 @@ func getS3buckets(s3svc s3iface.S3API) ([]*s3.Bucket, error) {
...
+		blockPublicAcls = false
+		blockPublicPolicy = false
+		ignorePublicAcls = false
+		restrictPublicBuckets = false
+	}
+
 	_, err := s3svc.PutPublicAccessBlock(&s3.PutPublicAccessBlockInput{
 		Bucket: aws.String(bucket),
 		PublicAccessBlockConfiguration: &s3.PublicAccessBlockConfiguration{
-			BlockPublicAcls:       aws.Bool(defaultBlockPublicAcls),
```
ciaranRoche, line 472: These defaults were an ask from BU not to have anything public, I know this is a customer issue, but has it been highlighted to the BU that this change is happening.

pkg/providers/aws/provider_blobstorage.go
briangallagher: The S3 URL is hard coded into the HTML of the 3Scale Portal, hence the need to make public

pkg/providers/aws/credentials.go
briangallagher: Will update if we get confirmation from BU, thanks!

pkg/providers/aws/provider_blobstorage.go
briangallagher: Waiting for feedback from BU, thanks.

apis/integreatly/v1alpha1/types/types.go
briangallagher: The reason for this approach was to modify one specific blobStorage/S3 bucket as apposed to all via the strategy map. However, we recently added the ability to overwrite aws strategies on a per instance basis for Postgres and Redis to integreatly. We could add code to do the same for BlobStorage and then take the approach suggested. 
@pmccarthy Do you have a preference on approach?

pmccarthy: @briangallagher @ciaranRoche  I don't have a particularly strong preference either. If there is an immediate need to have the permissions updated for specific customers then we should try make CRO flexible enough to support per instance configs, based on the work Brian has done recently. However, if this change is something we plan to roll out to all customers across the fleet then there's no need to add per instance config support to S3 IMO.

PR #212, opened 2021-05-17
update CRO CSVs

MStokluska:
## Overview
Adding CSV 0.24.1 that was generated as part of RHOAM 1.6.1 release.


Makefile
```diff
@@ -4,13 +4,13 @@ IMAGE_NAME ?= cloud-resource-operator
...
 NAMESPACE=cloud-resource-operator
-PREV_VERSION=0.24.0
+PREV_VERSION=0.24.1
 VERSION=0.25.0
 COMPILE_TARGET=./tmp/_output/bin/$(IMAGE_NAME)
 UPGRADE ?= true
 CHANNEL ?= rhmi
 
-PREVIOUS_OPERATOR_VERSIONS="0.24.0,0.23.0"
+PREVIOUS_OPERATOR_VERSIONS="0.24.1,0.24.0,0.23.0"
```
austincunningham, line 13: Don't recall having to do this in the passed @MStokluska Can you confirm why we need 3 versions here , is this due to the bundling ?

MStokluska: that is due to the bundles and how they work. So right now if you want to be 0.25.0 you will need to update from 0.24.1 which updates from 0.24.0 which updates from 0.23.0.

austincunningham: :+1:

austincunningham: /lgtm
/approve

PR #213, opened 2021-06-04
MGDAPI-1857: Add the ApplyImmediately flag to rds db engine upgrades

austincunningham:
## Overview

Jira: https://issues.redhat.com/browse/MGDAPI-1857

## Verification

- Clone this branch
- Edit the engine version on line https://github.com/integr8ly/cloud-resource-operator/blob/master/pkg/providers/aws/provider_postgres.go#L55 to `"10.13"`
- Run `make cluster/prepare`
- Run `make run`
- Edit the example-postgres cr and add applyImmediately: true to the spec
```yaml
spec:
  applyImmediately: true
  secretRef:
    name: example-postgres-sec
  tier: development
  type: managed
```
- Wait for rds to finish provisioning
- Stop `make run` ctrl-c
- Edit the engine version on line https://github.com/integr8ly/cloud-resource-operator/blob/master/pkg/providers/aws/provider_postgres.go#L55 to `"10.15"`
- `make run` again
- Confirm that the version change is applied immediately either in the aws console or with aws cli [describe-db-instances](https://docs.aws.amazon.com/cli/latest/reference/rds/describe-db-instances.html)
```bash
# This returns the instance Identifier
oc get postgres example-postgres -o json | jq '.metadata.annotations.resourceIdentifier'
"mwcollabbyocf88tpcloudresourceopera-tquz"
# Check the status to see , should go to `upgrading` after re-running the operator and then `available` once its finished upgrading  
$ aws rds describe-db-instances --db-instance-identifier=mwcollabbyocf88tpcloudresourceopera-tquz --region=eu-west-2 |  jq '.DBInstances[0].DBInstanceStatus'
# This returns the info on the rds instance engine version
$ aws rds describe-db-instances --db-instance-identifier=mwcollabbyocf88tpcloudresourceopera-tquz --region=eu-west-2 |  jq '.DBInstances[0].EngineVersion'
```
## Checklist

### DIFF
--- a/pkg/providers/aws/provider_postgres.go
+++ b/pkg/providers/aws/provider_postgres.go
@@ -309,7 +309,7 @@ func (p *PostgresProvider) createRDSInstance(ctx context.Context, cr *v1alpha1.P
 
 	// check if found instance and user strategy differs, and modify instance
 	logger.Infof("found existing rds instance: %s", *foundInstance.DBInstanceIdentifier)
-	mi, err := buildRDSUpdateStrategy(rdsCfg, foundInstance)
+	mi, err := buildRDSUpdateStrategy(rdsCfg, foundInstance, cr)
 
 	if err != nil {
 		errMsg := fmt.Sprintf("error building update config for rds instance: %s", *foundInstance.DBInstanceIdentifier)
@@ -647,7 +647,7 @@ func (p *PostgresProvider) isLastResource(ctx context.Context, namespace string)
 }
 
 // verifies if there is a change between a found instance and the configuration from the instance strat and verified the changes are not pending
-func buildRDSUpdateStrategy(rdsConfig *rds.CreateDBInstanceInput, foundConfig *rds.DBInstance) (*rds.ModifyDBInstanceInput, error) {
+func buildRDSUpdateStrategy(rdsConfig *rds.CreateDBInstanceInput, foundConfig *rds.DBInstance, cr *v1alpha1.Postgres) (*rds.ModifyDBInstanceInput, error) {
 	logrus.Infof("verifying that %s configuration is as expected", *foundConfig.DBInstanceIdentifier)
 	updateFound := false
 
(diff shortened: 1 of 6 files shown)


pkg/providers/aws/provider_postgres.go
```diff
@@ -61,6 +61,7 @@ const (
 	defaultPostgresUserKey               = "user"
 	defaultPostgresPasswordKey           = "password"
 	defaultStorageEncrypted              = true
+	defalutApplyImmediately              = true
```
laurafitzgerald, line 64: should it update immediately in all cases or is it something we want to be able to override in certain cases, e.g. RHOAM?

austincunningham: its only applied when an engine upgrade is detected

laurafitzgerald: yeah i think that's my question. I guess I'm thinking in the case of non rhoam users. that they may want to allow for the engine upgrades to only happen in the maintenance window. so we should have a way to override that that integreatly/rhoam can use.

austincunningham: Made this configurable via CR, updated the verification steps to include the CR changes.

psturc, line 64:
```suggestion
	defaultApplyImmediately              = true
```

psturc:
just a small typo, otherwise lgtm. going to verify it on my cluster

pkg/providers/aws/provider_postgres.go
austincunningham: :+1:

psturc:
ran the CRO with postgres engine v10.13
Screenshot 2021-06-10 at 13 32 16
reran CRO with postgres engine set to v10.15 and `spec.applyImmediately: true` in postgres CR -> upgrade has started immediately
Screenshot 2021-06-10 at 13 53 11
verified the new version in AWS:
Screenshot 2021-06-10 at 14 00 21

/lgtm

austincunningham:
/approve
really bold of me want to see if it works

austincunningham:
/remove-approve
it does

laurafitzgerald:
Did one final check after code change requests.
`git checkout v0.24.0`
`make run`
`make cluster/seed/managed/postgres`
`git checkout <this branch>`
Added ApplyImmediately to the Postgres cr - had to apply the new crd to the cluster first.
The rds was still in. installing state but once completed reported 
```
 "PendingModifiedValues": {
                "EngineVersion": "10.15"
            },
```
```
  "DBInstanceStatus": "upgrading",
```
/lgtm
/approve

PR #216, opened 2021-06-09
WIP: MGDAPI-1716 - add STS support

matskiv:
## Overview
Jira: https://issues.redhat.com/browse/MGDAPI-1716
This is initial implementation of STS support.

I managed to implement it without the use of temporary file, so there is actually not a lot of new code (some AWS specific bits in CreateSessionFromStrategy function and then pkg/providers/aws/credentials_sts_manager.go).
But then there is a refactor around credential handling in aws provider, to support STS use case which doesn't need to make CredentialRequest CRs. 
Also, I split non-STS related refactor into it's own commit.

Keeping as WIP because new unit tests need to be added.

## Verification
TODO - there is a lot to document, will need some time.. But I can show it in action if anyone is interested.

## Checklist

### DIFF
--- a/pkg/providers/aws/credentials.go
+++ b/pkg/providers/aws/credentials.go
@@ -141,9 +141,7 @@ type Credentials struct {
 //go:generate moq -out credentials_moq.go . CredentialManager
 type CredentialManager interface {
 	ReconcileProviderCredentials(ctx context.Context, ns string) (*Credentials, error)
-	ReconcileSESCredentials(ctx context.Context, name, ns string) (*Credentials, error)
-	ReoncileBucketOwnerCredentials(ctx context.Context, name, ns, bucket string) (*Credentials, *v1.CredentialsRequest, error)
-	ReconcileCredentials(ctx context.Context, name string, ns string, entries []v1.StatementEntry) (*v1.CredentialsRequest, *Credentials, error)
+	ReconcileBucketOwnerCredentials(ctx context.Context, name, ns, bucket string) (*Credentials, error)
 }
 
 var _ CredentialManager = (*CredentialMinterCredentialManager)(nil)
@@ -163,33 +161,25 @@ func NewCredentialMinterCredentialManager(client client.Client) *CredentialMinte
 
 //ReconcileProviderCredentials Ensure the credentials the AWS provider requires are available
 func (m *CredentialMinterCredentialManager) ReconcileProviderCredentials(ctx context.Context, ns string) (*Credentials, error) {
-	_, creds, err := m.ReconcileCredentials(ctx, m.ProviderCredentialName, ns, operatorEntries)
+	creds, err := m.reconcileCredentials(ctx, m.ProviderCredentialName, ns, operatorEntries)
 	if err != nil {
 		return nil, err
 	}
 	return creds, nil
 }
 
-func (m *CredentialMinterCredentialManager) ReconcileSESCredentials(ctx context.Context, name, ns string) (*Credentials, error) {
-	_, creds, err := m.ReconcileCredentials(ctx, name, ns, sendRawMailEntries)
+func (m *CredentialMinterCredentialManager) ReconcileBucketOwnerCredentials(ctx context.Context, name, ns, bucket string) (*Credentials, error) {
+	creds, err := m.reconcileCredentials(ctx, name, ns, buildPutBucketObjectEntries(bucket))
 	if err != nil {
 		return nil, err
 	}
 	return creds, nil
 }
 
-func (m *CredentialMinterCredentialManager) ReoncileBucketOwnerCredentials(ctx context.Context, name, ns, bucket string) (*Credentials, *v1.CredentialsRequest, error) {
-	cr, creds, err := m.ReconcileCredentials(ctx, name, ns, buildPutBucketObjectEntries(bucket))
-	if err != nil {
-		return nil, nil, err
-	}
-	return creds, cr, nil
-}
-
-func (m *CredentialMinterCredentialManager) ReconcileCredentials(ctx context.Context, name string, ns string, entries []v1.StatementEntry) (*v1.CredentialsRequest, *Credentials, error) {
+func (m *CredentialMinterCredentialManager) reconcileCredentials(ctx context.Context, name string, ns string, entries []v1.StatementEntry) (*Credentials, error) {
 	cr, err := m.reconcileCredentialRequest(ctx, name, ns, entries)
 	if err != nil {
-		return nil, nil, errorUtil.Wrapf(err, "failed to reconcile aws credential request %s", name)
+		return nil, errorUtil.Wrapf(err, "failed to reconcile aws credential request %s", name)
 	}
 	err = wait.PollImmediate(time.Second*5, time.Minute*5, func() (done bool, err error) {
 		if err = m.Client.Get(ctx, types.NamespacedName{Name: cr.Name, Namespace: cr.Namespace}, cr); err != nil {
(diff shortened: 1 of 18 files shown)


matskiv:
Failing unit test passes locally and doesn't seem to be related to my changes.

I'll have to investigate this:
```
 INFO[2021-06-09T22:03:00Z] === RUN   TestNetworkProvider_ReconcileNetworkProviderConfig/
verify_default_cidr_block_and_no_error_on_empty_cidr_block 
...
INFO[2021-06-09T22:03:00Z]     cluster_network_provider_test.go:1378: ReconcileNetworkProviderConfig() 
got = 172.16.0.0/26, want 10.6.0.0/26 
```

scripts/sts/sts-secret.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
```
laurafitzgerald, line 1: I had done something similar with an expected secret but put it into the config/secrets folder and used templating for providing the ARN https://github.com/integr8ly/cloud-resource-operator/pull/215/files#diff-dddcdd3e58af2f907e4c8812ca7953563c29620af6205c38aea99dcc66b0eeb0R1

laurafitzgerald: Although seems like you are doing lots more in the script so this is probably better.

pkg/providers/aws/credentials.go
```diff
@@ -136,14 +137,24 @@ type Credentials struct {
...
+	TokenFilePath   string
 }
 
 //go:generate moq -out credentials_moq.go . CredentialManager
 type CredentialManager interface {
 	ReconcileProviderCredentials(ctx context.Context, ns string) (*Credentials, error)
-	ReconcileSESCredentials(ctx context.Context, name, ns string) (*Credentials, error)
-	ReoncileBucketOwnerCredentials(ctx context.Context, name, ns, bucket string) (*Credentials, *v1.CredentialsRequest, error)
-	ReconcileCredentials(ctx context.Context, name string, ns string, entries []v1.StatementEntry) (*v1.CredentialsRequest, *Credentials, error)
+	ReconcileBucketOwnerCredentials(ctx context.Context, name, ns, bucket string) (*Credentials, error)
```
laurafitzgerald, line 147: ReconcileBucketOwnerCredentials. Is this the one we put in the secret in the case of the managed-api?

pkg/providers/aws/credentials.go
```diff
@@ -136,14 +137,24 @@ type Credentials struct {
...
+}
+
+func NewCredentialManager(client client.Client) CredentialManager {
```
laurafitzgerald, line 150: This appears to be used mostly in the aws providers but I see two entries where it's used in controllers. postgressnapshot and redissnapshot controllers. Should the controllers be provider independent?

pkg/providers/aws/credentials.go
```diff
@@ -136,14 +137,24 @@ type Credentials struct {
...
+	ns, _ := k8sutil.GetOperatorNamespace()
+	_, err := getSTSCredentialsSecret(context.TODO(), client, ns)
+	if err == nil {
```
laurafitzgerald, line 153: Could we be more specific around the error handling here. I think the case is that If the error is isNotFound we use credentialsminter but if it is found we use sts?

pkg/providers/aws/credentials_sts_manager.go
```diff
@@ -0,0 +1,65 @@
...
+	if credentials.RoleArn == "" {
+		return nil, errorUtil.New(fmt.Sprintf("%s key is undefined in secret %s", defaultRoleARNKeyName, secret.Name))
+	}
+	if credentials.TokenFilePath == "" {
+		return nil, errorUtil.New(fmt.Sprintf("%s key is undefined in secret %s", defaultTokenPathKeyName, secret.Name))
+	}
+	return credentials, nil
+}
+
+func (m *STSCredentialManager) ReconcileBucketOwnerCredentials(ctx context.Context, name, ns, bucket string) (*Credentials, error) {
```
laurafitzgerald, line 57: minor code suggestion.
```suggestion
func (m *STSCredentialManager) ReconcileBucketOwnerCredentials(_ context.Context, _, _, _ string) (*Credentials, error) {
```

pkg/providers/aws/credentials_sts_manager.go
```diff
@@ -0,0 +1,65 @@
...
+func NewSTSCredentialManager(client client.Client) *STSCredentialManager {
+	ns, _ := k8sutil.GetOperatorNamespace()
+	return &STSCredentialManager{
+		OperatorNamespace: ns,
+		Client:            client,
+	}
+}
+
+//ReconcileProviderCredentials Ensure the credentials the AWS provider requires are available
+func (m *STSCredentialManager) ReconcileProviderCredentials(ctx context.Context, ns string) (*Credentials, error) {
```
laurafitzgerald, line 38: minor code suggestions. seems ns isn't used in the function 
```suggestion
func (m *STSCredentialManager) ReconcileProviderCredentials(ctx context.Context, _ string) (*Credentials, error) {
```

pkg/providers/aws/credentials.go
matskiv: In case of managed-api this one is used by 3Scale, so we end up copying it over to a Secret in 3Scale ns.
I am not sure if this answers the question though.

scripts/sts/sts-secret.yaml
matskiv: The script is not using this secret actually.
I am happy to move it to a Template in config/secrets.

pkg/providers/aws/credentials.go
matskiv: Good catch.
It seems like `CredentialManager` field in those two controllers is unused, so I'll just remove it.

pkg/providers/aws/credentials.go
matskiv: Yes, that makes sense.
But to properly handle other errors we would have to return the err from this function and handle it in the layers above. And I suspect that it might be anti-pattern to return errors from this "NewSomething" constructor-like function, so I'll have to look into this and maybe come up with different approach.

pkg/providers/aws/credentials_sts_manager.go
```diff
@@ -0,0 +1,65 @@
...
+		return nil, errorUtil.New(fmt.Sprintf("%s key is undefined in secret %s", defaultRoleARNKeyName, secret.Name))
+	}
+	if credentials.TokenFilePath == "" {
+		return nil, errorUtil.New(fmt.Sprintf("%s key is undefined in secret %s", defaultTokenPathKeyName, secret.Name))
+	}
+	return credentials, nil
+}
+
+func (m *STSCredentialManager) ReconcileBucketOwnerCredentials(ctx context.Context, name, ns, bucket string) (*Credentials, error) {
+	return nil, nil
```
austincunningham, line 58: Might be missing something here but what is the point of this function it just seems to return `nil, nil` ?

scripts/sts/create-rosa-sts.sh
```diff
@@ -0,0 +1,257 @@
+#!/bin/bash
+set -eux
```
austincunningham, line 2:
```suggestion
# Usage :
# ./create-rosa-sts.sh <Cluster name>

set -eux
```

pkg/providers/aws/credentials_sts_manager.go
matskiv: It is defined in the `CredentialManager` interface, so I had to implement it here as well.

austincunningham: Ah the joy of OO

PR #221, opened 2021-06-16
MGDAPI-1857:bump version to v0.27.0, add applyImmediately to client util, remove the pending modifications check for engine upgrades

austincunningham:
## Overview
- bump version to v0.27.0
- add applyImmediately to client util
- remove the pending modifications check for engine upgrades

Jira: MGDAPI-1857

## Verification

- Clone this branch
- Edit the engine version on line https://github.com/integr8ly/cloud-resource-operator/blob/master/pkg/providers/aws/provider_postgres.go#L55 to `"10.13"`
- Run `make cluster/prepare`
- Run `make run`
- Wait for rds to finish provisioning
- Stop `make run` ctrl-c
- Edit the engine version on line https://github.com/integr8ly/cloud-resource-operator/blob/master/pkg/providers/aws/provider_postgres.go#L55 to `"10.15"`
- `make run` again
- leave the operator run for about 40 min
- Edit the example-postgres cr and add applyImmediately: true to the spec
```yaml
spec:
  applyImmediately: true
  secretRef:
    name: example-postgres-sec
  tier: development
  type: managed
```
- Confirm that the version change is applied immediately either in the aws console or with aws cli [describe-db-instances](https://docs.aws.amazon.com/cli/latest/reference/rds/describe-db-instances.html)
```bash
# This returns the instance Identifier
oc get postgres example-postgres -o json | jq '.metadata.annotations.resourceIdentifier'
"mwcollabbyocf88tpcloudresourceopera-tquz"
# Check the status to see , should go to `upgrading` after re-running the operator and then `available` once its finished upgrading  
$ aws rds describe-db-instances --db-instance-identifier=mwcollabbyocf88tpcloudresourceopera-tquz --region=eu-west-2 |  jq '.DBInstances[0].DBInstanceStatus'
# This returns the info on the rds instance engine version
$ aws rds describe-db-instances --db-instance-identifier=mwcollabbyocf88tpcloudresourceopera-tquz --region=eu-west-2 |  jq '.DBInstances[0].EngineVersion'
```

## Checklist


pkg/providers/aws/provider_postgres.go
```diff
@@ -700,11 +700,12 @@ func buildRDSUpdateStrategy(rdsConfig *rds.CreateDBInstanceInput, foundConfig *r
 			return nil, errorUtil.Wrap(err, "invalid postgres version")
 		}
 		if engineUpgradeNeeded {
+			logrus.Info(fmt.Sprintf("Engine upgrade found, the current EngineVersion is %s and is upgrading to %s", *foundConfig.EngineVersion, *rdsConfig.EngineVersion))
 			mi.EngineVersion = rdsConfig.EngineVersion
 			if cr.Spec.ApplyImmediately {
 				mi.ApplyImmediately = aws.Bool(cr.Spec.ApplyImmediately)
 			}
-			updateFound = true
+			return mi, nil
```
laurafitzgerald, line 708: should this be inside the if statement? in this case we return that an update is required even when it may not be.

laurafitzgerald: or alternatively we add a condition to line 711 where we don't return nil if applyimmediately is true

laurafitzgerald:
@austincunningham as part of the verification I think we need to have a wait inbetween aupgrding cro and applying the change to mimic any delay we might have in integreatly to ensure that if there are other product upgrade steps needed that the applyImmediatly does indeed get applied if there is some wait time and the pending modifications are already sitting in the rds.

pkg/providers/aws/provider_postgres.go
laurafitzgerald: ah okay I see the `engineUpgradeNeeded` if statement now. ignore the above. 👍

pkg/client/util_test.go
```diff
@@ -260,15 +263,16 @@ func TestReconcilePostgres(t *testing.T) {
...
+				ctx:              context.TODO(),
+				client:           fake.NewFakeClientWithScheme(scheme),
+				deploymentType:   "workshop",
+				tier:             "development",
+				productName:      "test",
+				name:             "test",
+				ns:               "test",
+				secretName:       "test",
+				secretNs:         "test",
+				applyImmediately: false,
```
laurafitzgerald, line 275: could we add a case here which tests applyImmediately is true.

austincunningham: sure

austincunningham: actually I will modify the 
```
			name: "test modification function",
```

pkg/providers/aws/provider_postgres.go
laurafitzgerald: @austincunningham after some further digging this will have the api called on every reconcile. I think it would be better to put his change into the check on engineupgrade in the verifyPendingModifications function. what we want to do is allow for an additional api call if engineupgrade is sitting in pendingmodifications on line 739 we check if the modifyInput object is equal tot he existing pending modifications. if that's true we don't call another update. so we could pass down the applyImmediately value and if it's true and there is a pending modification then return true rather than false

laurafitzgerald: as well as that we should revert this line to prevent sending a request to the api on every reconcile until the version is upgraded.

pkg/providers/aws/provider_postgres.go
```diff
@@ -700,11 +700,12 @@ func buildRDSUpdateStrategy(rdsConfig *rds.CreateDBInstanceInput, foundConfig *r
...
 		}
 	}
 	if !updateFound || !verifyPendingModification(mi, foundConfig.PendingModifiedValues) {
```
laurafitzgerald, line 711:
```suggestion
			updateFound = true
		}
	}
	if (!updateFound || !verifyPendingModification(mi, foundConfig.PendingModifiedValues)) && !mi.ApplyImmediately  {
```

sorry for all the updates @austincunningham but I believe this is the way to go. :)

laurafitzgerald: Feel free to give me a shout and we can chat through it.

pkg/providers/aws/provider_postgres.go
laurafitzgerald: This suggestion is outdated by the more recent one so resolving.

pkg/providers/aws/provider_postgres.go
austincunningham: was going with 
```go
	if mi.EngineVersion != nil && pm.EngineVersion != nil {
		if *mi.EngineVersion == *pm.EngineVersion {
			pendingModifications = false
		}
		if cr.Spec.ApplyImmediately {
			logrus.Infof(fmt.Sprintf("Engine version pending modification changeing %s version to %s", *pm.EngineVersion, *mi.EngineVersion))
			pendingModifications = true
		}
	}
```
but all would work, not invested in any way in particular

austincunningham: Looks like that trows an nil pointer on the unit tests, if your happy with
```go
if (!updateFound || !verifyPendingModification(mi, foundConfig.PendingModifiedValues)) && mi.ApplyImmediately != nil  {
```

pkg/providers/aws/provider_postgres.go
```diff
@@ -700,14 +700,19 @@ func buildRDSUpdateStrategy(rdsConfig *rds.CreateDBInstanceInput, foundConfig *r
...
 			}
 			updateFound = true
 		}
 	}
-	if !updateFound || !verifyPendingModification(mi, foundConfig.PendingModifiedValues) {
+	if mi.ApplyImmediately != nil {
+		if (!updateFound || !verifyPendingModification(mi, foundConfig.PendingModifiedValues)) && !*mi.ApplyImmediately {
+			return nil, nil
+		}
+	} else if !updateFound || !verifyPendingModification(mi, foundConfig.PendingModifiedValues) {
```
laurafitzgerald, line 715: Is there some duplication here?

laurafitzgerald, line 715: code suggestion but the logic is the same either way.
```suggestion
		if (!updateFound || !verifyPendingModification(mi, foundConfig.PendingModifiedValues)) && (mi.ApplyImmediately == nil || !*mi.ApplyImmediately) {
```

austincunningham: this looks a bit tidier :+1:

laurafitzgerald:
Verified with Austin.
The postgres was in pending modification for the 10.15 upgrade
once the applyimmediately parameter was added to the cr the postgres went into upgrading state
Pending modifications was emptied and the enginer version updated to 10.15

PR #226, opened 2021-07-06
add the applyImmediately flag

austincunningham:
## Overview

Jira: MGDAPI-2253

Add the applyImmediately flag to redis 

## Verification

- Clone this branch
- Run `make cluster/prepare`
- Change the defaultEngineVersion to 5.0.3 on [/pkg/providers/aws/provider_redis.go](https://github.com/integr8ly/cloud-resource-operator/blob/master/pkg/providers/aws/provider_redis.go#L45)
```go
	defaultEngineVersion     = "5.0.3"
```
- Run `make run`
- Wait for redis to provision 
- Stop the Operator running locally
- Change the defaultEngineVersion back to 5.0.6
- Add the applyImmediately flag to the redis cr
```yaml
spec:
  applyImmediately: true
  secretRef:
    name: example-redis-sec
  tier: development
  type: managed
```
-  Run `make run`
- The redis should upgrade the engin version to 5.0.6 and you should see that the serviceUpdate has been applied in the logs . You should see the **UpdateActionStatus** is **complete** and the **ServiceUpdateName** is **elasticache-20210615-002**
```bash
INFO[0217] there are elasticache service update actions 1 available : [{
  Engine: "redis",
  NodesUpdated: "2/2",
  ReplicationGroupId: "mwcollabbyoc7rk2xcloudresourceopera-4poh",
  ServiceUpdateName: "elasticache-20210615-002",
  ServiceUpdateRecommendedApplyByDate: 2021-06-30 15:59:59 +0000 UTC,
  ServiceUpdateReleaseDate: 2021-06-23 16:00:00 +0000 UTC,
  ServiceUpdateSeverity: "critical",
  ServiceUpdateStatus: "available",
  ServiceUpdateType: "security-update",
  SlaMet: "n/a",
  UpdateActionAvailableDate: 2021-07-06 09:25:35.668 +0000 UTC,
  UpdateActionStatus: "complete",
  UpdateActionStatusModifiedDate: 2021-07-06 09:25:35.73 +0000 UTC
}] 

```
Also in the aws console you should see that redis `Update Action Status` is up to date 
image

### DIFF
--- a/pkg/providers/aws/provider_redis.go
+++ b/pkg/providers/aws/provider_redis.go
@@ -276,7 +276,7 @@ func (p *RedisProvider) createElasticacheCluster(ctx context.Context, r *v1alpha
 	}
 
 	// check if any modifications are required to bring the elasticache instance up to date with the strategy map.
-	modifyInput, err := buildElasticacheUpdateStrategy(ec2Svc, elasticacheConfig, foundCache, replicationGroupClusters, logger)
+	modifyInput, err := buildElasticacheUpdateStrategy(ec2Svc, elasticacheConfig, foundCache, replicationGroupClusters, logger, r.Spec.ApplyImmediately)
 	if err != nil {
 		errMsg := "failed to build elasticache modify strategy"
 		return nil, croType.StatusMessage(errMsg), errorUtil.Wrap(err, errMsg)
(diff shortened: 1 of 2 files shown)


pkg/providers/aws/provider_redis.go
```diff
@@ -741,6 +741,10 @@ func buildElasticacheUpdateStrategy(ec2Client ec2iface.EC2API, elasticacheConfig
 			modifyInput.SnapshotWindow = elasticacheConfig.SnapshotWindow
 			updateFound = true
 		}
+
+		if applyImmediately {
```
philbrookes, line 745: Does this only set it to true? If it were false line 745 would not enter this if, right? Is that OK?

austincunningham: Good point at some time in the future we may want to set this to false on aws

PR #228, opened 2021-07-07
force apply any pending modifications on service upates for redis tha…

laurafitzgerald:
## Overview

Jira: <jira-id>

## Verification

- On a cluster where there is a redis which has the service update in pending modification

- Verify that the serviceUpdate entry with the offending service update is present in aws-cloud-resource-strategies config map.
- Verify that the redis cr contains the applyImmediately flag as true
- Check the `serviceUpdateApplyDate` is before time.Now()
- If all of the above are satisfied, 
- update the image to `quay.io/austincunningham/cloud-resource-operator:v0.29.0`

You should see logs similar to this -> 
```
	"Checking status of Update Action = %s ", *update.ServiceUpdateName)
	"UpdateActionStatus = %s ", *update.UpdateActionStatus)
         "ServiceUpdateSeverity = %s ", *update.ServiceUpdateSeverity)
	"ServiceUpdateStatus = %s ", *update.ServiceUpdateStatus)
```

You should also see the redis going into a modifying state
And after some time the UpdateActionStatus of the clsuter should no longer be `update available` but `up to date`

## Checklist


austincunningham:
# Verification
- `make cluster/prepare`
- `make cluster/seed/managed/redis`
-  `make run`
- wait for redis to install
- update the redis cr and add the applyImmediately flag to true
- operator doesn't crash
/lgtm
/approve

laurafitzgerald:
## Verification Steps taken on cluster with two redis in state `update available`

### First run (isolated to one redis by custom image)

Cluster went into `waiting-to-start` and then `in-progress` for ~40 minutes and then went into `up to date`. Metrics report same
No errors in logs. The cluster was in `available` status for much of the update. A very short period of `modifying` status was observed.

### Second run (isolated to the other redis by custom image)

ModifyCacheCluster function was removed from this run. PendingModifiedValues were watched but no values got pushed to PendingModifiedValues for the cluster.
Cluster went into  `in-progress`  and then went into `up to date`.

PR #231, opened 2021-07-20
Create LICENSE

austincunningham:
## Overview

Add licence to CRO, same as integreatly-operator

## Checklist

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
laurafitzgerald, line 189: Should this be updated?

austincunningham: good point , integreatly-operator is the same with the place holder still in place :(

looking for an example of where we have this done
aerogear/offix has the line removed https://github.com/aerogear/offix/blob/master/LICENSE#L187-L189
aerogear/aerogear-js-sdk has 
```
Copyright 2018 Aerogear
``` 
https://github.com/aerogear/aerogear-js-sdk/blob/master/LICENSE#L189

Guess we could stick in
```
Copyright 2019 Redhat
```

LICENSE
```diff
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright [yyyy] [name of copyright owner]
+   Copyright 2019 Redhat
```
matskiv, line 189:
```suggestion
   Copyright 2019 Red Hat, Inc.
```

matskiv: 🤔 should it be `2021` ?

austincunningham: Think it should be from when the repo was created . Thats a best guess.

PR #233, opened 2021-11-11
Mgdapi 1635

makslion:
## Overview

Jira: MGDAPI-1635

## Verification
In a perfect world, you would need to verify that CRO passes e2e tests on its own, CRO does well in case of the fresh install with RHOAM and doe not cause any issues in case of RHOAM upgrade. 

Sole CRO and E2Es:
- Get your hands on the BYOC cluster (to save nerves in case something goes wrong)
- Install CRO from this branch and run e2e tests 
- Verify all of them are passing 

> Below I'll be using "release" instead of "version of packagemanifests that refer to a new image" or how on earth it is named correctly? 

Fresh install of CRO with RHOAM
- get your hands on the BYOC cluster (since we would `useClusterStorage=false`) 
- use [the guide ](https://github.com/integr8ly/cloud-resource-operator/blob/master/Update_CRO_integreatly-operator.md#release-new-version-bundle-of-cloud-resource-operator) to release a new CRO version from this branch 
- use a master of RHOAM and make a new release to your own quay.io repo with a new release of CRO. There is also a [guide](https://github.com/RHCloudServices/integreatly-help/blob/master/guides/olm/install-upgrade-rhoam-bundle.md) but it seems every person has own preferred ways to generate bindles and index
- install a "new" RHOAM with a "new" CRO via olm 
- ensure installation and deletion succeeds 

Upgrade of CRO along with RHOAM upgrade 
- once again BYOC cluster 
- Here you would need to make bundles and index of CRO with `replaces` current version from integreatly. Or you can do the current release as well so it will not complain about nonexisting bundles 
- You would need to get two releases of RHOAM where latest one has a new CRO and replaces old one 
- Simulate upgrade scenario 
- Watch finalizers for CRO-related CRs get updated as well as installation succeeds 
- Ensure uninstallation completes with no issues 

## Checklist


pkg/providers/aws/config.go
```diff
@@ -26,7 +26,7 @@ import (
 const (
 	DefaultConfigMapName = "cloud-resources-aws-strategies"
 
-	DefaultFinalizer = "finalizers.cloud-resources-operator.integreatly.org"
+	DefaultFinalizer = "cloud-resources-operator.integreatly.org/finalizers"
```
MStokluska, line 29: @makslion we will need to handle this change somehow because when this change lands on an existing installation we won't be able to remove/reconcile finalizers.
So I think everywhere where we reconcile finalizers we need to check finalizer format > update if it's in old format.
Let me know if you have any questions, sample how this was done in intly: https://github.com/integr8ly/integreatly-operator/pull/1807/files

makslion: Seems like I can't verify it fast enough, so will just put here a theory on this. In the [integreatly operator](https://github.com/integr8ly/integreatly-operator/pull/1807/files#diff-239eca6331989433c0101fe047f78aa8ce2dab5047a733719972e99206a6d9faL315) we are setting finalizers by retrieving existing ones and appending them with a new one and then calling an `.Update` on the object. So if there was an old and not used finalizer it would be left there, hence the need in replacing it.
In the CRO it is a bit funny. If I'm not going crazy, we don't have any sort of `reconcileFinalizer` but instead, we are calling the [CreateFinalizer](https://github.com/integr8ly/cloud-resource-operator/blob/master/pkg/resources/finalizers.go#L48-L63) each loop and the "setting" is done by the [addFinalizer](https://github.com/integr8ly/cloud-resource-operator/blob/master/pkg/resources/finalizers.go#L19-L23) function. The `addFinalizer` checks for the existence of the finalizer we want and then if it is not found, we are reassigning that field. So in case when we will have an old finalizer it would not find the new one and will call the [setFinalizers](https://github.com/makslion/cloud-resource-operator/blob/MGDAPI-1635/vendor/k8s.io/apimachinery/pkg/apis/meta/v1/meta.go#L170) that would reassign (hence, replace) finalizer to be the new one. :thinking:

makslion:
e2e passed! :tada:

makslion:
Alright, so the OCM hates me and I wasn't able to personally verify that in case of the upgrade scenario finalizers will be replaced with a new format. However, I was able to [verify](https://drive.google.com/file/d/1i6c7Y80H3k-btdeO66OccDR-cWTwdBaT/view?usp=sharing) (sorry for the horrible video) that they get set to default value. So in case of the upgrade finalizers will be replaced. I still will try to sanity check myself on Monday, but I'm confident that this PR is ready for a review now.

laurafitzgerald:
Putting a hold on this so it doesn't get merged in error before we complete the planned upcoming release. 👍

makslion:
rebased. Keeping everything separate to make the life of a reviewer easier

laurafitzgerald:
@makslion can you share it for verification?

pkg/providers/aws/provider_postgressnapshot_test.go
```diff
@@ -83,8 +83,9 @@ func buildRdsClientMock(modifyFn func(*rdsClientMock)) *rdsClientMock {
 func buildTestPostgresSnapshotCr() *v1alpha1.PostgresSnapshot {
 	return &v1alpha1.PostgresSnapshot{
 		ObjectMeta: controllerruntime.ObjectMeta{
-			Name:      "test",
-			Namespace: "test",
+			Name:            "test",
+			Namespace:       "test",
+			ResourceVersion: fakeResourceVersion,
```
laurafitzgerald, line 88: Just a question. What caused the need to add this ResourceVersion to the object?

makslion:
I'll mention it on the standup, but it is still technically on hold until the new CRO arrives to save RHODS humans, right? :thinking:

adam-cattermole:
I have verified:
 - e2e tests all pass
 - Installing/uninstalling this version of CRO with RHOAM through OLM
 - Upgrade RHOAM version to one using this version of CRO through OLM

All seems to be working as intended. I see finalizers get updated:

finalizer-before

finalizer-after

I guess the last thing might be to squash some of the commits?

makslion:
squshed commits

adam-cattermole:
Nice one

/lgtm

PR #240, opened 2021-11-29
refactor to report the status of the postgres, if nil is returned fro…

laurafitzgerald:
…m the reconcile function, report the msg as returned, this covers secret, rds updates and service updates

## Overview

Jira: <jira-id>

## Verification

Upgrade

- Checkout this branch
- Run `make cluster/prepare`
- Checkout tag v0.30.0 
- Wait for the postgres to become available
- Checkout this branch
- Update postgres cr spec.applyImmediatley to true
- Run `make run`
- Watch the status of the rds in aws using `watch -n 1 "aws rds describe-db-instances --region=<region> | jq '.DBInstances[].DBInstanceStatus'"` vs the status of the cr. It should get updated to upgrading during the engine version upgrade and to maintenance when the service-update gets reflected to the cr with status
- `reconcileRDSInstance() in progress, current aws rds resource status is
    maintenance`

Install
- Run installation of this branch
- Verify Engine Version is 10.16
- Verify that serviceUpdate is applied
- Verify that the status of the aws gets reflected back to the cr during installation

## Checklist

### DIFF
--- a/controllers/postgres/postgres_controller.go
+++ b/controllers/postgres/postgres_controller.go
@@ -195,7 +195,7 @@ func (r *PostgresReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error
 		}
 
 		// create the postgres instance
-		ps, msg, err := p.CreatePostgres(ctx, instance)
+		ps, msg, err := p.ReconcilePostgres(ctx, instance)
 		if err != nil {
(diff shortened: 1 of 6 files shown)


MStokluska:
/lgtm verified on cluster with @laurafitzgerald and @austincunningham 
Code looks ok as well

PR #211, opened 2021-05-05
set default rds version to 10.15

briangallagher:
## Overview

Jira: <jira-id>

## Verification

- Clone this branch
- Run `make cluster/prepare`
- Run `make run`
- Ensure...

## Checklist
- [ ] This PR includes a change to an instance type, I have used script `hack/<provider>/supported_types.sh` and attached the result below

PR #243, opened 2021-12-01
MGDAPI-3225 add the engine version to the redis and postres CR

austincunningham:
## Overview

WIP PR for setting the engine versions in the postgres and redis CR's

## Verification

- Clone this branch
- Run `make cluster/prepare`
- Run `make run`
- Check the Postgres CR status when complete
```yaml
status:
  message: >-
    rds instance rhmicloudsccsqf8xmcloudresourceoper-eiw4 is as expected, aws
    rds status is available
  phase: complete
  provider: aws-rds
  secretRef:
    name: example-postgres-sec
  strategy: aws
  version: '10.16'
```
- Check the Redis CR status when complete
```yaml
status:
  message: 'successfully created and tagged, aws elasticache status is available'
  phase: complete
  provider: aws-elasticache
  secretRef:
    name: example-redis-sec
  strategy: aws
  version: 5.0.6
```

## Checklist


apis/integreatly/v1alpha1/types/types.go
```diff
@@ -44,11 +44,12 @@ func (sm StatusMessage) WrapError(err error) StatusMessage {
...
 // +kubebuilder:object:generate=true
 type ResourceTypeStatus struct {
-	Strategy  string        `json:"strategy,omitempty"`
-	Provider  string        `json:"provider,omitempty"`
-	SecretRef *SecretRef    `json:"secretRef,omitempty"`
-	Phase     StatusPhase   `json:"phase,omitempty"`
-	Message   StatusMessage `json:"message,omitempty"`
+	Strategy       string        `json:"strategy,omitempty"`
+	Provider       string        `json:"provider,omitempty"`
+	VersionDetails string        `json:"versionDetails,omitempty"`
```
laurafitzgerald, line 49: Naming Suggestion. wdyt about using `version` instead of `versionDetails`? Or if we think there could be other items we want to expose later, we could add a `config` type which contains `version` which could be extended at a later date? Also in terms of future proofing I think we're fairly safe to add a version. My guess is that any cloud provider would expose a version of the postgres at least but we should keep in mind that this Status applies to any and all resource objects managed by cro. 👍

pkg/providers/aws/provider_postgres.go
```diff
@@ -343,6 +343,13 @@ func (p *PostgresProvider) reconcileRDSInstance(ctx context.Context, cr *v1alpha
 			Database: *foundInstance.DBName,
 			Port:     int(*foundInstance.Endpoint.Port),
 		}
+		// set the rds engine version in the status
+		if cr.Status.VersionDetails != *foundInstance.EngineVersion {
+			cr.Status.VersionDetails = *foundInstance.EngineVersion
+			if err := p.Client.Update(ctx, cr); err != nil {
```
laurafitzgerald, line 349: I think we can just return `nil, message, err` here and the calling function will do a the p.Client.Update.

apis/integreatly/v1alpha1/types/types.go
austincunningham: I was aiming for generic, but would say that `version` is fairly universal no problem whit that change.

pkg/providers/aws/provider_postgres.go
austincunningham: yea wasn't to sure the in memory cr wouldn't get over written at some point after the return. worth trying it out to see .

austincunningham: after trying I don't need to return here just need to let it run on to the final return of the function

austincunningham:
unit test pass locally , may be a bit flaky
/test unit

ciaranRoche:
@austincunningham why is this being added? Can you not provide the version to the instance strategy?

austincunningham:
> @austincunningham why is this being added? Can you not provide the version to the instance strategy?
@ciaranRoche it's a good question we haven't been using the strategy to bump the default engine version for Redis or Postgres https://github.com/integr8ly/cloud-resource-operator/pull/211 , we were looking for a way to determine the current engine version in aws on RHOAM for Redis and Postgres and taught it might be useful to pass it back in the CR status.  

cc @laurafitzgerald

ciaranRoche:
Makes sense passing it back via the status, I don't think it is necessary to take in the version via the CR. As we would end up with 2 places in which we can define a status. 
We could get into a horrible place where the version defined in the strat is different to that in the CR. Which one would take precedence? It could get messy. 

Makes sense though to return info about the instance in the status. As CRO knows internally all the data about each instance. It might be worth returning an instance object in the status, that contains all the data. Instead of just the engine version. It would make it a bit more future proof.

austincunningham:
> d get into a horrible place where the version defined in the strat is different to that in the CR. Which one would take precedence? It could get messy.

Yea this is PR is only setting the engine version in the status not in the spec. Agree it would be messy if it did set version in the spec. Might be worth returning the instance object I might look into that, may expose to much information, but I suspect that version will always be available from other providers.

ciaranRoche:
my bad, i miss read the MR

/lgtm

pkg/providers/aws/provider_redis.go
```diff
@@ -308,6 +308,20 @@ func (p *RedisProvider) createElasticacheCluster(ctx context.Context, r *v1alpha
 			errMsg := fmt.Sprintf("failed to add tags to elasticache: %s", msg)
 			return nil, types.StatusMessage(errMsg), errorUtil.Wrap(err, errMsg)
 		}
+		cacheClusterOutput, err := cacheSvc.DescribeCacheClusters(&elasticache.DescribeCacheClustersInput{
```
ciaranRoche, line 311: Actually, just looking at this, we already Describe Cache Cluster on line 266. Probably worth moving, setting the `r.Status.Version` up there, using that output.

austincunningham: yea seen that was worried that 
```golang
		cacheClustersOutput, err := cacheSvc.DescribeCacheClusters(&elasticache.DescribeCacheClustersInput{})
```
would not return the same as 
```golang
		cacheClusterOutput, err := cacheSvc.DescribeCacheClusters(&elasticache.DescribeCacheClustersInput{
					CacheClusterId: cache.CacheClusterId,
		})
		
```
will debug and find out

ciaranRoche: In the loop 
```
for _, checkedCluster := range cacheClustersOutput.CacheClusters {
		if *checkedCluster.ReplicationGroupId == *foundCache.ReplicationGroupId {
			replicationGroupClusters = append(replicationGroupClusters, *checkedCluster)
		}
	}
```
The checkedCluster in the if statement, should be the same CacheCluster you are returning with 
```
cacheClusterOutput, err := cacheSvc.DescribeCacheClusters(&elasticache.DescribeCacheClustersInput{
					CacheClusterId: cache.CacheClusterId,
		})
```

Something like : 
```
	for _, checkedCluster := range cacheClustersOutput.CacheClusters {
		if checkedCluster.ReplicationGroupId != nil && foundCache.ReplicationGroupId != nil {
			if *checkedCluster.ReplicationGroupId == *foundCache.ReplicationGroupId {
				replicationGroupClusters = append(replicationGroupClusters, *checkedCluster)
				
				if checkedCluster.EngineVersion != nil && r.Status.Version != *checkedCluster.EngineVersion {
					r.Status.Version = *checkedCluster.EngineVersion
				}
			}
		}
	}
```

laurafitzgerald:
Hey @austincunningham I was thinking about this one a little more in the context of upgrade. Part of the intention of adding this was to report that the version needs an upgrade as quickly as possible so we can put the cr into an in progress state in case a different reconciler is creating the Postgres vs reconciling CRO. With this current implementation the version doesn't get reported until the end of the reconciliation so doesn't give us information any earlier that it would have had previously. Is that still something that we want to get from this change?

austincunningham:
> Hey @austincunningham I was thinking about this one a little more in the context of upgrade. Part of the intention of adding this was to report that the version needs an upgrade as quickly as possible so we can put the cr into an in progress state in case a different reconciler is creating the Postgres vs reconciling CRO. With this current implementation the version doesn't get reported until the end of the reconciliation so doesn't give us information any earlier that it would have had previously. Is that still something that we want to get from this change?

@laurafitzgerald Yes I think it is something we should be aiming for, will see about a refactor to report the version earlier to the status

pkg/providers/aws/provider_redis.go
```diff
@@ -255,26 +255,30 @@ func (p *RedisProvider) createElasticacheCluster(ctx context.Context, r *v1alpha
...
 	cacheClustersOutput, err := cacheSvc.DescribeCacheClusters(&elasticache.DescribeCacheClustersInput{})
 	if err != nil {
 		errMsg := "failed to describe clusters"
 		return nil, croType.StatusMessage(fmt.Sprintf(errMsg)), errorUtil.Wrapf(err, errMsg)
 	}
 	var replicationGroupClusters []elasticache.CacheCluster
 	for _, checkedCluster := range cacheClustersOutput.CacheClusters {
-		if *checkedCluster.ReplicationGroupId == *foundCache.ReplicationGroupId {
-			replicationGroupClusters = append(replicationGroupClusters, *checkedCluster)
+		if *checkedCluster.ReplicationGroupId != "" && *foundCache.ReplicationGroupId != "" {
```
ciaranRoche, line 265:
```suggestion
		if checkedCluster.ReplicationGroupId != nil && foundCache.ReplicationGroupId != nil {
```

pkg/providers/aws/provider_redis.go
```diff
@@ -308,6 +312,20 @@ func (p *RedisProvider) createElasticacheCluster(ctx context.Context, r *v1alpha
...
+		//	errMsg := fmt.Sprintf("error describing elasticache cluster %s", err)
+		//	return nil, croType.StatusMessage(errMsg), errorUtil.Wrap(err, errMsg)
+		//}
+
+		//for _, foundCacheCluster := range cacheClusterOutput.CacheClusters {
+		//	logrus.Info(foundCacheCluster)
+		//	if r.Status.Version != *foundCacheCluster.EngineVersion {
+		//		r.Status.Version = *foundCacheCluster.EngineVersion
+		//	}
+		//}
```
ciaranRoche, line 328: These commented out lines can be removed.

pkg/providers/aws/provider_redis.go
```diff
@@ -255,26 +255,30 @@ func (p *RedisProvider) createElasticacheCluster(ctx context.Context, r *v1alpha
...
+			}
 		}
 	}
 
+	// check elasticache phase
+	if *foundCache.Status != "available" {
+		logger.Infof("found instance %s current status %s", *foundCache.ReplicationGroupId, *foundCache.Status)
+		return nil, croType.StatusMessage(fmt.Sprintf("createReplicationGroup() in progress, current aws elasticache status is %s", *foundCache.Status)), nil
+	}
+	logger.Infof("found existing elasticache cluster %s", *foundCache.ReplicationGroupId)
```
ciaranRoche, line 280: I think this log message can be moved back. It makes sense to log that the cache has been found after we check that `foundCache` is not equal to nil

pkg/providers/aws/provider_redis.go
austincunningham: removing this nil check as it conflicts with changes @laurafitzgerald Laura has done on master

pkg/providers/aws/provider_redis.go
austincunningham: spotted on rebase :+1: and removed

pkg/providers/aws/provider_redis.go
austincunningham: :+1:

pkg/providers/aws/provider_redis.go
austincunningham: looks like the unit tests are failing without this  
image

pkg/providers/aws/provider_redis.go
```diff
@@ -271,11 +265,24 @@ func (p *RedisProvider) createElasticacheCluster(ctx context.Context, r *v1alpha
 	var replicationGroupClusters []elasticache.CacheCluster
 	for _, checkedCluster := range cacheClustersOutput.CacheClusters {
 		cluster := *checkedCluster
-		if resources.SafeStringDereference(cluster.ReplicationGroupId) == *foundCache.ReplicationGroupId {
-			replicationGroupClusters = append(replicationGroupClusters, *checkedCluster)
+		if checkedCluster.ReplicationGroupId != nil && foundCache.ReplicationGroupId != nil {
```
laurafitzgerald, line 268: The safeStringDereference below handles this being nil.
The function returns "" if it's nil

ciaranRoche: Yeah but it doesn't handle the case if `foundCache` is nil. 

I have no strong opinion on removing the nil check here or doing the stringDereference, or leaving both.

pkg/providers/aws/provider_redis.go
```diff
@@ -271,11 +265,24 @@ func (p *RedisProvider) createElasticacheCluster(ctx context.Context, r *v1alpha
 	var replicationGroupClusters []elasticache.CacheCluster
 	for _, checkedCluster := range cacheClustersOutput.CacheClusters {
 		cluster := *checkedCluster
```
ciaranRoche, line 267: NP : not related to this change, but if you are feeling it, I dont think we need to declare `cluster := *checkedCluster` we can just pass `checkedCluster.ReplicationGroupID` into the `SafeStringDereference`

pkg/providers/aws/provider_redis.go
```diff
@@ -270,12 +264,22 @@ func (p *RedisProvider) createElasticacheCluster(ctx context.Context, r *v1alpha
 	}
 	var replicationGroupClusters []elasticache.CacheCluster
 	for _, checkedCluster := range cacheClustersOutput.CacheClusters {
-		cluster := *checkedCluster
```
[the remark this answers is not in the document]
austincunningham: Ok will stick it back in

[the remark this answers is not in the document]
austincunningham: It's a small change so I am not invested either way . The change has been reverted for now.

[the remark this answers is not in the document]
laurafitzgerald: removing the line which dereferencees checkedCluster reintroduces the issue.
If you don't dereference it then safeDereference won't accept it.
*checkedCluster.ReplicationID on it's own doesn't safely dereference the replicationId as it's also a pointer which is what caused us the nil dereference originally.

pkg/providers/aws/provider_redis.go
```diff
@@ -255,12 +255,6 @@ func (p *RedisProvider) createElasticacheCluster(ctx context.Context, r *v1alpha
 		}
 		return nil, "started elasticache provision", nil
 	}
-
```
laurafitzgerald, line 258: I think that moving this check further down would slow down the reporting to the cr of the status being in modifying or maintenance etc.. 🤔  this was an issue for us before in upgrade scenarios.

austincunningham: Not sure it would as there is no return between this now and when `foundCache != nil` so should execute in order

laurafitzgerald: So if we remove this return we have to wait longer until we get to the cache being nil whereas with it in we get a return quicker if the cache is in any state other than available.

austincunningham: discussed off line

pkg/providers/aws/provider_postgres.go
```diff
@@ -52,7 +52,7 @@ const (
 	defaultAwsBackupRetentionPeriod      = 31
 	defaultAwsDBInstanceClass            = "db.t3.small"
 	defaultAwsEngine                     = "postgres"
-	defaultAwsEngineVersion              = "10.16"
+	DefaultAwsEngineVersion              = "10.16"
```
ciaranRoche, line 55: Why is this set to public?

austincunningham: We want it exposed in RHOAM

ciaranRoche: What is the reasoning for it being exposed in RHOAM? I am just curious as the instance Version will be exposed in the status.

austincunningham: Sure we are looking to solve a timing issue around upgrades to RDS and elasticache. And are looking to have as many tools as possible to help us solve the problem so we don't have to do something like this https://github.com/integr8ly/integreatly-operator/pull/2329 again

KevFan:
Verified engine version is available on both the Postrgres and Redis CRs :+1: 

image

/lgtm

austincunningham:
may be flaky
/test e2e

PR #244, opened 2021-12-01
MGDAPI-2906

laurafitzgerald:
# Changes Included in this pr
Refactor to how cro finds cluster vpc and route tables to handle case where the vpc and route tables are not labled with
shared/owned cluster tags - this happens in the case of an osd byovpc cluster installation
This approach should be compatible with non byovpc cases, e.g ccs and non ccs as the tags with be in place with either owned or shared

Instead of getting the cluster VPC by the tag, get the VPC ID from the associated subnets using either the owned or shared tag.
Instead of getting the cluster Route Tables by the tag, get the RouteTables associated with the Cluster VPC ID from above.

# Overview
Jira: https://issues.redhat.com/browse/MGDAPI-2906

# Verification
Please make sure there are no other instances of CRO running in the clusters you are testing.
Verification steps for BYOCVPC and non-BYOCVPC are identical. The only difference is a configuration of the aws resources (tags in particular).

Below you will find instructions on how to verify that VPC setup is as expected. After this you will find two sections with instructions on how to verify install and upgrade. 

Ideally you should do booth (install and upgrade) on each of the clusters (BYOCVPC and non-BYOCVPC)

## Case BYOVPC
Against a BYOC OSD - contact me or Maksym Vavilov for setting up byocvpc cluster

Verify there is no tags on the cluster VPC and Route Table
Cluster VPC (you might want to replace `mvavilov-vpc` with the name of your vpc):
```
aws ec2 describe-vpcs --filter Name=tag:Name,Values=mvavilov-vpc --output json | jq '.Vpcs[0].Tags'
```
Route Table (you might want to replace `mvavilov-vpc` with the name of your vpc):
```
VPC_ID=$(aws ec2 describe-vpcs --filter Name=tag:Name,Values=mvavilov-vpc --output json | jq '.Vpcs[0].VpcId')
aws ec2 describe-route-tables --filters "Name=vpc-id,Values=$VPC_ID" --output json | grep Tags
```
Tags should be empty and VPC should not have a cluster-assiciated tag. 

## CASE non-BYOVPC
Against a CCS OSD

Verify there IS a tag on the cluster VPC and Route Table

## Install
For the install we would go with the "local" approach. The main goal here is to check the logic. 
1. Checkout this branch
2. Run the `make cluster/prepare` 
3. Run `make run` to run the CRO. Keep an eye on the logs untill the end to ensure that there are no errors about the network, errors about credentials are acceptable a couple of times but should resolve.
4. Run `make test/e2e/local` in a separate terminal and ensure tests are suceeding 

## Upgrade
For an upgrade we will go with "bundles" approach. You can create your own bundles or use existing. 
Existing are located at [this repository](https://quay.io/repository/mvavilov/cloud-resource-operator?tab=tags) 
For the install you could use [v0.32.0](https://quay.io/repository/mvavilov/cloud-resource-operator?tag=index-v0.32.0&tab=tags) that was build from the master and as an upgrade version you can use [v0.33.1](https://quay.io/repository/mvavilov/cloud-resource-operator?tag=index-v0.33.1&tab=tags) that was build from this branch. 

1. Login to the cluster and run `make cluster/prepare` - this will create namespace for the CRO to be installed in
2. Create CatalogSource CR that would reference initial CRO. Here is a `cro-operators.yaml` template that could be reused along with `oc apply -f cro-operators.yaml`:
```
apiVersion: operators.coreos.com/v1alpha1
kind: CatalogSource
metadata:
  name: cro-operators
spec:
  sourceType: grpc
  image: quay.io/mvavilov/cloud-resource-operator:index-v0.32.0
```
> Note, that is you want to use own bundles you might want to
> replace the `image` field with you own index
3. Go to your cluster to the Operators -> Operators Hub and search for `Cloud Resource Operator`
4. Double check the referenced image before clicking `install`. In the `Installation mode` fields select `A specific namespace on the cluster` and then `cloud-resource-operator` from the dropdown. That namespace should be present after the wery first step. Set the `Update approval` to `Manual` - this will make verification faster
6. Populate cluster with redis and postgres by
```
make cluster/seed/managed/postgres
```
5. Ensure they (redis and postgres) after some time go into completed stage. While they are processing we will prepare the upgrade
6. Go the Operators -> Installed Operators -> Cloud Resource Operator -> Subscription. Locate the CatalogSource and edit the yaml so it (catalog source) point to a new version of CRO. In case you went with existing bundles it would be  `quay.io/mvavilov/cloud-resource-operator:index-v0.33.1`. After this field is updated you should see that upgrade is available.
7. Once Redis and Postgres CRs reporting the `complete` stage, initiate an upgrade 
8. During the upgrade you would see a new `cloud-resoure-operator-` pod go live. Open its logs and ensure that there are no errors about the network, errors about credentials are acceptable a couple of times but should resolve.
9. Delete them and make sure they aren't causing any troubles. 
10. Uninstall CRO and verify it cleans up it's resources:
```
aws ec2 describe-vpcs --output json | jq '.Vpcs[] | .Tags[0].Value'
```
This command will output names of all existing VPCs. If there is not VPC starting with `RHMI...` - CRO cleaned up it's resources. 

## Checklist
- [ ] This PR includes a change to an instance type, I have used script hack/<provider>/supported_types.sh and attached the result below

### DIFF
--- a/pkg/providers/aws/cluster_network_provider.go
+++ b/pkg/providers/aws/cluster_network_provider.go
@@ -57,6 +57,7 @@ const (
 	defaultNumberOfExpectedSubnets       = 2
 	clusterOwnedTagKeyPrefix             = "kubernetes.io/cluster/"
 	clusterOwnedTagValue                 = "owned"
+	clusterSharedTagValue                = "shared"
 	//network peering
 	tagVpcPeeringNameValue = "RHMI Cloud Resource Peering Connection"
 	// filter names for vpc peering connections
@@ -359,18 +360,12 @@ func (n *NetworkProvider) CreateNetworkConnection(ctx context.Context, network *
 		return nil, errorUtil.Wrap(err, "failure while reconciling standalone security group")
 	}
 
-	// a tag for identifying cluster owned vpc resources
-	clusterVpcRouteTableTag, err := getClusterOwnerTag(ctx, n.Client)
-	if err != nil {
-		return nil, errorUtil.Wrap(err, "error building cluster owner tag")
-	}
-
-	// find cluter vpc route tables using cluster vpc route table tag
+	// find cluster vpc route tables using associated cluster subnets
 	// multiples route tables can exist for a single vpc (main and secondary)
-	logger.Info("finding cluster route table(s)")
-	clusterVpcRouteTables, err := n.getVPCRouteTable(clusterVpcRouteTableTag)
+	logger.Info("finding cluster route table(s) using associated cluster subnets")
+	clusterVpcRouteTables, err := n.getClusterRouteTables(ctx)
 	if err != nil {
-		return nil, errorUtil.Wrap(err, "failure while getting vpc route table")
+		return nil, errorUtil.Wrap(err, "failure while getting vpc route tables")
 	}
 	logger.Infof("found %d cluster vpc route tables", len(clusterVpcRouteTables))
 
@@ -408,15 +403,9 @@ func (n *NetworkProvider) CreateNetworkConnection(ctx context.Context, network *
 		}
 	}
 
-	// get croOwner tag to use in getting standalone vpc route tables
-	croOwnerTag, err := getCloudResourceOperatorOwnerTag(ctx, n.Client)
-	if err != nil {
-		return nil, errorUtil.Wrap(err, "error generating cloud resource owner tag")
-	}
-
 	// get standalone vpc route table using cro owner tag
 	logger.Info("finding standalone route table(s)")
-	standAloneVpcRouteTables, err := n.getVPCRouteTable(genericToEc2Tag(croOwnerTag))
+	standAloneVpcRouteTables, err := n.getCRORouteTables(ctx)
 	if err != nil {
 		return nil, errorUtil.Wrap(err, "error getting standalone vpc route tables")
 	}
@@ -479,16 +468,10 @@ func (n *NetworkProvider) DeleteNetworkConnection(ctx context.Context, networkPe
 		}
 	}
 
-	// a tag for identifying cluster owned vpc resources
-	clusterVpcRouteTableTag, err := getClusterOwnerTag(ctx, n.Client)
-	if err != nil {
-		return errorUtil.Wrap(err, "error building cluster owner tag")
-	}
-
-	// find cluster vpc route tables using cluster vpc route table tag
+	// find cluster vpc route tables using cluster vpcID
 	// multiple route tables can exist for a single vpc (main and secondary)
 	logger.Info("finding cluster route table(s)")
-	clusterVpcRouteTables, err := n.getVPCRouteTable(clusterVpcRouteTableTag)
+	clusterVpcRouteTables, err := n.getClusterRouteTables(ctx)
 	if err != nil {
 		return errorUtil.Wrap(err, "failure while getting vpc route table")
 	}
@@ -533,7 +516,7 @@ func (n *NetworkProvider) CreateNetworkPeering(ctx context.Context, network *Net
 
 	clusterVpc, err := getClusterVpc(ctx, n.Client, n.Ec2Api, n.Logger)
 	if err != nil {
-		return nil, errorUtil.Wrap(err, "failed to get cluster vpc")
+		return nil, errorUtil.Wrap(err, "failed to get cluster vpc, no vpc found")
 	}
 
 	peeringConnection, err := n.getNetworkPeering(ctx, network)
@@ -1176,9 +1159,37 @@ func (n *NetworkProvider) reconcileStandaloneVPCSubnets(ctx context.Context, log
 	return subs, nil
 }
 
-// getVPCRouteTable will return a list of route tables based on a route table tag
+func (n *NetworkProvider) getClusterRouteTables(ctx context.Context) ([]*ec2.RouteTable, error) {
+	routeTables, err := n.Ec2Api.DescribeRouteTables(&ec2.DescribeRouteTablesInput{})
+	if err != nil {
+		return nil, errorUtil.Wrap(err, "failed to get route tables")
+	}
+
+	clusterVPC, err := getClusterVpc(ctx, n.Client, n.Ec2Api, n.Logger)
+	if err != nil {
+		return nil, errorUtil.Wrap(err, "failed to get cluster vpc")
+	}
+	var foundRouteTables []*ec2.RouteTable
+	for _, routeTable := range routeTables.RouteTables {
+		if routeTable != nil && *routeTable.VpcId == *clusterVPC.VpcId {
+			foundRouteTables = append(foundRouteTables, routeTable)
+		}
+	}
+	if len(foundRouteTables) == 0 {
+		return nil, errorUtil.New(fmt.Sprintf("could not find any route table with the associated vpc id %s", aws.StringValue(clusterVPC.VpcId)))
+	}
+	return foundRouteTables, nil
+}
+
+// getCRORouteTables will return a list of route tables based on a route table tag
 // we expect there to be route tables, if none are found we return an error
-func (n *NetworkProvider) getVPCRouteTable(routeTableTag *ec2.Tag) ([]*ec2.RouteTable, error) {
+func (n *NetworkProvider) getCRORouteTables(ctx context.Context) ([]*ec2.RouteTable, error) {
+	// get croOwner tag to use in getting standalone vpc route tables
+	croOwnerTag, err := getCloudResourceOperatorOwnerTag(ctx, n.Client)
+	if err != nil {
+		return nil, errorUtil.Wrap(err, "error generating cloud resource owner tag")
+	}
+
 	routeTables, err := n.Ec2Api.DescribeRouteTables(&ec2.DescribeRouteTablesInput{})
 	if err != nil {
 		return nil, errorUtil.Wrap(err, "failed to get route tables")
@@ -1187,13 +1198,13 @@ func (n *NetworkProvider) getVPCRouteTable(routeTableTag *ec2.Tag) ([]*ec2.Route
 	var foundRouteTables []*ec2.RouteTable
 	for _, routeTable := range routeTables.RouteTables {
 		routeTableTags := ec2TagsToGeneric(routeTable.Tags)
-		if tagsContains(routeTableTags, aws.StringValue(routeTableTag.Key), aws.StringValue(routeTableTag.Value)) {
+		if tagsContains(routeTableTags, aws.StringValue(genericToEc2Tag(croOwnerTag).Key), aws.StringValue(genericToEc2Tag(croOwnerTag).Value)) {
 			foundRouteTables = append(foundRouteTables, routeTable)
 		}
 	}
 
 	if len(foundRouteTables) == 0 {
-		return nil, errorUtil.New(fmt.Sprintf("could not find any route table with the tag key: %s and value: %s", aws.StringValue(routeTableTag.Key), aws.StringValue(routeTableTag.Value)))
+		return nil, errorUtil.New(fmt.Sprintf("could not find any route table with the tag key: %s and value: %s", aws.StringValue(genericToEc2Tag(croOwnerTag).Key), aws.StringValue(genericToEc2Tag(croOwnerTag).Value)))
 	}
 	return foundRouteTables, nil
 }
@@ -1747,6 +1758,14 @@ func getCloudResourceOperatorOwnerTag(ctx context.Context, client client.Client)
 // denoted -> ` kubernetes.io/cluster/<cluster-id>=owned
 // this utility function builds and returns the tag
 func getClusterOwnerTag(ctx context.Context, client client.Client) (*ec2.Tag, error) {
+	return getClusterTag(ctx, client, clusterOwnedTagValue)
+}
+
+func getClusterSharedTag(ctx context.Context, client client.Client) (*ec2.Tag, error) {
+	return getClusterTag(ctx, client, clusterSharedTagValue)
+}
+
+func getClusterTag(ctx context.Context, client client.Client, tagValue string) (*ec2.Tag, error) {
 	clusterID, err := resources.GetClusterID(ctx, client)
 	if err != nil {
 		return nil, errorUtil.Wrap(err, "error getting clusterID")
@@ -1754,8 +1773,8 @@ func getClusterOwnerTag(ctx context.Context, client client.Client) (*ec2.Tag, er
 
 	// a tag for identifying cluster owned vpcs
 	return &ec2.Tag{
-		Key:   aws.String(fmt.Sprintf("%s%s", clusterOwnedTagKeyPrefix, clusterID)),
-		Value: aws.String(clusterOwnedTagValue),
+		Key:   aws.String(getOSDClusterTagKey(clusterID)),
+		Value: aws.String(tagValue),
 	}, nil
 }
 
--- a/pkg/providers/aws/cluster_network_provider_test.go
+++ b/pkg/providers/aws/cluster_network_provider_test.go
@@ -49,6 +49,7 @@ const (
 	defaultSecurityGroupName      = "testsecuritygroup"
 	defaultSecurityGroupId        = "testSecurityGroupId"
 	defaultStandaloneRouteTableId = "testRouteTableId"
+	defaultClusterName            = "kubernetes.io/cluster/test"
 )
 
 func buildMockNetwork(modifyFn func(n *Network)) *Network {
@@ -120,7 +121,7 @@ func buildMockVpc(modifyFn func(*ec2.Vpc)) *ec2.Vpc {
 func buildMockEc2Tag(modifyFn func(*ec2.Tag)) *ec2.Tag {
 	mock := &ec2.Tag{
 		Key:   aws.String(defaultRHMISubnetTag),
-		Value: aws.String(dafaultInfraName),
+		Value: aws.String(defaultInfraName),
 	}
 	if modifyFn != nil {
 		modifyFn(mock)
@@ -188,6 +189,10 @@ func buildSubnet(vpcID, subnetId, azId, cidrBlock string) *ec2.Subnet {
 				Key:   aws.String(defaultAWSPrivateSubnetTagKey),
 				Value: aws.String("1"),
 			},
+			{
+				Key:   aws.String(getOSDClusterTagKey(defaultInfraName)),
+				Value: aws.String(clusterOwnedTagValue),
+			},
 		},
 	}
 }
@@ -204,7 +209,7 @@ func buildBundledSubnets() []*ec2.Subnet {
 	}
 }
 
-func buildValidBundleSubnets() []*ec2.Subnet {
+func buildValidRHMIBundleSubnets() []*ec2.Subnet {
 	return []*ec2.Subnet{
 		{
 			SubnetId:         aws.String("test-id"),
@@ -214,6 +219,10 @@ func buildValidBundleSubnets() []*ec2.Subnet {
 				{
 					Key:   aws.String(defaultRHMISubnetTag),
 					Value: aws.String("test"),
+				},
+				{
+					Key:   aws.String(getOSDClusterTagKey(defaultInfraName)),
+					Value: aws.String(clusterOwnedTagValue),
 				},
 				{
 					Key:   aws.String(defaultAWSPrivateSubnetTagKey),
@@ -224,7 +233,27 @@ func buildValidBundleSubnets() []*ec2.Subnet {
 	}
 }
 
-func buildMultipleValidBundleSubnets() []*ec2.Subnet {
+func buildValidBundleSubnets() []*ec2.Subnet {
+	return []*ec2.Subnet{
+		{
+			SubnetId:         aws.String("test-id"),
+			VpcId:            aws.String(defaultVpcId),
+			AvailabilityZone: aws.String("test"),
+			Tags: []*ec2.Tag{
+				{
+					Key:   aws.String(getOSDClusterTagKey(defaultInfraName)),
+					Value: aws.String(clusterOwnedTagValue),
+				},
+				{
+					Key:   aws.String(defaultAWSPrivateSubnetTagKey),
+					Value: aws.String("1"),
+				},
+			},
+		},
+	}
+}
+
+func buildMultipleValidRHMIBundleSubnets() []*ec2.Subnet {
 	return []*ec2.Subnet{
 		{
 			SubnetId:         aws.String("test-id"),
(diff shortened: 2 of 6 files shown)


pkg/providers/aws/cluster_vpc.go
```diff
@@ -457,37 +457,53 @@ func getSubnets(ec2Svc ec2iface.EC2API) ([]*ec2.Subnet, error) {
 	return subs, nil
 }
 
+func getVPCIDByClusterSubnets(ec2Svc ec2iface.EC2API, clusterID string) (*string, error) {
```
ciaranRoche, line 460: Do we need to return a pointer to a string here? I might clean it up a little bit to avoid returning `aws.String("")`

pkg/providers/aws/cluster_vpc.go
```diff
@@ -457,37 +457,53 @@ func getSubnets(ec2Svc ec2iface.EC2API) ([]*ec2.Subnet, error) {
...
+	var subs []*ec2.Subnet
+	listOutput, err := ec2Svc.DescribeSubnets(&ec2.DescribeSubnetsInput{})
+	if err != nil {
+		return aws.String(""), err
+	}
+	subs = listOutput.Subnets
+	for _, sub := range subs {
+		for _, tag := range sub.Tags {
+			if (*tag.Value == "owned" || *tag.Value == "shared") &&
```
ciaranRoche, line 469: Probably worth adding a nil check to the tag object since we are dereferencing them.

pkg/providers/aws/cluster_vpc.go
```diff
@@ -457,37 +457,53 @@ func getSubnets(ec2Svc ec2iface.EC2API) ([]*ec2.Subnet, error) {
...
+				*tag.Key == fmt.Sprintf("%s%s", clusterOwnedTagKeyPrefix, clusterID) {
```
ciaranRoche, line 470: `fmt.Sprintf("%s%s", clusterOwnedTagKeyPrefix, clusterID)` is used in multiple places, might be worth creating a little helper func here, just in the future if this tag is to ever change.

pkg/providers/aws/cluster_vpc.go
```diff
@@ -457,37 +457,53 @@ func getSubnets(ec2Svc ec2iface.EC2API) ([]*ec2.Subnet, error) {
...
+				return sub.VpcId, nil
+			}
+		}
+	}
+	return aws.String(""), errorUtil.New(fmt.Sprintf("failed to get cluster vpc, no vpc found: error getting vpc id from associated subnets: could not find cluster associated subnets with clusterID %s", clusterID))
```
ciaranRoche, line 475: NP, the error is verbose, and makes sense now as the problem is fresh in our heads. Maybe changing it to something like `failed to find cluster vpc id, no subnets found with osd cluster tag`

pkg/providers/aws/cluster_vpc.go
```diff
@@ -457,37 +457,53 @@ func getSubnets(ec2Svc ec2iface.EC2API) ([]*ec2.Subnet, error) {
...
-				foundVPC = vpc
-			}
-		}
+	vpcId, err := getVPCIDByClusterSubnets(ec2Svc, clusterID)
+	if err != nil {
+		return nil, errorUtil.Wrap(err, "error getting vpc id from associated subnets")
 	}
 
-	if foundVPC == nil {
+	vpcs, err := ec2Svc.DescribeVpcs(&ec2.DescribeVpcsInput{VpcIds: []*string{vpcId}})
```
ciaranRoche, line 494: I wonder could this be changed to something like `ec2Svc.DescribeVpcs(&ec2.DescribeVpcsInput{VpcIds: []aws.string{vpcId}})`

pkg/providers/aws/cluster_vpc.go
```diff
@@ -457,37 +457,53 @@ func getSubnets(ec2Svc ec2iface.EC2API) ([]*ec2.Subnet, error) {
 	return subs, nil
 }
 
+func getVPCIDByClusterSubnets(ec2Svc ec2iface.EC2API, clusterID string) (*string, error) {
+	var subs []*ec2.Subnet
+	listOutput, err := ec2Svc.DescribeSubnets(&ec2.DescribeSubnetsInput{})
+	if err != nil {
+		return aws.String(""), err
+	}
+	subs = listOutput.Subnets
```
ciaranRoche, line 466: NP, I dont think we need to declare a `subs` variable, can just loop through `listOutput.Subnets`

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1176,9 +1171,31 @@ func (n *NetworkProvider) reconcileStandaloneVPCSubnets(ctx context.Context, log
...
+		return nil, errorUtil.Wrap(err, "failed to get route tables")
+	}
+
+	clusterVPC, err := getClusterVpc(ctx, n.Client, n.Ec2Api, n.Logger)
+	if err != nil {
+		return nil, errorUtil.Wrap(err, "failed to get cluster vpc")
+	}
+	var foundRouteTables []*ec2.RouteTable
+	for _, routeTable := range routeTables.RouteTables {
+		if *routeTable.VpcId == *clusterVPC.VpcId {
```
ciaranRoche, line 1186: Think a nil check on `routeTable` and `clusterVPC`, better safe than sorry.

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1176,9 +1171,31 @@ func (n *NetworkProvider) reconcileStandaloneVPCSubnets(ctx context.Context, log
...
+	if len(foundRouteTables) == 0 {
+		return nil, errorUtil.New(fmt.Sprintf("could not find any route table with the associated vpc id %s", aws.StringValue(clusterVPC.VpcId)))
+	}
+	return foundRouteTables, nil
+}
+
+// getVPCRouteTableByTag will return a list of route tables based on a route table tag
 // we expect there to be route tables, if none are found we return an error
-func (n *NetworkProvider) getVPCRouteTable(routeTableTag *ec2.Tag) ([]*ec2.RouteTable, error) {
+func (n *NetworkProvider) getVPCRouteTableByTag(routeTableTag *ec2.Tag) ([]*ec2.RouteTable, error) {
```
ciaranRoche, line 1198: Do we need this function to take in a route table tag now, maybe change to `getCRORouteTable() ([]*ec2.RouteTable, error)` considering we now have `getClusterRouteTables`

pkg/providers/aws/cluster_vpc.go
makslion: hey, Ciran, Is tthis what you meant? `vpcs, err := ec2Svc.DescribeVpcs(&ec2.DescribeVpcsInput{VpcIds: []*string{aws.String(*vpcId)}})
`
It feels odd :thinking:

pkg/providers/aws/cluster_vpc.go
makslion: addressed. Now returning `nil` instead of a pointer to an empty string

pkg/providers/aws/cluster_vpc.go
makslion: addressed

pkg/providers/aws/cluster_vpc.go
makslion: addressed. Created a helper  function

pkg/providers/aws/cluster_vpc.go
makslion: addressed. Rewritten error message

pkg/providers/aws/cluster_network_provider.go
makslion: addressed

pkg/providers/aws/cluster_network_provider.go
makslion: addressed. Retrieving a tag from inside the function

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1176,9 +1159,37 @@ func (n *NetworkProvider) reconcileStandaloneVPCSubnets(ctx context.Context, log
...
+	if len(foundRouteTables) == 0 {
+		return nil, errorUtil.New(fmt.Sprintf("could not find any route table with the associated vpc id %s", aws.StringValue(clusterVPC.VpcId)))
+	}
+	return foundRouteTables, nil
+}
+
+// getVPCRouteTableByTag will return a list of route tables based on a route table tag
 // we expect there to be route tables, if none are found we return an error
-func (n *NetworkProvider) getVPCRouteTable(routeTableTag *ec2.Tag) ([]*ec2.RouteTable, error) {
+func (n *NetworkProvider) getVPCRouteTableByTag(ctx context.Context) ([]*ec2.RouteTable, error) {
```
ciaranRoche, line 1186: Can we change this function name, to `getCRORouteTable`

pkg/providers/aws/cluster_vpc.go
ciaranRoche: Is there a reason why this is being returned at a pointer?

pkg/providers/aws/cluster_vpc.go
```diff
@@ -511,3 +526,7 @@ func getSecurityGroup(ec2Svc ec2iface.EC2API, secName string) (*ec2.SecurityGrou
 
 	return foundSecGroup, nil
 }
+
+func getPrefixedClusterID(clusterID string) string{
```
ciaranRoche, line 530: `getOSDClusterTagKey` might be a more descriptive name for this function. Its confusing to read in places that its called in conditionals.

pkg/providers/aws/cluster_vpc.go
makslion: addressed. Pointers in no more

pkg/providers/aws/cluster_network_provider.go
makslion: renamed

pkg/providers/aws/cluster_vpc.go
makslion: renamed (never were able to give a good name to anything in my code :disappointed: )

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1176,9 +1159,37 @@ func (n *NetworkProvider) reconcileStandaloneVPCSubnets(ctx context.Context, log
...
+	if len(foundRouteTables) == 0 {
+		return nil, errorUtil.New(fmt.Sprintf("could not find any route table with the associated vpc id %s", aws.StringValue(clusterVPC.VpcId)))
+	}
+	return foundRouteTables, nil
+}
+
+// getCRORouteTable will return a list of route tables based on a route table tag
 // we expect there to be route tables, if none are found we return an error
-func (n *NetworkProvider) getVPCRouteTable(routeTableTag *ec2.Tag) ([]*ec2.RouteTable, error) {
+func (n *NetworkProvider) getCRORouteTable(ctx context.Context) ([]*ec2.RouteTable, error) {
```
ciaranRoche, line 1186: NP, change to `getCRORouteTables` as it returns a list of tables

makslion: renamed

pkg/providers/aws/cluster_network_provider_test.go
```diff
@@ -412,6 +443,12 @@ func buildValidIpNet(CIDR string) *net.IPNet {
 	return ip
 }
 
+func getEmptyDescribeSubnetFn() func(input *ec2.DescribeSubnetsInput) (*ec2.DescribeSubnetsOutput, error) {
```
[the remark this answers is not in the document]
makslion: removed a helper function and the `mockClient` now will always have an empty `describeSubnets` function

pkg/providers/aws/cluster_network_provider_test.go
```diff
@@ -239,18 +244,40 @@ func buildMultipleValidBundleSubnets() []*ec2.Subnet {
...
-			VpcId:            aws.String("testID"),
+			VpcId:            aws.String(defaultVpcId),
 			AvailabilityZone: aws.String("test"),
 			Tags: []*ec2.Tag{
 				{
 					Key:   aws.String(defaultRHMISubnetTag),
 					Value: aws.String("test"),
 				},
+				{
+					Key:   aws.String(getOSDClusterTagKey(defaultInfraName)),
```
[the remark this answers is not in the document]
makslion: I need to check on it. The RHMI subnet tag here is "integreatly.org/clusterID" with a "clusterID" and OSD is "kubernetes.io/cluster/test" with "owned\shared". There is one test that checks for the values in tags to match the cluster ID. Every other test is happy to proceed with only OSD tag but the `TestNetworkProvider_IsEnabled/verify isEnabled is false, a single bundle subnet is found in cluster vpc" needs booth. That thing was added a long time ago. For now I've made booth subnets to have identical tags and will report on results of "checking out"

makslion:
/retest
(suspecting aws outage to be the cause)

pkg/providers/aws/cluster_network_provider_test.go
```diff
@@ -599,6 +681,11 @@ func TestNetworkProvider_CreateNetwork(t *testing.T) {
 				Ec2Api: buildMockEc2Client(func(ec2Client *mockEc2Client) {
 					ec2Client.vpcs = buildValidClusterVPC(defaultNonOverlappingCidr)
 					ec2Client.vpc = buildValidStandaloneVPC(validCIDRTwentySix)
+					ec2Client.describeSubnetsFn = func(input *ec2.DescribeSubnetsInput) (*ec2.DescribeSubnetsOutput, error) {
+						return &ec2.DescribeSubnetsOutput{
+							Subnets: buildValidBundleSubnets(),
```
ciaranRoche, line 686: Same as above

pkg/providers/aws/cluster_network_provider_test.go
```diff
@@ -675,13 +762,19 @@ func TestNetworkProvider_CreateNetwork(t *testing.T) {
...
 					ec2Client.describeRouteTablesFn = func(input *ec2.DescribeRouteTablesInput) (*ec2.DescribeRouteTablesOutput, error) {
 						return &ec2.DescribeRouteTablesOutput{
 							RouteTables: []*ec2.RouteTable{
 								buildMockEc2RouteTable(nil),
 							},
 						}, nil
 					}
+					ec2Client.describeSubnetsFn = func(input *ec2.DescribeSubnetsInput) (*ec2.DescribeSubnetsOutput, error) {
+						return &ec2.DescribeSubnetsOutput{
+							Subnets: buildValidBundleSubnets(),
```
ciaranRoche, line 775: Same as above

pkg/providers/aws/cluster_network_provider_test.go
```diff
@@ -704,6 +797,11 @@ func TestNetworkProvider_CreateNetwork(t *testing.T) {
 					ec2Client.subnets = buildStandaloneVPCAssociatedSubnets(defaultValidSubnetMaskOneA, defaultValidSubnetMaskOneB)
 					ec2Client.azs = buildSortedStandaloneAZs()
 					ec2Client.firstSubnet = buildSubnet(defaultStandaloneVpcId, defaultSubnetIdOne, defaultAzIdOne, defaultValidSubnetMaskOneA)
+					ec2Client.describeSubnetsFn = func(input *ec2.DescribeSubnetsInput) (*ec2.DescribeSubnetsOutput, error) {
+						return &ec2.DescribeSubnetsOutput{
+							Subnets: buildValidBundleSubnets(),
```
ciaranRoche, line 802: I think this might be the same, as we are testing a non tagged standalone vpc.

pkg/providers/aws/cluster_network_provider_test.go
```diff
@@ -734,6 +832,11 @@ func TestNetworkProvider_CreateNetwork(t *testing.T) {
 					ec2Client.deleteVpcFn = func(input *ec2.DeleteVpcInput) (*ec2.DeleteVpcOutput, error) {
 						return nil, errorUtil.New("can't delete VPC, it does not exists")
 					}
+					ec2Client.describeSubnetsFn = func(input *ec2.DescribeSubnetsInput) (*ec2.DescribeSubnetsOutput, error) {
+						return &ec2.DescribeSubnetsOutput{
+							Subnets: buildValidBundleSubnets(),
```
ciaranRoche, line 837: same as above

pkg/providers/aws/cluster_network_provider_test.go
```diff
@@ -1262,6 +1376,11 @@ func TestNetworkProvider_ReconcileNetworkProviderConfig(t *testing.T) {
 							}),
 						}}, nil
 					}
+					ec2Client.describeSubnetsFn = func(input *ec2.DescribeSubnetsInput) (*ec2.DescribeSubnetsOutput, error) {
+						return &ec2.DescribeSubnetsOutput{
+							Subnets: buildValidBundleSubnets(),
```
ciaranRoche, line 1381: this too, as we are trying to verify the cidr block, we can not be using a valid bundled subnet

ciaranRoche (commented):
I think it would be good to verify the `buildValidBundleSubnets` and ensure the tests are correctly testing what is expected.

pkg/providers/aws/cluster_network_provider_test.go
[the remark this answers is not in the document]
makslion: We would need those tags to identify the subnet with a "new" approach :thinking:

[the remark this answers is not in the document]
ciaranRoche: Yes we need those tags to identify the subnets in a bundled VPC. A bundled VPC is legacy code, where CRO used to bundle the subnets into the existing cluster VPC. 

Now we have added the tags here to this builder function, and used it for tests which we want to test for a standalone VPC (the new and currently flow of creating a CRO VPC and peering it) Most of these tests are only verifying the existence of an error, and not an output, so I suspect these tests are validating the old legacy flow as they are checking bundled subnets.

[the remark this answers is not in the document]
makslion: hmm, I've removed the `defaultRHMISubnetTag` from bundled subnets and left it in `buildMultipleValidRHMIBundleSubnets` (previosly was buildMultipleValidBundleSubnets) and in a new helper `buildValidRHMIBundleSubnets`. They are used in one test case each. But those two lads still need the OSD tag in order to extract VPC :thinking: For some reason I have a feeling like [this](https://github.com/laurafitzgerald/cloud-resource-operator/blob/MGDAPI-2906/pkg/providers/aws/cluster_network_provider.go#L691-L704) part of code should be removed if it is here only to support legacy (RHMI that will not have this new version of CRO as far as I know) - we not getting a full list of VPCs and iterating through them to find _the chosen one_ :thinking: what do you think?

pkg/providers/aws/cluster_network_provider_test.go
```diff
@@ -560,9 +630,16 @@ func TestNetworkProvider_CreateNetwork(t *testing.T) {
...
-				Client:         fake.NewFakeClientWithScheme(scheme, buildTestInfra()),
-				RdsApi:         &mockRdsClient{},
-				Ec2Api:         &mockEc2Client{vpcs: buildValidClusterVPC(validCIDREighteen)},
+				Client: fake.NewFakeClientWithScheme(scheme, buildTestInfra()),
+				RdsApi: &mockRdsClient{},
+				Ec2Api: &mockEc2Client{
+					vpcs: buildValidClusterVPC(validCIDREighteen),
+					describeSubnetsFn: func(input *ec2.DescribeSubnetsInput) (*ec2.DescribeSubnetsOutput, error) {
+						return &ec2.DescribeSubnetsOutput{
+							Subnets: buildValidBundleSubnets(),
```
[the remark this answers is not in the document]
makslion: replaced with standalone

pkg/providers/aws/cluster_network_provider_test.go
```diff
@@ -580,6 +657,11 @@ func TestNetworkProvider_CreateNetwork(t *testing.T) {
 				Ec2Api: buildMockEc2Client(func(ec2Client *mockEc2Client) {
 					ec2Client.vpcs = buildValidClusterVPC(defaultNonOverlappingCidr)
 					ec2Client.vpc = buildValidStandaloneVPC(validCIDRSixteen)
+					ec2Client.describeSubnetsFn = func(input *ec2.DescribeSubnetsInput) (*ec2.DescribeSubnetsOutput, error) {
+						return &ec2.DescribeSubnetsOutput{
+							Subnets: buildValidBundleSubnets(),
```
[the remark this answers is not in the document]
makslion: replaced with standalone

pkg/providers/aws/cluster_network_provider_test.go
makslion: replaced with standalone

pkg/providers/aws/cluster_network_provider_test.go
makslion: replaced with standalone

pkg/providers/aws/cluster_network_provider_test.go
makslion: replaced with standalone

pkg/providers/aws/cluster_network_provider_test.go
makslion: replaced with standalone

pkg/providers/aws/cluster_network_provider_test.go
makslion: replaced with standalone

makslion:
/retest
I swear it worked on my PC (c) :laughing:

pkg/providers/aws/cluster_network_provider_test.go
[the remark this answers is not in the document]
ciaranRoche: I would lean against removing the code due to failing tests. It was left there to maintain backwards compatibility if needed, and I would be cautious about removing it. 

It looks like the tests are now passing though.

ciaranRoche:
@makslion can you squash all those commits into one? 

Code changes look good. Can you get someone from the team to verify/review it before merging?

ciaranRoche:
@makslion what work is left to do?

laurafitzgerald:
Hi @ciaranRoche as part of supporting installations on byovpc we had an inquiry about not using the standalone VPC but rather using the older bundled approach. We're in the process of investigating options to allow this configuration. 

There are two options.
1. Introduce a Network cr where a boolean flag can be exposed along with some other existing values like the cidr range.
2. Introduce a config map titled `bundle-vpc` or something similar.

1 requires a lot more work than 2 but puts us in a slightly better position to handle multiple products depending on cro in one custer or indeed multiple OSDs inside one byovpc.

2. comes with the benefit of being a pretty small change so we could ship it quickly.

We have _very early_ WIP pr for 1. -> https://github.com/laurafitzgerald/cloud-resource-operator/pull/1 as part of the investigation of what route to go but it's making me think we might be better going with 2 for now.

ciaranRoche:
@laurafitzgerald Is there a jira or a design doc for this change or allow resources be provisioned in the same cluster VPC? 

I am also unsure how supporting CRO provision resources in the same VPC blocks this change from being merged? This change fixes for the most part, the issue at hand and allows CRO work for addons in a BYO-VPC environment.

laurafitzgerald:
Hi @ciaranRoche agreed, this change could be merged on it's own as long as we have tested the cases listed in the verification steps. I'll verify that now. 👍

laurafitzgerald:
Verified this yesterday as per the steps in the description in both cases byovpc and byoc.
/approve
/lgtm

laurafitzgerald:
@ciaranRoche if you are happy that code review is addressed, could you add lgtm here. 👍

ciaranRoche:
Yeah sure @laurafitzgerald , @makslion can you squash those commits and I will add the label.

makslion:
@ciaranRoche Squashed. Now it is one and chunky lad of a commit.

ciaranRoche:
@makslion it's now one logical commit, unsure how you guys work with commits/PR's these days, but seems to be very opposed to squashing commits :thinking: 

/lgtm based on Laura's verification

makslion:
I guess it just me (too fearful to force push to other people branches) :thinking: we have a bot that squashes everything tho 
Thanks for pointing that out anyway

laurafitzgerald:
Yes, I had suggested squashing into one commit. We could have done separate logical commits so it's my mistake. The only requirement we have atm is that we have the JIRA number in the commit which is satisfied. 
@ciaranRoche  the `/lgtm` needs to be added on it's own line in a comment. 👍

PR #249, opened 2021-12-17
MGDAPI-3203 - emit metric around vpc limits hit

adam-cattermole:
## Overview

Jira: [MGDAPI-3203](https://issues.redhat.com/browse/MGDAPI-3203)

## Verification

Against a CCS OpenShift cluster which has no space for more VPC's:
- Clone this branch
- Run `make cluster/prepare`
- Run `make run`
- See metric emitted `cro_vpc_action{action='create', status='failed', error='VpcLimitExceeded'} 1`

## Checklist

### DIFF
--- a/pkg/providers/aws/cluster_network_provider_test.go
+++ b/pkg/providers/aws/cluster_network_provider_test.go
@@ -1062,6 +1062,56 @@ func TestNetworkProvider_CreateNetwork(t *testing.T) {
 			wantErr: true,
 		},
 		{
+			name: "verify ec2 VpcLimitExceeded returns an error",
+			fields: fields{
+				Client: fake.NewFakeClientWithScheme(scheme, buildTestInfra()),
+				RdsApi: buildMockRdsClient(nil),
+				Ec2Api: buildMockEc2Client(func(ec2Client *mockEc2Client) {
+					ec2Client.vpcs = buildValidClusterVPC(defaultNonOverlappingCidr)
+					ec2Client.describeSubnetsFn = func(input *ec2.DescribeSubnetsInput) (*ec2.DescribeSubnetsOutput, error) {
+						return &ec2.DescribeSubnetsOutput{
+							Subnets: buildStandaloneSubnets(),
+						}, nil
+					}
+					ec2Client.createVpcFn = func(input *ec2.CreateVpcInput) (*ec2.CreateVpcOutput, error) {
+						return &ec2.CreateVpcOutput{}, awserr.New("VpcLimitExceeded", "The maximum number of VPCs has been reached.", nil)
+					}
+				}),
+				ElasticacheApi: buildMockElasticacheClient(nil),
+				Logger:         logrus.NewEntry(logrus.StandardLogger()),
+			},
+			args: args{
+				ctx:  context.TODO(),
+				CIDR: buildValidCIDR(validCIDRSixteen),
+			},
+			wantErr: true,
+		},
+		{
+			name: "verify ec2 InvalidVpcRange returns an error",
+			fields: fields{
+				Client: fake.NewFakeClientWithScheme(scheme, buildTestInfra()),
+				RdsApi: buildMockRdsClient(nil),
+				Ec2Api: buildMockEc2Client(func(ec2Client *mockEc2Client) {
+					ec2Client.vpcs = buildValidClusterVPC(defaultNonOverlappingCidr)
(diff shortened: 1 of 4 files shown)


pkg/providers/aws/cluster_network_provider_test.go
```diff
@@ -1061,6 +1061,29 @@ func TestNetworkProvider_CreateNetwork(t *testing.T) {
...
+				}),
+				ElasticacheApi: buildMockElasticacheClient(nil),
+				Logger:         logrus.NewEntry(logrus.StandardLogger()),
+			},
+			args: args{
+				ctx:  context.TODO(),
+				CIDR: buildValidCIDR(validCIDRSixteen),
+			},
+			wantErr: true,
+		},
```
makslion, line 1086: Hm, I remember that it is nearly impossible to reach the `InvalidVpcRange` error code here, but still - do we want to have a unit test for it? I mean CRO is famous for the unit test coverage :thinking:

makslion: Hm, when running that unit test I got him passing, but the error I've got was `failed to get cluster vpc`. Shouldn't it be `VpcLimitExceeded`? :thinking: 
It had an issue with `failed to get cluster vpc id, no vpc found with osd cluster tag: could not find cluster associated subnets with clusterID test` and I remember messing around this [here](https://github.com/integr8ly/cloud-resource-operator/pull/244) and, probably, the way I was fixing unit tests will work as well :thinking:

adam-cattermole: Good spot, I should've been creating the subnets in the test. I've updated in latest commit

adam-cattermole: >Hm, I remember that it is nearly impossible to reach the InvalidVpcRange error code here, but still - do we want to have a unit test for it? I mean CRO is famous for the unit test coverage thinking

I could add an additional test for the `InvalidVpcRange` error also - as you say though I don't think that AWS error is reachable as it should be caught further up

makslion: Hmm, that's a tricky one. Right now this error is unreachable, yes, but we still could care about it because in the future it might change (the catch of this error will be removed and the person will see that we catch it here) or something mysterious might happen and previous catch won't work. Personally, for me, it is better to be safe than sorry. Now the philosophical one: if we decided to have a piece of code, even it is unreachable now, and could have a test for it - shouldn't we? On one hand, we have potentially wasted development time and already way too chunky test file, on the other hand, it won't hurt to have en extra coverage? :thinking:

adam-cattermole: Sure, I agree. I'll add the test as a fall through in case the other check is removed in favour of the AWS errors.

pkg/providers/aws/cluster_network_provider_test.go
```diff
@@ -1061,6 +1061,56 @@ func TestNetworkProvider_CreateNetwork(t *testing.T) {
...
+				RdsApi: buildMockRdsClient(nil),
+				Ec2Api: buildMockEc2Client(func(ec2Client *mockEc2Client) {
+					ec2Client.vpcs = buildValidClusterVPC(defaultNonOverlappingCidr)
+					ec2Client.describeSubnetsFn = func(input *ec2.DescribeSubnetsInput) (*ec2.DescribeSubnetsOutput, error) {
+						return &ec2.DescribeSubnetsOutput{
+							Subnets: buildStandaloneSubnets(),
+						}, nil
+					}
+					ec2Client.createVpcFn = func(input *ec2.CreateVpcInput) (*ec2.CreateVpcOutput, error) {
+						return &ec2.CreateVpcOutput{}, awserr.New("InvalidVpcRange", "The maximum number of VPCs has been reached.", nil)
```
makslion, line 1102: Shouldn't the message match the one we receive from [aws](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/errors-overview.html#api-error-codes-table-client)? :thinking: (`The specified CIDR block range is not valid. The block range must be between a /28 netmask and /16 netmask`)

adam-cattermole: Woops

adam-cattermole: Fixed

makslion:
Verified that the metric is being emitted 
Screenshot from 2022-01-10 11-16-26
Great work!

/lgtm

PR #263, opened 2022-01-10
make isLastResource namespaced scoped so multiple cro operator can uninstall

austincunningham:
## Overview

Jira: [MGDAPI-3024](https://issues.redhat.com/browse/MGDAPI-3024) 

## Verification

- Clone this branch
- Run the following commands to provision redis and posgres in the `cloud-resources-operator` namespace
```bash
make cluster/prepare
make cluster/seed/managed/postgres
make run
```
- We need to run both operators at the same time locally to start the second instance of the operator
- Edit the port number on to 8384 on line 61 of main.go https://github.com/integr8ly/cloud-resource-operator/blob/master/main.go#L61
- Edit the namespace env var to `cloud-resources-operator2` in the makefile https://github.com/integr8ly/cloud-resource-operator/blob/master/Makefile#L6
- In a new terminal Run the following to provision redis and postgres in `cloud-resources-operator2` namespace
```bash
make cluster/prepare
make cluster/seed/managed/postgres
make run
```
- Wait for the redis and posgres to be installed
- Delete the Redis and Posgres CR in the `cloud-resources-operator2` namespace
- Confirm that the resources are removed and there are no errors around deleting network resources 
- Delete the Redis and Posgres CR in the `cloud-resources-operator` namespace
- Confirm that the resources are removed and there are no errors around deleting network resources 
 
Index for rhoam upgrade test 
quay.io/austincunningham/integreatly-index:v1.15.2
quay.io/austincunningham/integreatly-index:v1.16.0

## Checklist

### DIFF
--- a/pkg/providers/aws/provider_postgres.go
+++ b/pkg/providers/aws/provider_postgres.go
@@ -9,8 +9,6 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws/awserr"
-	"github.com/integr8ly/cloud-resource-operator/internal/k8sutil"
-
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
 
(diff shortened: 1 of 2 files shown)


pkg/providers/aws/provider_postgres.go
```diff
@@ -671,7 +671,7 @@ func (p *PostgresProvider) getRDSConfig(ctx context.Context, r *v1alpha1.Postgre
 
 func (p *PostgresProvider) isLastResource(ctx context.Context, namespace string) (bool, error) {
 	listOptions := client.ListOptions{
-		Namespace: namespace,
+		Namespace: "",
```
KevFan, line 674: I think we can remove the namespace variable if it's unused elsewhere :thinking:

austincunningham: :+1:

KevFan:
Verified resources are removed and there are no errors around deleting network resources for when both operators :+1: 

/lgtm

PR #318, opened 2022-03-02
[MGDAPI-1716] fix: address feedback from #216

cecobask:
## Overview
Jira: https://issues.redhat.com/browse/MGDAPI-1716
Finished the work from the original PR (#216). It involved addressing feedback and adding unit tests.

## Verification
- Run the unit tests with coverage for the aws provider package (`pkg/providers/aws`). It can be done by right clicking the package and selecting `Run with coverage`.

### DIFF
--- a/controllers/postgressnapshot/postgressnapshot_controller.go
+++ b/controllers/postgressnapshot/postgressnapshot_controller.go
@@ -51,11 +51,10 @@ type PostgresSnapshotReconciler struct {
 	// This client, initialized using mgr.Client() above, is a split client
 	// that reads objects from the cache and writes to the apiserver
 	k8sclient.Client
-	scheme            *runtime.Scheme
-	logger            *logrus.Entry
-	provider          providers.PostgresSnapshotProvider
-	ConfigManager     croAws.ConfigManager
-	CredentialManager croAws.CredentialManager
+	scheme        *runtime.Scheme
+	logger        *logrus.Entry
+	provider      providers.PostgresSnapshotProvider
+	ConfigManager croAws.ConfigManager
 }
 
 // New returns a new reconcile.Reconciler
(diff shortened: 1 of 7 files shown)


pkg/providers/aws/credentials_sts_manager.go
```diff
@@ -35,12 +35,9 @@ func NewSTSCredentialManager(client client.Client) *STSCredentialManager {
...
 //ReconcileProviderCredentials Ensure the credentials the AWS provider requires are available
-func (m *STSCredentialManager) ReconcileProviderCredentials(ctx context.Context, ns string) (*Credentials, error) {
-	secret, err := getSTSCredentialsSecret(ctx, m.Client, m.OperatorNamespace)
-	if err != nil {
-		return nil, errorUtil.Wrapf(err, "failed to get aws STS credentials secret %s", defaultSTSCredentialSecretName)
-	}
-
+func (m *STSCredentialManager) ReconcileProviderCredentials(ctx context.Context, _ string) (*Credentials, error) {
+	// ignoring the error here as this is validated when calling NewCredentialManager()
+	secret, _ := getSTSCredentialsSecret(ctx, m.Client, m.OperatorNamespace)
```
KevFan, line 40: I think it might be better to still keep the error check as while unlikely, it can hit a nil pointer due to the error being ignored :thinking: (e.g. secret was there during NewCredentialManager, but was unable to get the secret during this function for some reason )

KevFan (commented):
Looks good :+1: 

Just have a minor comment

KevFan:
Verified :+1: 

/lgtm

controllers/postgressnapshot/postgressnapshot_controller.go
```diff
@@ -51,11 +51,10 @@ type PostgresSnapshotReconciler struct {
 	// This client, initialized using mgr.Client() above, is a split client
 	// that reads objects from the cache and writes to the apiserver
 	k8sclient.Client
-	scheme            *runtime.Scheme
-	logger            *logrus.Entry
-	provider          providers.PostgresSnapshotProvider
-	ConfigManager     croAws.ConfigManager
-	CredentialManager croAws.CredentialManager
```
austincunningham, line 58: May be missing from this PR but why are we removing the `CredentialManager` here and in the other controllers ? Have we tested this on a non STS cluster ?

cecobask: @austincunningham It was removed as a result from feedback on the original PR by Oleg. It's not being used at all in `postgressnapshot` and `redissnapshot` controllers. Here's the link to the comment: https://github.com/integr8ly/cloud-resource-operator/pull/216#pullrequestreview-680520734

austincunningham: Thanks @cecobask yep that clears it up 
no credentials calls in the controller , calls for credentials are done at the provider level
/lgtm
/approve

PR #327, opened 2022-03-08
MGDAPI-1726 - STS tagging of CRO resources

adam-cattermole:
## Overview

Jira: [MGDAPI-1726](https://issues.redhat.com/browse/MGDAPI-1726)

## Changes
- Default tag added to resources (`{key: "red-hat-managed", value: "true"}`)
- When STS is enabled tags are added to resources at time of creation and the reconcile of tags is disabled
- Refactored tagging
    - Function in tags.go provides a default that postgres/redis/blobstorage types extend
    - Default function in networks that is used to create tags for each network resource
    - Custom function for subnets as the tagging differs from other resources
- Added some unit tests for ensuring that the tags generated are expected for VPC's and subnets (however more could be added)

I have left the commits separate for now to make it a little easier to read but can squash before it's merged.

## Verification

This needs testing using a set of restricted credentials (as a result of [MGDAPI-1719](https://issues.redhat.com/browse/MGDAPI-1719)):

On an STS cluster:
- Configure a role that has the minimum permissions to modify/create/manage resources with the tag `{key: "red-hat-managed", value: "true"}`
- Check that `postgres/redis/blobstorage` instances can be created and deleted successfully

On a regular CCS cluster:
- Verify upgrade to ensure that tagging is unaffected (and that `{key: "red-hat-managed", value: "true"}` is reconciled onto resources as expected)

## Checklist


cecobask (approved):
Verified that the following resources have successfully been tagged with `red-hat-managed:true` when running CRO against an STS cluster:
- Vpc
- VpcPeeringConnection
- SecurityGroup
- RouteTable
- Subnet
- SubnetGroups
- Bucket
- DBInstance
- ReplicationGroups
- Snapshots

/lgtm

PR #331, opened 2022-03-10
MGDAPI-1725 - test: unit test for local sts

KevFan:
## Overview

Jira: https://issues.redhat.com/browse/MGDAPI-1725

Fixes tests failing locally due to not running in a pod, and added unit test for locally running STS 

## Verification

- Clone this branch
- Run `make test/unit`
- Ensure unit tests passes
  * Note: `verify_default_cidr_block_and_no_error_on_empty_cidr_block` is flaky and was pre-exisitng

## Checklist
- [ ] This PR includes a change to an instance type, I have used script `hack/<provider>/supported_types.sh` and attached the result below

PR #333, opened 2022-03-11
[MGDAPI-3609] fix: prevent nil pointer when creating blob storage

cecobask:
## Overview

Jira: https://issues.redhat.com/browse/MGDAPI-3609

## Verification

- Verify all unit tests are passing
- There is an existing unit test failure when ran locally in `TestCreateSessionFromStrategy`, but it's being addressed here: https://github.com/integr8ly/cloud-resource-operator/pull/331

## Checklist


scripts/sts/create-rhoam-policy.sh
```diff
@@ -1,53 +0,0 @@
...
-      "Condition": {}
-    }
-  ]
-}
-EOM
-aws iam create-role --role-name $ROLE_NAME --assume-role-policy-document "file://$ROLE_NAME.json" || true
-# TODO - attach policy with only the required permissions by CRO
-aws iam attach-role-policy --role-name $ROLE_NAME --policy-arn arn:aws:iam::aws:policy/AdministratorAccess || true
-
-sed "s|ROLE_ARN|$(get_role_arn)|g" scripts/sts/sts-secret.yaml | oc apply -f - -n $NAMESPACE
```
KevFan, line 53: While this has been moved to https://github.com/integr8ly/delorean/pull/254/files, I think parts around applying the secret needs to be kept :thinking: I think we can later remove this / make this change once the script has been finalized and merged in https://github.com/integr8ly/delorean/pull/254/files :thinking: What do you think ?

scripts/sts/sts-secret.yaml
```diff
@@ -1,8 +0,0 @@
-apiVersion: v1
```
KevFan, line 1: Think we should keep this to aid local development as STS support is technically RHOAM agnostic, and I don't think it's applied as part of https://github.com/integr8ly/delorean/pull/254/files :thinking:

KevFan (commented):
Looks good, just have some minor comments :+1:

KevFan:
Verified blobstorage provision and deletion on an STS cluster does not cause a nil pointer :+1: 

/lgtm

PR #338, opened 2022-03-17
MGDAPI-3616 Added check to confirm that standalone VPC exists before deleting it

carlkyrillos:
## Overview

Jira: https://issues.redhat.com/browse/MGDAPI-3616

This PR adds a check to the IsEnabled() function to confirm that a valid standalone VPC actually exists before returning that it does. This is required in order to fix the deletion logic for incomplete installations of the RHOAM operator where the installation got stuck because the user's VPC limit was already reached before the installation started. In this scenario, the RHOAM operator will get stuck in the Authentication stage because CRO can't create a VPC since the limit is already reached. If the user then tried to delete the RHMI CR, the uninstall would get stuck because CRO would attempt to delete a standalone VPC that doesn't exist which throws a nil pointer dereference error.

## Verification
- [Provision](https://qaprodauth.cloud.redhat.com/) a CCS cluster (requires AWS credentials) in an empty region that doesn't have any existing clusters
- Once the cluster is provisioned, log into the [AWS dashboard](https://console.aws.amazon.com/) (requires an AWS account unique to you, reach out to me with any questions) and max out the number of VPCs in the region you provisioned the cluster. The default max number of VPCs is 5 so if 3 already exist, then you only need to create 2 more
- Next target the cluster and attempt to install the RHOAM operator from the master branch of the [integreatly-operator repo](https://github.com/integr8ly/integreatly-operator)
- Verify that the installation gets stuck in the `authentication` stage because RHSSO is `awaiting cloud resources`
- Next try to delete the RHMI CR either through the GUI or by running `oc delete rhmi rhoam` from the command line
- After a few seconds/minutes the logs in the CRO operator pod should be firing a nil pointer dereference error
- Next update the yaml for the Cloud Resource Operator (Operators->Installed Operators->Cloud Resource Operator->YAML) by changing the `containerImage` and `image` fields with `quay.io/ckyrillo/cloud-resource-operator:v0.37.1` instead of `quay.io/integreatly/cloud-resource-operator:v0.37.1`
- Confirm that the RHOAM operator finishes uninstalling and that the RHMI CR is deleted

### DIFF
--- a/pkg/providers/aws/cluster_network_provider_test.go
+++ b/pkg/providers/aws/cluster_network_provider_test.go
@@ -479,7 +479,7 @@ func TestNetworkProvider_IsEnabled(t *testing.T) {
 	type fields struct {
 		Logger *logrus.Entry
 		Client client.Client
-		Ec2Svc ec2iface.EC2API
+		Ec2Api ec2iface.EC2API
 	}
 	type args struct {
 		ctx context.Context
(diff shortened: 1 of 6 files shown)


pkg/providers/aws/cluster_network_provider_test.go
```diff
@@ -1835,7 +1834,7 @@ func TestNetworkProvider_GetClusterNetworkPeering(t *testing.T) {
 			args: args{
 				ctx: context.TODO(),
 			},
-			wantErr: "failed to get standalone vpc: error getting vpcs: ec2 get vpcs error",
+			wantErr: true,
```
adam-cattermole, line 1837: Quick question regarding these changes, is this just for consistency with other tests? even though it's a big strange comparing the error string I guess these ensure that we get the desired error rather than any error

carlkyrillos: Yes I made that change for consistency and because comparing strings seemed wrong. But I see what you mean, comparing the error strings gives us a more specific test. I'll revert those changes.

pkg/providers/aws/provider_postgres.go
```diff
@@ -552,35 +552,43 @@ func (p *PostgresProvider) deleteRDSInstance(ctx context.Context, pg *v1alpha1.P
 		return croType.StatusMessage(fmt.Sprintf("deletion protection detected, modifyDBInstance() in progress, current aws rds status is %s", *foundInstance.DBInstanceStatus)), nil
 	}
 
-	// standaloneNetworkExists if no bundled resources are found in the cluster vpc
-	if standaloneNetworkExists && isLastResource {
-		logger.Info("found the last instance of types postgres and redis so deleting the standalone network")
-		networkPeering, err := networkManager.GetClusterNetworkPeering(ctx)
+	// isEnabled is true if no bundled resources are found in the cluster vpc
+	if isEnabled && isLastResource {
+		saVPC, err := getStandaloneVpc(ctx, p.Client, ec2Svc, logger)
```
laurafitzgerald, line 557: do we need to handle not found error here?

adam-cattermole: If you're meaning that the standaloneVPC isn't found then there won't be any error thrown and standaloneVPC will be nil. This is where we hit the nil pointer exception before. Instead of erroring here we assume that as we're in the delete portion of the code and the standalone isn't found (and we were supposed to deploy a standalone), either it was re-tagged so we couldn't find it, it's been deleted manually, or it was never created - in all three cases we skip cleaning up the associated networking resources and move onto removing the related CR.

After thinking a bit more about it I can't really imagine a situation where there is no VPC in the region for the describe VPC function to actually return an error - wouldn't that mean there's no standalone, no default, no cluster vpc, how would CRO even be running?

laurafitzgerald:
I see this has been verified and other review comments are addressed or don't require changes after discussion. so 👍 
/approve

PR #341, opened 2022-03-21
MGDAPI-1732 - add user-defined tags from Infrastructure CR

adam-cattermole:
## Overview

Jira: [MGDAPI-1732](https://issues.redhat.com/browse/MGDAPI-1732)

## Changes

- Add to the types to extract the tags from infrastructure CR
- Merge the default CRO tags and the user-defined tags found within the CR
    - In the case the same key exists in both the value from CRO takes precedence 

## Verification

- Create a new CCS cluster
- Check the tags on the infrastructure CR - `oc get infrastructure cluster -o json | jq`
- Deploy a managed postgres/redis/blobstorage instance
- See the default tags from the infrastructure CR are added, along with default tags to the AWS resources

## Checklist

### DIFF
--- a/pkg/providers/aws/cluster_network_provider.go
+++ b/pkg/providers/aws/cluster_network_provider.go
@@ -1809,8 +1809,8 @@ func getCloudResourceOperatorOwnerTag(ctx context.Context, client client.Client)
 	return genericTag, nil
 }
 
-func getDefaultTagSpec(ctx context.Context, client client.Client, t *tag, resourceType string) ([]*ec2.TagSpecification, error) {
-	tags, err := getDefaultNetworkTags(ctx, client, t)
+func getDefaultTagSpec(ctx context.Context, client client.Client, customTag *tag, resourceType string) ([]*ec2.TagSpecification, error) {
+	tags, err := getDefaultNetworkTags(ctx, client, customTag)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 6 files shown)


cecobask (approved):
Verified on a ROSA cluster with STS enabled. All resources created by CRO have the default tags to the infrastructure CR, along with default tags to the AWS resources.

/lgtm

adam-cattermole:
I've spotted that the function [here](https://github.com/integr8ly/cloud-resource-operator/pull/341/files#diff-6aa319203c220a223cc85fefbba474a0a64b1f8f54f26474608a4d9977cbe138R101-R107) is non-determinstic and the tags could be retrieved in any order. To ensure that the tags are ordered and the tests don't fail I've added a quick sort to the tags.

adam-cattermole:
coverage failed because [of flaky test](https://issues.redhat.com/browse/MGDAPI-3664) (with the same issue as this)

/test coverage

cecobask (approved):
/lgtm

pkg/providers/aws/tags.go
```diff
@@ -98,6 +100,33 @@ func genericToEc2Tags(tags []*tag) []*ec2.Tag {
...
 }
 
+// sort the keys to ensure tags returned in same order
+func mapToGeneric(tags map[string]string) []*tag {
+	var genTags []*tag
+	keys := make([]string, 0, len(tags))
+	for k := range tags {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
```
ciaranRoche, line 110: Why the sort? does the order matter in CRO for a list of tags? I would not expect this list to be very large.

If you don't care about the sorting I thinking both these functions could be shortened to something like : 
```
	var dupMap = make(map[string]bool)
	for _, tag := range generalTags {
		dupMap[tag.key] = true
	}

	for _, tag := range infraTags {
		if _, exists := dupMap[tag.key]; !exists {
			generalTags = append(generalTags, tag)
		}
	}
	return generalTags
```

IMO I would try and avoid converting tags to a map[string]string and then converting them back to an array of tags. 

If you want to keep the sort, I would also suggest adding a sort to the generic tag type, can use this as an example - https://github.com/integr8ly/cloud-resource-operator/blob/master/pkg/providers/aws/sort.go

pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1830,17 +1830,25 @@ func getDefaultNetworkTags(ctx context.Context, client client.Client, t *tag) ([
...
-			buildManagedTag(),
-		}, nil
-	}
-	return []*tag{
+	tags := []*tag{
 		croTag,
 		buildManagedTag(),
-	}, nil
+	}
+	if t != nil {
```
ciaranRoche, line 1837: NP, rename this to something a bit more meaningful, I know this is done all over CRO and some people like to super abbreviate variable names in GO, but it tends to not be the most readable code in the long run. So good habit to update stuff like this when you can.

pkg/providers/aws/tags.go
adam-cattermole: The sort was just to ensure the order the tags are retrieved is consistent.

Your solution is a little more elegant so I definitely prefer this over converting to map and back again - made changes to use this instead.

pkg/providers/aws/cluster_network_provider.go
adam-cattermole: Agreed - updated 2 occurrences.

ciaranRoche:
/lgtm

changes look good to me, thanks.

PR #357, opened 2022-04-04
MGDAPI-3652 - Use bundle format and file-based catalog images

adam-cattermole:
## Overview

Jira:
[MGDAPI-3652](https://issues.redhat.com/browse/MGDAPI-3652)
[MGDAPI-3326](https://issues.redhat.com/browse/MGDAPI-3326)

## Changes
- Added new bundle format as a result of moving from pkgman to bundles
- Added a new script, gen-csv.sh which updates the base YAML files to build a new bundle and indexes
- Updated script bundle-rhmi-operators.sh to create and push the bundle, and to generate a new file-based index using the bundles
- Made minor changes to the documentation to reflect the change to bundles (although I have aimed to keep the process mostly the same)

## Verification
- Building a new bundle and index image works as expected
- Upgrade scenario from old index to new file-based index is successful
- Upgrade from a skip version to latest using new file-based index is successful


Makefile
```diff
@@ -230,7 +222,7 @@ setup/moq:
 
 .PHONY: create/olm/bundle
 create/olm/bundle:
-	IMAGE_ORG=${IMAGE_ORG} IMAGE_REG=${IMAGE_REG} CHANNEL=${CHANNEL} UPGRADE=${UPGRADE} PREVIOUS_OPERATOR_VERSIONS=${PREVIOUS_OPERATOR_VERSIONS} ./scripts/bundle-rhmi-operators.sh
+	@CHANNEL=$(CHANNEL) PREV_VERSION=$(PREV_VERSION) PREVIOUS_OPERATOR_VERSIONS=$(PREVIOUS_OPERATOR_VERSIONS) ./scripts/bundle-rhmi-operators.sh
```
laurafitzgerald, line 225: @adam-cattermole small nit, could we update the name of the script to not include the acronym `rhmi`?

bundles/0.10.0/bundle.Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM scratch
```
laurafitzgerald, line 1: We've been chatting about the possibility of not maintaining old versions of the bundles. As the repo itself is versioned we technically just need one copy that we could update all of the time. The change for that doesn't technically need to be done in this pr unless it makes sense to do that. Maybe we can chat about it?

Makefile
adam-cattermole: Updated to create-olm-bundle.sh

PR #358, opened 2022-04-04
MGDAPI-3664 / sorting CIDR ranges

den-rgb:
## Overview

Jira:[ MGDAPI-3664](https://issues.redhat.com/browse/MGDAPI-3664)

#What
Sorting CIDR ranges In order to stop flakiness of [TestNetworkProvider_ReconcileNetworkProviderConfig()](https://github.com/integr8ly/cloud-resource-operator/blob/master/pkg/providers/aws/cluster_network_provider_test.go#L1319)

#Test
Run the test a couple of times cleaning test cache with `go clean -testcache` after each run and make sure test passes.

### DIFF
--- a/pkg/providers/aws/cluster_network_provider.go
+++ b/pkg/providers/aws/cluster_network_provider.go
@@ -74,6 +74,11 @@ type Network struct {
 	Subnets []*ec2.Subnet
 }
 
+type cidrList struct {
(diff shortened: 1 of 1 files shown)


Patryk-Stefanski:
code looks good and tests are passing.
/lgtm

PR #375, opened 2022-04-19
[MGDAPI-3827] Tag redis & postgres snapshots on creation

cecobask:
## Overview

Jira: https://issues.redhat.com/browse/MGDAPI-3827

## Verification

- Provision a ROSA cluster using the script in delorean: `make -f make/ocm.mk ocm/sts/cluster/create OCM_ENV=production CLUSTER_NAME=sts-test`
- Setup and wait for cluster admin credentials to become available`rosa create admin --cluster=sts-test`
- Log into the cluster `oc login $(cluster_url) --username cluster-admin --password $(password)`
- Prepare the cluster `make cluster/prepare` using the make commands from the CRO repo
- Setup the sts secret `make setup/sts OCM_ENV=production CLUSTER_NAME=sts-test`
- Run CRO `make run`
- Deploy a redis instance`make cluster/seed/managed/redis`
- Deploy a postgres instance `make cluster/seed/managed/postgres`
- Wait for successful reconciliation
- Create a redis snapshot `make cluster/seed/managed/redissnapshot`
- Create a postgres snapshot `make cluster/seed/managed/postgressnapshot`
- Wait for successful reconciliation
- Ensure the redis and postgres snapshots have been tagged with `red-hat-managed` and other tags
- Trigger deletion of the redis snapshot `oc delete redissnapshot example-redissnapshot -n cloud-resource-operator`
- Trigger deletion of the postgres snapshot `oc delete postgressnapshot example-postgressnapshot -n cloud-resource-operator`
- Wait for successful reconciliation 
- Ensure there are no redis, redissnapshot, postgres, postgressnapshot CRs left in the CRO namespace
- Delete the ROSA STS cluster using the script from delorean: `make -f make/ocm.mk ocm/sts/cluster/delete CLUSTER_NAME=sts-test`

## Checklist
- [ ] This PR includes a change to an instance type, I have used script `hack/<provider>/supported_types.sh` and attached the result below

PR #386, opened 2022-04-28
[MGDAPI-3909] include isSTSCluster struct property to NetworkProvider in CRO

cecobask:
## Overview

Jira: https://issues.redhat.com/browse/MGDAPI-3909

## Verification
This PR isn't changing any workflows, it's simply a performance/code readability change. It should be sufficient if all prow checks pass.
If someone wants to verify CRO is working as expected in STS mode on ROSA, they can follow the verification steps from this PR: https://github.com/integr8ly/cloud-resource-operator/pull/375

## Checklist

### DIFF
--- a/pkg/providers/aws/cluster_network_provider.go
+++ b/pkg/providers/aws/cluster_network_provider.go
@@ -58,7 +58,6 @@ const (
 	defaultNumberOfExpectedSubnets       = 2
 	clusterOwnedTagKeyPrefix             = "kubernetes.io/cluster/"
 	clusterOwnedTagValue                 = "owned"
-	clusterSharedTagValue                = "shared"
 	//network peering
 	tagVpcPeeringNameValue = "RHMI Cloud Resource Peering Connection"
 	// filter names for vpc peering connections
@@ -69,7 +68,7 @@ const (
 	defaultCIDRMask = "26"
 )
 
-// wrapper for ec2 vpcs, to allow for extensibility
+// Network wrapper for ec2 vpcs, to allow for extensibility
 type Network struct {
 	Vpc     *ec2.Vpc
 	Subnets []*ec2.Subnet
@@ -80,13 +79,13 @@ type cidrList struct {
 	defaultVal string
 }
 
-// used to map expected ip addresses to availability zones
+// NetworkAZSubnet used to map expected ip addresses to availability zones
 type NetworkAZSubnet struct {
 	IP net.IPNet
 	AZ *ec2.AvailabilityZone
 }
 
-// wrapper for ec2 vpc peering connections, to allow for extensibility
+// NetworkPeering wrapper for ec2 vpc peering connections, to allow for extensibility
 type NetworkPeering struct {
 	PeeringConnection *ec2.VpcPeeringConnection
 }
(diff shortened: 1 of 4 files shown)


pkg/providers/aws/cluster_network_provider.go
```diff
@@ -1862,32 +1848,8 @@ func getDefaultNetworkTags(ctx context.Context, client client.Client, customTag
...
-		return nil, errorUtil.Wrap(err, "error getting clusterID")
-	}
-
-	// a tag for identifying cluster owned vpcs
-	return &ec2.Tag{
-		Key:   aws.String(getOSDClusterTagKey(clusterID)),
-		Value: aws.String(tagValue),
-	}, nil
-}
-
```
cecobask, line 1888: Not used, cleaning up

laurafitzgerald: @cecobask this functionality was added to support byovpc. Could you point to where the cluster vpc is gotten now? If it's moved that's fine but we still need to include both `owned` and `shared` otherwise we'll remove support for byovpc.

laurafitzgerald (commented):
One question inline. Although I've just had time to do a quick eyeball this time.

pkg/providers/aws/cluster_network_provider.go
adam-cattermole: I'm unsure when the references were removed but there are no uses of these functions on current master version - or at least that's what my linter tells me.

cecobask: @adam-cattermole +1
@laurafitzgerald Not sure how byovpc works, but I believe the function was abandoned in this PR: https://github.com/integr8ly/cloud-resource-operator/pull/244/files

laurafitzgerald: The pr you linked is where the shared tag was added. 🤔  https://github.com/integr8ly/cloud-resource-operator/pull/244/files#diff-691829d67b270bfdcac4bcee0b5a0c2517c3885fa20c7d78cb27703ed6136ac5R60

laurafitzgerald: Hi @cecobask sorry for the delay. I was thinking of the shared tag here-> https://github.com/laurafitzgerald/cloud-resource-operator/blob/852d9dd0d43b2fa74a82cb431208e692f88d3087/pkg/providers/aws/cluster_vpc.go#L468 so we can remove these functions.

pkg/providers/aws/provider_redis.go
```diff
@@ -501,7 +501,7 @@ func (p *RedisProvider) DeleteRedis(ctx context.Context, r *v1alpha1.Redis) (cro
 	}
 
 	// network manager required for cleaning up network.
-	networkManager := NewNetworkManager(sess, p.Client, logger)
+	networkManager := NewNetworkManager(sess, p.Client, logger, isSTSCluster(ctx, p.Client))
```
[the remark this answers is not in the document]
cecobask: @laurafitzgerald That function was created previously and has been running without issues for everyone using the STS feature branch so far. If the function `isSTSCluster` gets invoked from a non STS cluster it will return false. I couldn't identify any possible errors down the line as all error cases seem to be handled.

PR #396, opened 2022-05-11
[MGDAPI-3942] improve code coverage on the STS feature branch

cecobask:
## Overview

Jira: https://issues.redhat.com/browse/MGDAPI-3942

## Verification

- Check that code coverage is satisfactory and that the changes make sense

## Checklist

### DIFF
--- a/pkg/providers/aws/config.go
+++ b/pkg/providers/aws/config.go
@@ -124,7 +124,7 @@ func BuildDefaultConfigMap(name, namespace string) *v1.ConfigMap {
 	}
 }
 
-// BuildSubnetGroupName builds and returns an id used for infra resources
+// BuildInfraName builds and returns an id used for infra resources
 func BuildInfraName(ctx context.Context, c client.Client, postfix string, n int) (string, error) {
 	// get cluster id
 	clusterID, err := resources.GetClusterID(ctx, c)
@@ -184,11 +184,7 @@ func CreateSessionFromStrategy(ctx context.Context, c client.Client, credentials
 	} else {
 		awsConfig.Credentials = awsCredentials.NewStaticCredentials(credentials.AccessKeyID, credentials.SecretAccessKey, "")
 	}
-
-	sess, err := session.NewSession(&awsConfig)
-	if err != nil {
-		return nil, errorUtil.Wrapf(err, "failed to create aws session from strategy, region=%s", region)
-	}
+	sess := session.Must(session.NewSession(&awsConfig))
 	return sess, nil
 }
 
--- a/pkg/providers/aws/credentials_sts_manager.go
+++ b/pkg/providers/aws/credentials_sts_manager.go
@@ -36,10 +36,8 @@ func NewSTSCredentialManager(client client.Client) *STSCredentialManager {
 
 //ReconcileProviderCredentials Ensure the credentials the AWS provider requires are available
 func (m *STSCredentialManager) ReconcileProviderCredentials(ctx context.Context, _ string) (*Credentials, error) {
-	secret, err := getSTSCredentialsSecret(ctx, m.Client, m.OperatorNamespace)
-	if err != nil {
-		return nil, errorUtil.Wrapf(err, "failed to get aws sts credentials secret %s", defaultSTSCredentialSecretName)
-	}
+	// if execution has reached this point, the secret exists - it was validated in NewCredentialManager()
+	secret, _ := getSTSCredentialsSecret(ctx, m.Client, m.OperatorNamespace)
 	credentials := &Credentials{
 		RoleArn:       string(secret.Data[defaultRoleARNKeyName]),
 		TokenFilePath: string(secret.Data[defaultTokenPathKeyName]),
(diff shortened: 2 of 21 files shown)


pkg/providers/aws/config.go
```diff
@@ -184,11 +184,7 @@ func CreateSessionFromStrategy(ctx context.Context, c client.Client, credentials
 	} else {
 		awsConfig.Credentials = awsCredentials.NewStaticCredentials(credentials.AccessKeyID, credentials.SecretAccessKey, "")
 	}
-
-	sess, err := session.NewSession(&awsConfig)
-	if err != nil {
-		return nil, errorUtil.Wrapf(err, "failed to create aws session from strategy, region=%s", region)
-	}
+	sess := session.Must(session.NewSession(&awsConfig))
```
laurafitzgerald, line 187: I'm not sure about this change. The error is handled by panicing rather than reporting an error back. I beleive this pattern makes it more difficult to debug issues.

laurafitzgerald (commented):
Overall the changes look good. Just two questions inline.

pkg/providers/aws/config.go
cecobask: https://docs.aws.amazon.com/sdk-for-go/api/aws/session/#Must

The `Must` helper function seems to be widely used in AWS SDK code examples.
I agree it may cause debugging to be slightly more difficult, but it saves us a refactor to include this error check in code coverage. This file is the only place in CRO where we're initializing new sessions, so maybe it won't be that difficult to debug. I can change that piece to handle the error in our code, if requested.

adam-cattermole (commented):
I like that you've added some tests to increase coverage, but I'm going to be honest I really dislike the changes removing error handling to increase coverage.

I understand the premise that to reach this line of code the previous function has to have ran successfully, but given that the execution of these functions are based on the response of an API call I don't see how we can guarantee that an error is not thrown. These typically are reading from the API server so if for any reason that connection fails or the API server is down for even a fraction of a second this error can still occur (even though it may be incredibly unlikely).

cecobask:
> I like that you've added some tests to increase coverage, but I'm going to be honest I really dislike the changes removing error handling to increase coverage.
> 
> I understand the premise that to reach these lines of code the previous function has to have ran successfully, but given that the execution of these functions are based on the response of an API call I don't see how we can guarantee that an error is not thrown. These typically are reading from the API server so if for any reason that connection fails or the API server is down for even a fraction of a second this error can still occur (even though it may be incredibly unlikely).
> 
> edit: unless the result of the get is cached for any reason so we skip the connection in which case it's guaranteed?

Ah, didn't think of that scenario. I like the idea of caching. The cluster id value is retrieved hundreds of times in the codebase and I think we'll see a decent performance improvement if we do it. I don't think the cluster id is currently cached anywhere. What do you think about this, @laurafitzgerald?

pkg/providers/aws/credentials_sts_manager.go
```diff
@@ -36,10 +36,8 @@ func NewSTSCredentialManager(client client.Client) *STSCredentialManager {
 
 //ReconcileProviderCredentials Ensure the credentials the AWS provider requires are available
 func (m *STSCredentialManager) ReconcileProviderCredentials(ctx context.Context, _ string) (*Credentials, error) {
-	secret, err := getSTSCredentialsSecret(ctx, m.Client, m.OperatorNamespace)
-	if err != nil {
-		return nil, errorUtil.Wrapf(err, "failed to get aws sts credentials secret %s", defaultSTSCredentialSecretName)
-	}
+	// if execution has reached this point, the secret exists - it was validated in NewCredentialManager()
+	secret, _ := getSTSCredentialsSecret(ctx, m.Client, m.OperatorNamespace)
```
KevFan, line 40: I would prefer keeping the error check. While I understand this is an edge case, technically it can happen. For example, secret was present during `NewCredentialManager()`, but was unable to get the secret for some reason / got deleted just before this function call, it would cause a nil pointer here and cause the operator to crash :thinking:

pkg/providers/aws/provider_postgres.go
```diff
@@ -281,7 +281,8 @@ func (p *PostgresProvider) reconcileRDSInstance(ctx context.Context, cr *v1alpha
 	postgresPass := string(credSec.Data[defaultPostgresPasswordKey])
 	if postgresPass == "" {
 		msg := "unable to retrieve rds password"
-		return nil, croType.StatusMessage(msg), errorUtil.Wrap(err, msg)
+		return nil, croType.StatusMessage(msg), errorUtil.Errorf(msg)
+		//return nil, croType.StatusMessage(msg), errorUtil.Wrap(err, msg)
```
KevFan, line 285: Is there a reason to keep the commented out line ? :thinking:

cecobask: No reason to keep it. I must have forgotten to delete it 😄

laurafitzgerald:
i think it's unlikely that the clusterid will change so I think we could look at caching that alright. Also I agree with `I really dislike the changes removing error handling to increase coverage.` Sorry if I missed that somewhere in the review.

laurafitzgerald:
/approve
e2e test failure looks to be infrastructure issue.

PR #414, opened 2022-05-30
check for nil AutoAppliedAfterDate on pending maintenance action

laurafitzgerald:
## Overview

Jira: [https://issues.redhat.com/browse/MGDAPI-4115](https://issues.redhat.com/browse/MGDAPI-4115)

## Verification

- Remove the added code `pmac.AutoAppliedAfterDate != nil check 
- Run the test. it will fail on the line 1282 with nil dereference
- Add the line back in a re run the test, it will pass

## Checklist


cecobask:
Tried to reproduce the following pending maintenance with a seeded postgres instance from CRO, but I couldn't: 
image

However, the unit test and the extra logic that you added seem like it will handle the error case. The change can be tested on the eng-long-lived cluster.

/lgtm

PR #440, opened 2022-06-27
[MGDAPI-4169] set default credential manager to minter and introduce metric

cecobask:
## Overview
Jira: https://issues.redhat.com/browse/MGDAPI-4169

## Verification
- Clone this branch
- Run `make cluster/prepare`
- Run `make run`
- Run `curl -s http://localhost:8383/metrics | grep "cro_sts_credentials_secret"` and see metric emitted:
```
cro_sts_credentials_secret{error="secrets \"sts-credentials\" not found",namespace="cloud-resource-operator"} 1
```

### DIFF
--- a/pkg/providers/aws/credentials.go
+++ b/pkg/providers/aws/credentials.go
@@ -150,9 +150,9 @@ func NewCredentialManager(client client.Client) (CredentialManager, error) {
 	secret, err := getSTSCredentialsSecret(context.TODO(), client, ns)
 	if err != nil {
 		if errors.IsNotFound(err) {
-			resources.SetSTSCredentialsSecretMetric(ns, err)
 			return NewCredentialMinterCredentialManager(client), nil
 		}
+		resources.SetSTSCredentialsSecretMetric(ns, err)
 		return nil, err
 	}
 	resources.ResetSTSCredentialsSecretMetric()
--- a/controllers/blobstorage/blobstorage_controller.go
+++ b/controllers/blobstorage/blobstorage_controller.go
@@ -68,11 +68,11 @@ func New(mgr manager.Manager) (*BlobStorageReconciler, error) {
 	}
 
 	logger := logrus.WithFields(logrus.Fields{"controller": "controller_blobstorage"})
-	blobStorageProvider, err := aws.NewAWSBlobStorageProvider(client, logger)
+	awsBlobStorageProvider, err := aws.NewAWSBlobStorageProvider(client, logger)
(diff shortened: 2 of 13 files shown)


bryankeane0:
/lgtm
Code looks good and metric was retrieved successfully after running cro on a cluster

pkg/providers/aws/credentials.go
```diff
@@ -144,9 +145,11 @@ type CredentialManager interface {
 func NewCredentialManager(client client.Client) CredentialManager {
 	ns, _ := k8sutil.GetOperatorNamespace()
```
laurafitzgerald, line 146: We have an unhandled error here.

pkg/providers/aws/credentials_test.go
```diff
@@ -73,8 +80,11 @@ func TestCredentialMinterManager_ReconcileProviderCredentials(t *testing.T) {
 	}
 	for _, tc := range cases {
 		t.Run(tc.name, func(t *testing.T) {
-			cm := NewCredentialManager(tc.client).(*CredentialMinterCredentialManager)
-			awsCreds, err := cm.ReconcileProviderCredentials(context.TODO(), "testNamespace")
+			cm, err := NewCredentialManager(tc.client)
+			if err != nil {
```
laurafitzgerald, line 84: The results in fatal here but not in a previous test. Any reason for the difference?

pkg/providers/aws/credentials.go
```diff
@@ -141,13 +142,24 @@ type CredentialManager interface {
...
-		return NewCredentialMinterCredentialManager(client)
+func NewCredentialManager(client client.Client) (CredentialManager, error) {
+	ns, err := k8sutil.GetOperatorNamespace()
+	if err != nil {
+		return nil, err
+	}
+	secret, err := getSTSCredentialsSecret(context.TODO(), client, ns)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			resources.SetSTSCredentialsSecretMetric(ns, err)
```
laurafitzgerald, line 153: I think this metric set could be moved outside of the isNotFound error as we are accepting that one as not an error effectively.

laurafitzgerald (changes requested):
Left a couple comments inline but I the update to the provider is looking good.

pkg/providers/aws/credentials.go
cecobask: Done 👍🏻

PR #466, opened 2022-07-21
MGDAPI-4387 - tag VPC on creation for all cluster types

adam-cattermole:
## Overview

[MGDAPI-4387](https://issues.redhat.com/browse/MGDAPI-4387)

- We no longer check whether we are in STS cluster or not and create the VPC initially with tags
- Fixed a regression where we would try to reconcile tags onto STS clusters

## Verification

- Provision a CCS cluster
- From 1.24.0
	- Create at least 2 CR's in quick succession
	- See that two VPCs are created in AWS
- Do the same from this PR
- Verify that only a single VPC is created, and tagged on creation

## Checklist

### DIFF
--- a/pkg/providers/aws/cluster_network_provider.go
+++ b/pkg/providers/aws/cluster_network_provider.go
@@ -183,13 +183,12 @@ func (n *NetworkProvider) CreateNetwork(ctx context.Context, vpcCidrBlock *net.I
 		vpcConfig := &ec2.CreateVpcInput{
 			CidrBlock: aws.String(vpcCidrBlock.String()),
 		}
-		if n.IsSTSCluster {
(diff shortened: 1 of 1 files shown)


adam-cattermole:
Looks like an issue with prow - [tests passed](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/integr8ly_cloud-resource-operator/466/pull-ci-integr8ly-cloud-resource-operator-master-e2e/1550090943766466560/artifacts/e2e/test/build-log.txt) but will re-run.

/test e2e

KevFan:
* Created a posgres and redis cr is quick succession
image
* Verified that only one VPC is created and tagged
* Verified resources are provisioned 
image

/lgtm

PR #468, opened 2022-07-22
MGDAPI-4355: Allow configuration of CRO maintenance window

kryanbeane:
## Overview
Jira: [MGDAPI-4355](https://issues.redhat.com/browse/MGDAPI-4355)
- Added function which checks for config map containing db provider names. If present then service updates are allowed to go ahead
- **TODO**: Add unit tests

## Verification
- TBD

## Checklist


adam-cattermole:
The verification of this ticket is difficult given that service updates aren't usually present.
- Provision a CCS cluster and test upgrade + install from this PR
- Test that setting `updatesAllowed: true` on a postgres/redis CR results in normal behaviour and the CR should revert back to false after a successful reconcile

pkg/providers/aws/provider_redis.go
```diff
@@ -131,6 +131,20 @@ func (p *RedisProvider) CreateRedis(ctx context.Context, r *v1alpha1.Redis) (*pr
...
+	updatesAllowed, err := resources.VerifyRedisUpdatesAllowed(ctx, p.Client, r.Namespace, r.Name)
+	if err != nil {
+		msg := "failed to verify if updates are allowed"
+		return nil, croType.StatusMessage(msg), errorUtil.Wrap(err, msg)
+	}
+
+	if updatesAllowed {
+		msg, err := p.checkAndApplyMaintenanceUpdates(r, elasticache.New(sess), ec2.New(sess), elasticacheCreateConfig, serviceUpdates)
+		// if there is a status update or error return to controller
+		if err != nil && msg != "" {
```
valerymo, line 143: Same as for Postgres - 
seems to me it should look like:
if err != nil {
return nil, statusMsg, err
}
It could be the case with NO error, but with status message, and it considered as error ... Not sure it's ok, as sutus could be also OK/valid, But if not, if status always not valid, so seems to me there is something not clear here with logics. But if we will change it  - it will require the change in checkAndApplyMaintenanceUpdates(...). We could discuss, if my notes are not so clear. Thank you

pkg/providers/aws/provider_postgres.go
```diff
@@ -224,26 +238,56 @@ func (p *PostgresProvider) ReconcilePostgres(ctx context.Context, pg *v1alpha1.P
...
+func (p *PostgresProvider) checkAndApplyMaintenanceUpdates(pg *v1alpha1.Postgres, rdsSvc rdsiface.RDSAPI, rdsCfg *rds.CreateDBInstanceInput, serviceUpdates *ServiceUpdate) (croType.StatusMessage, error) {
+	logger := p.Logger.WithField("action", "checkAndApplyMaintenanceUpdates")
+	pi, err := getRDSInstances(rdsSvc)
 	if err != nil {
-		errMsg := "failed to service update rds instance"
-		return nil, croType.StatusMessage(errMsg), err
-	}
-	if updating {
-		return nil, message, nil
+		msg := "failed to get rds replication groups"
```
valerymo, line 257: seems message text is not correct (copied from Redis), it's not replication groups like in redis (?). Maybe could be like "failed to get Postgres Instances"

pkg/providers/aws/provider_redis.go
```diff
@@ -131,6 +131,20 @@ func (p *RedisProvider) CreateRedis(ctx context.Context, r *v1alpha1.Redis) (*pr
 		return nil, croType.StatusMessage(errMsg), errorUtil.Wrap(err, errMsg)
 	}
 
+	updatesAllowed, err := resources.VerifyRedisUpdatesAllowed(ctx, p.Client, r.Namespace, r.Name)
+	if err != nil {
+		msg := "failed to verify if updates are allowed"
```
valerymo, line 136: message test, - could we add "Redis"? , like ""failed to verify if Redis updates are allowed" (check also for Postgres). Make errors Info more specific, to make debugging/investigation easier.

valerymo (commented):
Look very good, just common comment - could we make errors Info more specific, to make debugging/investigation easier. I leaved few comments, but they are minor. It's before testing, I will try test it.

pkg/providers/aws/provider_postgres.go
```diff
@@ -147,6 +147,27 @@ func (p *PostgresProvider) ReconcilePostgres(ctx context.Context, pg *v1alpha1.P
...
+	updatesAllowed, err := resources.VerifyPostgresUpdatesAllowed(ctx, p.Client, pg.Namespace, pg.Name)
+	if err != nil {
+		msg := "failed to verify if postgres updates are allowed"
+		return nil, croType.StatusMessage(msg), errorUtil.Wrap(err, msg)
+	}
+
+	if updatesAllowed {
+		msg, err := p.checkAndApplyMaintenanceUpdates(pg, rds.New(sess), rdsCfg, serviceUpdates)
+		// if there is a status update or error return to controller
+		if err != nil && msg != "" {
```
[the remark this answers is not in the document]
adam-cattermole: The controller above this expects one of 3 cases. Either there is a `PostgresInstance` returned, if so it is phase success, there is an error, in which case this is `PhaseFailed`, or there is no error no postgres in which case it is `PhaseInProgress.

Here the `msg` is only returned if the service updates are in progress, and we return to the controller as soon as possible to update the status to in progress. If we only return an error we can't tell whether the updates are in progress or if there's been a problem performing the updates. In the case where there is no message and no error we have successfully passed through this function and so no updates are pending. I agree I'm not completely sold on the handling of states via strings but I'd prefer not to overhaul the implementation of the controller.

pkg/providers/aws/provider_postgres.go
```diff
@@ -147,6 +147,27 @@ func (p *PostgresProvider) ReconcilePostgres(ctx context.Context, pg *v1alpha1.P
...
+			return nil, msg, err
+		}
+	}
+
 	// create credentials secret
 	sec := buildDefaultRDSSecret(pg)
```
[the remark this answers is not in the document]
adam-cattermole: I believe the error for this is handled on the next line, L173

pkg/providers/aws/provider_redis.go
adam-cattermole: See reasoning above.

pkg/providers/aws/provider_postgres.go
adam-cattermole: RDS is from the Amazon Relational Database Service (RDS) and is used for Postgres instances. Redis is deployed by Amazon's ElastiCache service. Even though it's a little confusing it is accurate.

pkg/providers/aws/provider_postgres.go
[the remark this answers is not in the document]
adam-cattermole: This should be an or though not an and, let me fix it in a new commit.

pkg/providers/aws/provider_redis.go
```diff
@@ -346,6 +372,75 @@ func (p *RedisProvider) createElasticacheCluster(ctx context.Context, r *v1alpha
...
+			cluster := *checkedCluster
+			if resources.SafeStringDereference(cluster.ReplicationGroupId) == *foundCache.ReplicationGroupId {
+				replicationGroupClusters = append(replicationGroupClusters, *checkedCluster)
+
+				if checkedCluster.EngineVersion != nil && r.Status.Version != *checkedCluster.EngineVersion {
+					r.Status.Version = *checkedCluster.EngineVersion
+				}
+			}
+		}
+
```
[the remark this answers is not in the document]
adam-cattermole: I think this would only be possible if the replication group existed without any cache clusters, which I'm not sure is possible. This logic has been used on L310 during provisioning of redis for the last 2 years so I think we might be safe

pkg/providers/aws/provider_redis.go
adam-cattermole: Agreed, addressed in new commit

pkg/providers/aws/provider_postgres.go
valerymo: In Redis:  "errMsg := "error getting replication groups"
In Postgres:  msg := "failed to get rds replication groups"
Formally All is good. Postgres is RDS, so we know if "rds" appear in the error message - error from Postgres. But from my pov - errors notification are intended to help people understand the issue as soon as possible. I would add Redis/Postgres in this messages (or write RDS), but ok, let's say it's minor, it's up to author and approvers. Maybe if such error happens, it will be clear from other notifications around what is about. Thank you

valerymo:
@adam-cattermole , @laurafitzgerald from my pov overall it looks ok (before test), but some  things - like TODO and TBD in Overview and Verification, and Checklist... (?). 
Maybe it's possible to add - how this task was verified by author, as if it was not verified, unfortunately seems need bring it back to Progress and check.
Regarding Coverage, it's better but still could be improved (I checked in Goland), but if Task is blocker, maybe it could be done in separate task. Thank you

adam-cattermole:
@valerymo

- PR header is not editable at this time and should be disregarded
- Every branch of `checkAndApplyMaintenanceUpdates` that could be tested conveniently has been tested. Tests were omitted for branches where there is a default timeout of 5 minutes to retrieve information from AWS as unit tests would be delayed during this timeout. I can however see if I can add a test or two for some of the branches in the main ReconcilePostgres/Redis function if you would like - codecov is here https://app.codecov.io/gh/integr8ly/cloud-resource-operator/compare/468/diff

valerymo:
I installed Rhoam with CRO (branch MGDAPI-4355) - and allowUpdates property is missing in CR

valerymo:
you mean - need add omitempty in ResourceTypeSpec ? ,  I could try but actually it's not for review

adam-cattermole:
Aha I forgot to update the CRD's so the field wasn't present. New commit with the CRD's is in now.

pkg/providers/aws/provider_postgres.go
```diff
@@ -147,6 +147,19 @@ func (p *PostgresProvider) ReconcilePostgres(ctx context.Context, pg *v1alpha1.P
 		return nil, croType.StatusMessage(msg), errorUtil.Wrap(err, msg)
 	}
 
+	// setup aws RDS instance sdk session
+	sess, err := CreateSessionFromStrategy(ctx, p.Client, providerCreds, strategyConfig)
```
valerymo, line 151: CreateSessionFromStrategy(..) seems could be moved to "if updatesAllowed {..."

adam-cattermole:
Ran e2e manually on a cluster and all tests passed so possibly prow being flaky but cannot say for sure:
```
Ran 15 of 15 Specs in 1801.158 seconds
SUCCESS! -- 15 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestAPIs (1811.12s)
PASS
ok      github.com/integr8ly/cloud-resource-operator/test/e2e   1811.138s
```

pkg/providers/aws/provider_postgres.go
```diff
@@ -326,22 +342,24 @@ func (p *PostgresProvider) reconcileRDSInstance(ctx context.Context, cr *v1alpha
...
-		// check if found instance and user strategy differs, and modify instance
-		logger.Infof("found existing rds instance: %s", *foundInstance.DBInstanceIdentifier)
-		mi, err := buildRDSUpdateStrategy(rdsCfg, foundInstance, cr)
-		if err != nil {
-			errMsg := fmt.Sprintf("error building update config for rds instance: %s", *foundInstance.DBInstanceIdentifier)
-			return nil, croType.StatusMessage(errMsg), errorUtil.Wrap(err, errMsg)
-		}
-		if mi != nil {
-			_, err := rdsSvc.ModifyDBInstance(mi)
+		if maintenanceWindow {
```
valerymo, line 345: this part of code seems not covered by Unit tests

valerymo:
- I checked code, logics, UTs. Looks Good for me. Code coverage could be a bit better, some new sections are not covered in UTs. But total coverage looks ok
- New local release of CRO was created for local test - cloud-resource-operator/bundles/0.40.1
- CRO service/bundle/index images created in repo: https://quay.io/repository/vmogilev_rhmi/cloud-resource-operator?tab=tags&tag=v0.40.1
- CRO was installed with Rhoam (olm install, cluster - valerym02 (not ccs)). Rhoam index: https://quay.io/repository/vmogilev_rhmi/managed-api-service?tab=tags  (v1.26.1)
  - Rhoam was pointed to use local CRO, from branch MGDAPI-4355:  go.mod: replace : github.com/integr8ly/cloud-resource-operator => /home/vmogilev/GitHub/cloud-resource-operator
  - Redis and Postgres CRDs created, and contains new property:
    - oc get crd postgres.integreatly.org  -oyaml;  oc get crd redis.integreatly.org  -oyaml
  -             maintenanceWindow:
                type: boolean

- Instances of Postgres and Redis CRDs in redhat-rhoam-operator are Missing new property maintenanceWindow, but it need to be handled in Integreatly.
- Suggest to open follow-on Jira - Integration with Integreatly, that will include both Integreatly update to use new Property, and Integration Test (suggest 5 points). I will add this note also to Integreatly PR
- Merge this PR before Integretly PR https://github.com/integr8ly/integreatly-operator/pull/2809/files
- It looks good for me

pkg/providers/aws/provider_postgres.go
adam-cattermole: I think this is covered by the new / modified original tests in ther postgres_provider_test.go. You can see the covered lines here https://app.codecov.io/gh/integr8ly/cloud-resource-operator/pull/468 and all of the lines in that maintenanceWindow block are covered.

valerymo:
code logics suppose was verified in development stage and in UTs (although more UTs were recommended, but optional). In Review - just sanity of Integration with clusterstorage - true. I wrote before that additional development and integration testing (including all logics of checking) is required at Rhoam side (new task suggested for this). So I think that lgtm is ok. But I'm creating one more cluster (ccs) and will install Rhoam with cluster storage false, and it can be used for more verification. cc @adam-cattermole

valerymo:
New - CCS cluster is ready. Rhoam (clusterStorage false) installation is running.
Using project "default".
Console: https://console-openshift-console.apps.vmogilev-ccs.2np1.s1.devshift.org/
Console Login: https://oauth-openshift.apps.vmogilev-ccs.2np1.s1.devshift.org/login?then=%2Foauth%2Fauthorize%3Fclient_id%3Dconsole%26idp%3Dkubeadmin%26redirect_uri%3Dhttps%253A%252F%252Fconsole-openshift-console.apps.vmogilev-ccs.2np1.s1.devshift.org%252Fauth%252Fcallback%26response_type%3Dcode%26scope%3Duser%253Afull
{
  "user": "kubeadmin",
  "password": "6xqIK-fpm6y-VkVFq-zcPaG"
}
The cluster could be using for additional verification

valerymo:
@adam-cattermole , could I ask - why we don't have default *maintenanceWindow: false*  in CRs?(similar to other properties, like secretRef, tier, ...), If this property for User, how user could know the name of the property ? Seems it will be nice to have it, but it's an improvement, so no impact on /lgtm  cc @laurafitzgerald

adam-cattermole:
> @adam-cattermole , could I ask - why we don't have default _maintenanceWindow: false_ in CRs?(similar to other properties, like secretRef, tier, ...), If this property for User, how user could know the name of the property ? Seems it will be nice to have it, but it's an improvement, so no impact on /lgtm cc @laurafitzgerald @MStokluska

We set it in the ones that are created from CRO (https://github.com/integr8ly/cloud-resource-operator/pull/468/files#diff-8874384cb56827b59626fda8e175afcac39b11330fbd2752dea1c6fb6c92c036R18). The omitempty fields are the same for all user-defined CR's.

PR #500, opened 2022-09-01
MGDAPI-4554 - update expected file structure of sts secret and remove needing token file path field

KevFan:
## Overview
* Update the expected sts secret format
* Remove the need for token file path as it's hardcoded in the pod template anyway
  * https://github.com/integr8ly/cloud-resource-operator/blob/master/config/manager/manager.yaml#L36-L38

Jira: https://issues.redhat.com/browse/MGDAPI-4554

## Verification
* Passing prow checks should do

## Checklist


austincunningham:
/lgtm
coverage dropped but you are removing code so that's the reason

PR #513, opened 2022-09-19
MGDAPI-4472 CRO can reconnect in the case of a postgres cr recreation

valerymo:
## Overview

Jira: https://issues.redhat.com/browse/MGDAPI-4472

## WHAT
If a postgres is removed from the cluster in a case where the RDS is still in place, recreation of a postgres cr could reconnect with the RDS and CRO puts back in place the requried connection details into the cluster for applications to reconnect.

### SCENARIO

Install CRO, install a Postgres cr, Delete the namespace which removes CRO and the postgres but not the DB because the CRO controller wont have time to cleanup.

Reinstall CRO and the Postgres cr. Does it put back the connection secret etc. If not update the logic in CRO to cover this.

## Verification

- Clone this branch
- Run `make cluster/prepare`
- Run `make run`
- Follow as in Scenario

## Checklist
Please see Scenario - CRO recovers Postgres CR after deletion

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -128,7 +128,7 @@ cluster/seed/managed/postgressnapshot:
 
 .PHONY: cluster/clean
 cluster/clean:
-	@$(KUSTOMIZE) build config/crd | kubectl delete -f -
+	@$(KUSTOMIZE) build config/crd | oc delete -f -
 	@$(KUSTOMIZE) build config/rbac | oc delete --force -f -	
 	oc delete project $(NAMESPACE)
 
--- a/pkg/providers/aws/provider_postgres.go
+++ b/pkg/providers/aws/provider_postgres.go
@@ -379,6 +379,14 @@ func (p *PostgresProvider) reconcileRDSInstance(ctx context.Context, cr *v1alpha
 			Database: *foundInstance.DBName,
 			Port:     int(*foundInstance.Endpoint.Port),
 		}
+
+		if !annotations.Has(cr, ResourceIdentifierAnnotation) {
+			statusMsg, err := addAnnotation(ctx, p.Client, cr, *rdsCfg.DBInstanceIdentifier)
+			if err != nil {
+				return nil, statusMsg, err
+			}
+		}
+
 		// return secret information
 		return &providers.PostgresInstance{DeploymentDetails: pdd}, croType.StatusMessage(fmt.Sprintf("%s, aws rds status is %s", msg, *foundInstance.DBInstanceStatus)), nil
 	}
@@ -404,9 +412,9 @@ func (p *PostgresProvider) reconcileRDSInstance(ctx context.Context, cr *v1alpha
 		return nil, croType.StatusMessage(fmt.Sprintf("error creating rds instance %s", err)), err
 	}
 
-	annotations.Add(cr, ResourceIdentifierAnnotation, *rdsCfg.DBInstanceIdentifier)
-	if err := p.Client.Update(ctx, cr); err != nil {
-		return nil, "failed to add annotation", err
+	statusMsg, err := addAnnotation(ctx, p.Client, cr, *rdsCfg.DBInstanceIdentifier)
+	if err != nil {
+		return nil, statusMsg, err
 	}
 	return nil, "started rds provision", nil
 }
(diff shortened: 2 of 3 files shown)


pkg/providers/aws/provider_postgres.go
```diff
@@ -379,6 +379,14 @@ func (p *PostgresProvider) reconcileRDSInstance(ctx context.Context, cr *v1alpha
 			Database: *foundInstance.DBName,
 			Port:     int(*foundInstance.Endpoint.Port),
 		}
+
+		if !annotations.Has(cr, ResourceIdentifierAnnotation) {
```
cecobask, line 383: I believe this path isn't covered by unit tests. Would you be able to add a test case for it, if possible?

cecobask:
Verification:
- Created an AWS CCS cluster
- Ran CRO locally
- Created a Postgres CR and waited for completion
- Deleted the CRO namespace - had to manually remove the `cloud-resources-operator.integreatly.org/finalizers` finalizer from the Postgres CR in order to proceed
- Ran CRO locally again
- Created another Postgres CR with the same spec as earlier
- Confirmed that CRO reconnected to the existing RDS instance and had the secret reference in place

@valerymo I'm ready to place an `lgtm` but I'll wait until the unit test case for the missing coverage is added 👍🏻

pkg/providers/aws/provider_postgres.go
valerymo: @cecobask Thank you for comment and discussion. This line is covered, but error is not (L 386). Problem actually seems not to cover the error, but in checking Annotations. It's Not done in any place for Redis, Blobstorage, Postgres. If we just mock this, as was discussed, we will do almost nothing, checking error message, but will add significant duplication of code in test (as to come to Annotation related code - we will need duplicate (for negative case) 2 tests: "test rds CreateReplicationGroup is called (valid cluster bundle subnets)" and "test rds exists and is available (valid cluster bundle subnets)". But this duplication in test is not look nice. Need see how to test annotations better in All places. For this probably will need do changes in Instance creation (in test) and in buildTestPostgresCR. Maybe you want open separate Jira? Thanks again for this comment.

cecobask:
Thank you for adding the test cases, @valerymo!
/lgtm

cecobask (approved):
/lgtm

Makefile
```diff
@@ -128,7 +128,7 @@ cluster/seed/managed/postgressnapshot:
 
 .PHONY: cluster/clean
 cluster/clean:
-	@$(KUSTOMIZE) build config/crd | kubectl delete -f -
+	@$(KUSTOMIZE) build config/crd | oc delete -f -
```
laurafitzgerald, line 131: any reason to use oc over kubectl here?

pkg/providers/aws/provider_postgres.go
```diff
@@ -1324,3 +1332,11 @@ func (p *PostgresProvider) rdsApplyServiceUpdates(rdsSvc rdsiface.RDSAPI, servic
 	}
 	return upgrading, "completed check for service updates", nil
 }
+
+func addAnnotation(ctx context.Context, client client.Client, cr *v1alpha1.Postgres, rdsDBInstanceIdentifier string) (string, error) {
+	annotations.Add(cr, ResourceIdentifierAnnotation, rdsDBInstanceIdentifier)
+	if err := client.Update(ctx, cr); err != nil {
+		return "failed to add annotation", err
+	}
+	return "", nil
```
laurafitzgerald, line 1341: can we use cro.StatusEmpty here.

Makefile
valerymo: @laurafitzgerald  I didn't have `kubectl` locally : ), so i found that it was the only place where `oc` was not used but kubectl. But , as I see now - Makefile changed, and maybe need rebase ... No already `kubectl` anyway in Makefile

pkg/providers/aws/provider_postgres.go
valerymo: Thank you, I will check

Makefile
valerymo: sorry, ignore please my reply, regarding rebase & missing kubectl. I looked at not correct Makefile.
So answer -  I didn't have kubectl locally : ), and this is the only place where kubectl was used, so didn't see any reason to use it

laurafitzgerald: ok. no problem. I think we had agreed a long time ago that we would use kubectl in all cases where possible and only oc where we had to but looks like it's used elsewhere so I think this is fine.

PR #539, opened 2022-10-24
MGDAPI-4719 bump CRO dependency to prepare for OSD 4.12 

MStokluska:
## Overview
Bump CRO dependency to support changes to other operators that RHOAM consumes to keep k8s api and controller runtime version in line. 

Jira: https://issues.redhat.com/browse/MGDAPI-4719

Goal was to bump deps and clear out the replaces section as it got very tangled and unnecessary. 

## Verification
- Verify that the e2e pass
- Verify that the unit test pass
- Verify CRO can provision AWS resources as before


adam-cattermole:
All the code changes look reasonable, and I've verified on a CCS cluster that I can deploy / delete all resources in AWS.

/lgtm

adam-cattermole:
The coverage check should be overridden here, although I don't think I have permission to do it myself

/override codecov/patch

PR #542, opened 2022-10-26
MGDAPI-4666 implement gcp redis delete

cecobask:
## Overview

Jira: https://issues.redhat.com/browse/MGDAPI-4666

## Verification

- Provision a GCP cluster
  - From the master branch of the [Delorean](https://github.com/integr8ly/delorean) repo create a file in the root of the repo called gcp_service_account.json. Request GCP credentials and add these to the file
  - Run:
    ```
    make -f make/ocm.mk ocm/cluster.json BYOC=true CLOUD_PROVIDER=gcp OCM_CLUSTER_REGION=europe-west2
    ```
  - Edit the cluster.json file to update the name (and remove the expiration timestamp if necessary). You can also reduce the number of nodes to 3
  - Run:
    ```
    make ocm/cluster/create
    ```

- Create a redis instance on GCP
  - Install [gcloud cli](https://cloud.google.com/sdk/docs/install)
  - Login and create a redis instance by running:
    ```
    gcloud redis instances create example-redis --size=5 --region=europe-west2
    ```
  - View the current list of redis instances:
    ```
    gcloud redis instances list --region=europe-west2
    ```
  - Alternatively you can view your instance on the GCP [console](https://console.cloud.google.com/memorystore/redis/instances?referrer=search&project=rhoam-317914)

- Delete the redis instance
  - Checkout my pull request
    ```
    gh pr checkout 542
    ```
  - Log in to your cluster and run:
    ```
    make cluster/prepare
    ```
  - Create a Redis CR with all needed properties to mimic it being created by CRO:
    ```
    PROVIDER=gcp make cluster/seed/redis
    oc patch redis/example-redis -n cloud-resource-operator --type=merge -p '{"metadata":{"annotations":{"resourceIdentifier":"example-redis"},"finalizers":["cloud-resources-operator.integreatly.org/finalizers"]},"spec":{"skipCreate":true}}'
    ```
  - Run CRO:
    ```
    make run
    ```
  - Delete the Redis CR:
    ```
    oc delete redis/example-redis -n cloud-resource-operator
    ```
  - Check that the GCP redis instance has been deleted:
    ```
    gcloud redis instances list --region=europe-west2
    ```


pkg/providers/gcp/provider_redis_test.go
```diff
@@ -148,12 +148,24 @@ func TestRedisProvider_deleteRedisCluster(t *testing.T) {
...
+								GCP: &v1.GCPPlatformStatus{
+									ProjectID: "projectID",
+									Region:    "region",
+								},
+								Type: v1.GCPPlatformType,
+							}
+						case *corev1.ConfigMap:
+							cr.Data = map[string]string{"redis": `{"development":{}}`}
+						}
+						return nil
```
adam-cattermole, line 168: After our discussion on unit testing did we not agree on using the moq client with a set of objects instead of explicitly overriding the functions? I thought the intention was to override explicitly only when necessary (i.e. k8s api errors)

pkg/providers/gcp/provider_redis.go
```diff
@@ -51,114 +54,195 @@ func (rp RedisProvider) GetReconcileTime(r *v1alpha1.Redis) time.Duration {
...
 
-func (p *RedisProvider) CreateRedis(ctx context.Context, r *v1alpha1.Redis) (*providers.RedisCluster, croType.StatusMessage, error) {
-	logger := p.Logger.WithField("action", "CreateRedis")
+func (rp *RedisProvider) CreateRedis(ctx context.Context, r *v1alpha1.Redis) (*providers.RedisCluster, croType.StatusMessage, error) {
+	logger := rp.Logger.WithField("action", "CreateRedis")
 	logger.Infof("reconciling redis %s", r.Name)
-	// handle provider-specific finalizer
-	if err := resources.CreateFinalizer(ctx, p.Client, r, DefaultFinalizer); err != nil {
-		return nil, "failed to set finalizer", err
+	var statusMessage string
```
adam-cattermole, line 63: I'm not overly bothered by this but is there any benefit to explicitly defining the statusMessage early? In the case of no errors it's unused and there is only a single assignment in each case on error the same as before. But if I understand correctly the original approach avoids creating the errMsg string at all on successful reconcile?

Fiona-Waters:
As per verification steps above I provisioned a GCP cluster, created a Redis instance through gcloud cli, and a Redis CR referencing this instance. I can confirm that the Redis instance is deleted successfully, as well as the Redis CR. 
/lgtm

pkg/providers/gcp/provider_redis.go
cecobask: Yes, that's correct. I've updated the statusMessage variables to not be declared before they're needed 👍🏻

pkg/providers/gcp/provider_redis_test.go
adam-cattermole: One minor note on the resolution here is perhaps we should define the duplicated code in functions instead of re-defining in each individual test. For example you can use [this function](https://github.com/adam-cattermole/cloud-resource-operator/blob/MGDAPI-4734/pkg/providers/gcp/cluster_network_provider_test.go#L63-L80) for a valid infrastructure CR. This eases maintainability, as if something changes with the CR we only need to change it in one location instead of for each test. It also makes writing new tests less error prone.

adam-cattermole: Sorry for the nitpicking - thanks for the changes

adam-cattermole:
Thanks for the rebase, this all still looks good and the previous review applies as well as verification completed successfully

/lgtm

PR #554, opened 2022-11-04
MGDAPI-4863 CRO 0.42.0 release

MStokluska:
## Overview

https://issues.redhat.com/browse/MGDAPI-4863 + tidying up the PREVIOUS_OPERATOR_VERSION env which was not used anymore

## Verification
- prow e2e pass


scripts/create-olm-bundle.sh
```diff
@@ -12,7 +12,6 @@ UPGRADE_CRO="${UPGRADE:-true}"
 VERSIONS="${BUNDLE_VERSIONS:-$LATEST_VERSION}"
 ROOT=$(pwd)
 INDEX_IMAGE=""
-PREVIOUS_OPERATOR_VERSIONS="${PREVIOUS_OPERATOR_VERSIONS}"
```
austincunningham, line 15: @MStokluska why was this line removed doesn't seem like part of the upgrade ?

MStokluska: as per the PR description
```
tidying up the PREVIOUS_OPERATOR_VERSION env which was not used anymore
```
The PREVIOUS_OPERATOR_VERSION is not used anywhere and only brings confusion in the release process

adam-cattermole: I discussed this with Michal and this variable is no longer used once we moved to the new index style and is a remnant from the original script that I didn't remove as part of that work.

PR #570, opened 2022-11-22
MGDAPI-4667 create gcp redis

cecobask:
## Overview

Jira: https://issues.redhat.com/browse/MGDAPI-4667

## Verification

- Clone this branch
- Run `make cluster/prepare`
- Run `make run`
- Ensure...


Fiona-Waters:
Followed verification steps and successfully created a Redis instance and subsequently deleted it.
Codecov is showing a few places that aren't covered by tests, otherwise looks good to me. @adam-cattermole to review code.

pkg/providers/gcp/provider_redis.go
```diff
@@ -244,34 +281,53 @@ func (p *RedisProvider) getRedisConfig(ctx context.Context, r *v1alpha1.Redis) (
...
+	}
+	if createInstanceRequest.Instance == nil {
+		createInstanceRequest.Instance = &redispb.Instance{
+			Name:              fmt.Sprintf(redisInstanceNameFormat, strategyConfig.ProjectID, strategyConfig.Region, createInstanceRequest.InstanceId),
+			Tier:              redispb.Instance_STANDARD_HA,
+			MemorySizeGb:      redisMemorySizeGB,
+			AuthorizedNetwork: strings.Split(networkAddress, "v1/")[1],
+			RedisVersion:      redisVersion,
+		}
+	}
```
adam-cattermole, line 316: I think we should check for the presence of these and set them separately, similar to what is done in Postgres/AWS. This allows the user to provide a create strategy with fewer fields (`"createStrategy": { "tier": "BASIC" }`) and still retain CRO defaults for other values.

adam-cattermole: I kind of expected we would have to set `ReservedIpRange` to the IP address range from the network reconciler - If there are multiple IP ranges is ours always chosen or is there a chance it picks the other one? If this does need to be set is the `AuthorizedNetwork` field required or redundant?

pkg/providers/gcp/provider_redis.go
```diff
@@ -23,7 +24,10 @@ import (
 
 const (
 	redisInstanceNameFormat = "projects/%s/locations/%s/instances/%s"
+	redisMemorySizeGB       = 16
```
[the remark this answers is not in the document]
adam-cattermole: After some discussion this morning we've decided to drop this to 1 GiB with no read replicas for the time being. If this needs to be increased it will be discovered during performance testing.

cecobask:
@adam-cattermole and @Fiona-Waters, thank you for the review and verifications.
I've addressed the feedback 👍🏻 Let me know if anything else needs to be added.

adam-cattermole:
Thanks for the changes, they look good as far as I'm concerned - assuming it's all fine by you @Fiona-Waters we can lgtm

Fiona-Waters:
Yes @adam-cattermole all good.
/lgtm

PR #609, opened 2023-01-09
MGDAPI-4929 - feat: allow defining cache node size in Redis CR

KevFan:
## Overview

Jira: https://issues.redhat.com/browse/MGDAPI-4929

## Verification

- Clone this branch
- Run `make cluster/prepare`
- Run `make run`
- Verify default node size is used when not specified in CR
```
```
- Verify node size is used when not specified in CR but specified in strategy config map
  - Update `cloud-resources-aws-strategies` config map for `redis` to 
  ```
  oc patch configmaps cloud-resources-aws-strategies -p '{"data":{"redis":"{\"development\": { \"region\": \"\", \"createStrategy\": {\"cacheNodeType\": \"cache.t3.small\"}, \"deleteStrategy\": {}, \"serviceUpdates\": [\"elasticache-20210615-002\"]  }}\n"}}' --type merge
  ```
  - Create second redis
  ```
  make cluster/seed/managed/redis REDIS_NAME=example-redis2
  ```
- Verify node size is used when specified in the CR
  ```
  make cluster/seed/managed/redis REDIS_NAME=example-redis3 REDIS_NODE_SIZE=cache.m5.large
  ```

image

## Checklist

### DIFF
--- a/pkg/providers/aws/provider_redis.go
+++ b/pkg/providers/aws/provider_redis.go
@@ -681,6 +681,11 @@ func (p *RedisProvider) getElasticacheConfig(ctx context.Context, r *v1alpha1.Re
 		return nil, nil, nil, nil, errorUtil.Wrap(err, "failed to unmarshal aws elasticache cluster configuration")
 	}
 
+	// Override if node size is defined in the CR spec
+	if r.Spec.Size != "" {
+		elasticacheCreateConfig.CacheNodeType = aws.String(r.Spec.Size)
+	}
+
 	elasticacheDeleteConfig := &elasticache.DeleteReplicationGroupInput{}
 	if err := json.Unmarshal(stratCfg.DeleteStrategy, elasticacheDeleteConfig); err != nil {
 		return nil, nil, nil, nil, errorUtil.Wrap(err, "failed to unmarshal aws elasticache cluster configuration")
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,8 @@ VERSION=0.42.0
 COMPILE_TARGET=./tmp/_output/bin/$(IMAGE_NAME)
 UPGRADE ?= true
 CHANNEL ?= rhmi
+REDIS_NODE_SIZE ?= ""
+REDIS_NAME ?= example-redis
 
 SHELL=/bin/bash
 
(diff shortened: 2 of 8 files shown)


pkg/providers/aws/provider_redis.go
```diff
@@ -681,6 +681,11 @@ func (p *RedisProvider) getElasticacheConfig(ctx context.Context, r *v1alpha1.Re
 		return nil, nil, nil, nil, errorUtil.Wrap(err, "failed to unmarshal aws elasticache cluster configuration")
 	}
 
+	// Override if node size is defined in the CR spec
+	if r.Spec.Size != "" {
+		elasticacheCreateConfig.CacheNodeType = aws.String(r.Spec.Size)
```
valerymo, line 686: Size (any string) -> CacheNodeType. Maybe need validation ?

valerymo (commented):
It looks good to me , just wanted to ask -- 
1) regarding Size string validation, if needed.
2) regarding description in Task "... CRO should be able to provision this node size only for 100M quota option".   How this condition checked  - `only for 100M quota option` ?

I also wanted to see it in Redis Cluster (I'm waiting for AWS credentials), although I did verification as in Verification notes, but on regular (not ccs cluster). Will update these notes after checking in AWS Redis clusters app.

pkg/providers/aws/provider_redis.go
KevFan: I would rather not maintain and perform validation to match the node sizes at https://aws.amazon.com/elasticache/pricing/ since these can change and are outside of our control

If you know an existing aws function that performs this validation, I can add it but otherwise, I think allowing the validation be done server side by aws itself is fine with me along as this error is propagated back up to the user in the CR status 🤔

KevFan:
> 2. regarding description in Task "... CRO should be able to provision this node size only for 100M quota option".   How this condition checked  - `only for 100M quota option` ?

@valerymo This condition will be checked on the RHOAM / integreatly operator side as the consumer of CRO, instead of being checked in CRO

valerymo:
for me it looks good
@Boomatang , could you please look also

PR #638, opened 2023-02-05
MGDAPI-5084 CRO Service Maintenance on GCP - Postgres and Redis

valerymo:
## Overview
[Spike] CRO Service Maintenance on GCP - Postgres
Jira: https://issues.redhat.com/browse/MGDAPI-5084
Spike doc: https://docs.google.com/document/d/1YzqR-zkL_m0-GyMvwNUrJzg5NFWHUI5AajmbCTpVm-o/edit#

## Verification
Please notes, that meanwhile we can't verify it, so please just code review

- Clone this branch
- Run `make cluster/prepare`
- Run `make run`
- Ensure...

## Checklist

### DIFF
--- a/pkg/providers/gcp/provider_postgres.go
+++ b/pkg/providers/gcp/provider_postgres.go
@@ -190,6 +190,7 @@ func (p *PostgresProvider) reconcileCloudSQLInstance(ctx context.Context, pg *v1
 	}
 
 	if maintenanceWindowEnabled {
+
 		logger.Infof("building cloudSQL update config for: %s", foundInstance.Name)
 		modifiedInstance, err := p.buildCloudSQLUpdateStrategy(gcpInstanceConfig, foundInstance)
 		if err != nil {
@@ -203,6 +204,19 @@ func (p *PostgresProvider) reconcileCloudSQLInstance(ctx context.Context, pg *v1
 				msg := fmt.Sprintf("failed to modify cloudsql instance: %s", foundInstance.Name)
 				return nil, croType.StatusMessage(msg), errorUtil.Wrap(err, msg)
 			}
+
+			// Disable Maintenance - set Deny
+			logger.Infof("disable Maintenance - set Deny maintenance period 90 days (max allowed by GCP)")
+			timeNow := time.Now()
+			modifiedInstance.Settings.DenyMaintenancePeriods = []*sqladmin.DenyMaintenancePeriod{{
+				EndDate:   timeNow.Add(time.Hour * 2160).Format("2006-01-02"),
+				StartDate: timeNow.Format("2006-01-02"),
+			}}
+			_, err = sqladminService.ModifyInstance(ctx, strategyConfig.ProjectID, foundInstance.Name, modifiedInstance)
+			if err != nil && !resources.IsConflictError(err) {
+				msg := fmt.Sprintf("failed to disable Maintenance - failed to modify cloudsql instance")
+				return nil, croType.StatusMessage(msg), errorUtil.Wrap(err, msg)
+			}
 		}
 		_, err = controllerutil.CreateOrUpdate(ctx, p.Client, pg, func() error {
 			pg.Spec.MaintenanceWindow = false
@@ -210,6 +224,12 @@ func (p *PostgresProvider) reconcileCloudSQLInstance(ctx context.Context, pg *v1
 		})
 		if err != nil {
 			msg := "failed to set postgres maintenance window to false"
+			return nil, croType.StatusMessage(msg), errorUtil.Wrap(err, msg)
+		}
+	} else {
+		err = p.checkUpdateDenyPeriod(ctx, sqladminService, strategyConfig.ProjectID, foundInstance)
+		if err != nil {
+			msg := "error check and update Maintenance deny period for cloudsql instance"
 			return nil, croType.StatusMessage(msg), errorUtil.Wrap(err, msg)
 		}
 	}
@@ -636,6 +656,12 @@ func (p *PostgresProvider) buildCloudSQLUpdateStrategy(cloudSQLConfig *gcpiface.
 		return nil, nil
 	}
 
+	modifiedInstance.Settings = &sqladmin.Settings{
+		DenyMaintenancePeriods: []*sqladmin.DenyMaintenancePeriod{
+			{NullFields: []string{"EndDate", "StartDate"}},
+		},
+	}
+
 	return modifiedInstance, nil
 }
 
@@ -708,4 +734,58 @@ func buildDefaultCloudSQLSecret(p *v1alpha1.Postgres) (*v1.Secret, error) {
 		},
 		Type: v1.SecretTypeOpaque,
 	}, nil
+}
+
+func (p *PostgresProvider) checkUpdateDenyPeriod(ctx context.Context, sqladminService gcpiface.SQLAdminService,
+	projectID string, instance *sqladmin.DatabaseInstance) error {
+	// if current Time minus EndDateOnInstace < 90 days (2160 hours without few hours = 2150), do nothing,
+	// otherwise - update Maintenance Deny period. 90 days - max allowed period.
+	if instance != nil && instance.Settings != nil {
+		timeNow := time.Now()
+		if instance.Settings.DenyMaintenancePeriods != nil {
+			// get EndDate on current Instance if DenyMaintenancePeriods already set
+			// if DenyMaintenancePeriods is nil (not set) - it will be defined in modifiedInstance (see below)
+			dateString := instance.Settings.DenyMaintenancePeriods[0].EndDate
+			denyMaintenanceEndDate, err := parseMaintenanceDateString(dateString)
+			if err != nil {
+				// do nothing here, as it could be case when maintenance period changed manually.
+				// It will be replaced in modifiedInstance
+				p.Logger.Info("can't parse dateString: ")
+			} else {
+				if denyMaintenanceEndDate.Sub(timeNow).Hours() > 10 {
+					p.Logger.Info("no need update Deny maintenance period")
+					return nil
+				}
+			}
+		}
+		p.Logger.Info("updating Deny maintenance period ...")
+		modifiedInstance := instance
+		modifiedInstance.Settings = &sqladmin.Settings{
+			DenyMaintenancePeriods: []*sqladmin.DenyMaintenancePeriod{{
+				EndDate:   timeNow.Add(time.Hour * 2160).Format("2006-01-02"),
+				StartDate: timeNow.Format("2006-01-02"),
+			}},
+		}
+		_, err := sqladminService.ModifyInstance(ctx, projectID, instance.Name, modifiedInstance)
+		if err != nil && !resources.IsConflictError(err) {
+			msg := fmt.Sprintf("failed to disable Maintenance - failed to modify cloudsql instance")
+			return errorUtil.Wrap(err, msg)
+		}
+	}
+	return nil
+}
+
+func parseMaintenanceDateString(dateString string) (time.Time, error) {
+	var allowedTimestampFormats = []string{
+		"2006-1-2",
+		"2006-1-02",
+		"2006-01-2",
+		"2006-01-02",
+	}
+	for _, format := range allowedTimestampFormats {
+		if t, err := time.Parse(format, dateString); err == nil {
+			return t, nil
+		}
+	}
+	return time.Now(), fmt.Errorf("can't parse maintenance date string, not supported format")
 }
--- a/pkg/providers/gcp/provider_redis.go
+++ b/pkg/providers/gcp/provider_redis.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"google.golang.org/protobuf/types/known/fieldmaskpb"
+	"google.golang.org/protobuf/types/known/timestamppb"
 	"reflect"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"strings"
@@ -154,6 +155,10 @@ func (p *RedisProvider) createRedisInstance(ctx context.Context, networkManager 
 		return nil, croType.StatusMessage(statusMessage), errorUtil.Wrap(err, statusMessage)
 	}
 	if maintenanceWindowEnabled {
+		//rescudule maintenance to start immediately
+		foundInstance.MaintenanceSchedule = &redispb.MaintenanceSchedule{
+			StartTime: timestamppb.New(time.Now()),
+		}
 		if updateInstanceRequest := p.buildUpdateInstanceRequest(createInstanceRequest.Instance, foundInstance); updateInstanceRequest != nil {
 			_, err = redisClient.UpdateInstance(ctx, updateInstanceRequest)
 			if err != nil {
@@ -175,8 +180,16 @@ func (p *RedisProvider) createRedisInstance(ctx context.Context, networkManager 
 		if err != nil {
 			statusMessage := "failed to set redis maintenance window to false"
 			return nil, croType.StatusMessage(statusMessage), errorUtil.Wrap(err, statusMessage)
+
+		}
+	} else {
+		err = p.rescheduleMaintenance(ctx, redisClient, foundInstance)
+		if err != nil {
+			msg := "error check or reschedule/defer redis instance maintenance"
+			return nil, croType.StatusMessage(msg), errorUtil.Wrap(err, msg)
 		}
 	}
+
 	rdd := &providers.RedisDeploymentDetails{
 		URI:  foundInstance.Host,
 		Port: int64(foundInstance.Port),
@@ -509,5 +522,46 @@ func (p *RedisProvider) buildUpgradeInstanceRequest(instanceConfig *redispb.Inst
 			RedisVersion: instanceConfig.RedisVersion,
 		}
 	}
+	return nil
+}
+
+func (p *RedisProvider) rescheduleMaintenance(ctx context.Context, redisClient gcpiface.RedisAPI, instance *redispb.Instance) error {
+	// Redis maintenance can be deferred up to 7 days - to next week
+	// If No maintenance has been scheduled yet - do nothing (see next note):
+	// RescheduleMaintenance could be done only for instances with valid upcoming maintenance event, othewise - error "FAILED_PRECONDITION"
+	// If MaintenancePolicy is not set (set to Any day), - Updates may occur any day of the week, need set Policy manually (only once)
+	// or MaintenancePolicy need to be set in Instance creation time (for example - Sunday, 01:00)
+	if instance == nil {
+		return errorUtil.Errorf("redis Instance is null, can't reschedule redis maintenance")
+	}
+	if instance.MaintenancePolicy == nil {
+		p.Logger.Info("no maintenance policy set, unable to defer maintenance")
+		return nil
+	}
+	if instance.MaintenanceSchedule == nil {
+		p.Logger.Info("no maintenance has been scheduled yet. No maintenance event.")
+		return nil
+	}
+	scheduledStartTime := instance.MaintenanceSchedule.StartTime.AsTime()
+	hoursBeforeMaintenance := scheduledStartTime.Sub(time.Now()).Hours()
+	msg := fmt.Sprintf("time remaining until maintenance: %s hours", fmt.Sprintf("%f", hoursBeforeMaintenance))
+	p.Logger.Info(msg)
+	if hoursBeforeMaintenance > 10 {
+		p.Logger.Info("no need reschedule maintenance")
+		return nil
+	}
+
+	p.Logger.Info("Starting Reschedule Maintenance ...")
+	redisInstanceResourceName := "https://redis.googleapis.com/v1/" + instance.GetName()
+	req := &redispb.RescheduleMaintenanceRequest{
+		Name:           redisInstanceResourceName,
+		RescheduleType: redispb.RescheduleMaintenanceRequest_SPECIFIC_TIME,
+		ScheduleTime:   timestamppb.New(scheduledStartTime.Add(time.Hour * 168)),
+	}
+	_, err := redisClient.RescheduleMaintenance(ctx, req)
+	if err != nil {
+		return errorUtil.Errorf("error in RescheduleMaintenance")
+	}
+
 	return nil
 }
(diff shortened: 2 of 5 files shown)


pkg/providers/gcp/provider_postgres.go
```diff
@@ -203,6 +211,18 @@ func (p *PostgresProvider) reconcileCloudSQLInstance(ctx context.Context, pg *v1
 				msg := fmt.Sprintf("failed to modify cloudsql instance: %s", foundInstance.Name)
 				return nil, croType.StatusMessage(msg), errorUtil.Wrap(err, msg)
 			}
+
+			// Disable Maintenance - set Deny
+			logger.Infof("disable Maintenance - set Deny maintenance period 90 days (max allowed by GCP)")
+			modifiedInstance.Settings.DenyMaintenancePeriods = []*sqladmin.DenyMaintenancePeriod{{
+				EndDate:   time.Now().Add(time.Hour * 2160).Format("01-02-2006"),
+				StartDate: time.Now().Format("01-02-2006"),
```
adam-cattermole, line 219: There is a very small chance that this results in >90 days due to two separate calls to `time.Now()`. This should be changed to set a variable to `time.Now()` and use this as the base for the start and end date.

pkg/providers/gcp/provider_postgres.go
```diff
@@ -472,7 +498,30 @@ func (p *PostgresProvider) buildCloudSQLCreateStrategy(ctx context.Context, pg *
...
+					RetentionUnit:   defaultBackupRetentionSettingsRetentionUnit,
+					RetainedBackups: defaultBackupRetentionSettingsRetainedBackups,
+				},
+			},
+			DataDiskSizeGb:            defaultDataDiskSizeGb,
+			DeletionProtectionEnabled: utils.Bool(defaultDeleteProtectionEnabled),
+			IpConfiguration: &gcpiface.IpConfiguration{
+				Ipv4Enabled: utils.Bool(defaultIPConfigIPV4Enabled),
+			},
+			UserLabels: tags,
```
adam-cattermole, line 519: I think this is a rebase error, setting all of the defaults in the initial struct was removed in a recent PR. Instead you can check that the `DenyMaintenancePeriods != nil` and set it there.

pkg/providers/gcp/provider_redis.go
```diff
@@ -154,6 +161,10 @@ func (p *RedisProvider) createRedisInstance(ctx context.Context, networkManager
 		return nil, croType.StatusMessage(statusMessage), errorUtil.Wrap(err, statusMessage)
 	}
 	if maintenanceWindowEnabled {
+		//rescudule maintenance to start immediately
+		foundInstance.MaintenanceSchedule = &redispb.MaintenanceSchedule{
+			StartTime: timestamppb.New(time.Now()),
+		}
```
adam-cattermole, line 167: I'm wondering if this has been tested at all - looking at the redispb library it states that `MaintenanceSchedule` is output-only -> https://github.com/integr8ly/cloud-resource-operator/blob/mgdapi-3425-gcp/vendor/google.golang.org/genproto/googleapis/cloud/redis/v1/cloud_redis.pb.go#L641-L643

pkg/providers/gcp/provider_postgres.go
```diff
@@ -203,6 +211,18 @@ func (p *PostgresProvider) reconcileCloudSQLInstance(ctx context.Context, pg *v1
...
 				return nil, croType.StatusMessage(msg), errorUtil.Wrap(err, msg)
 			}
+
+			// Disable Maintenance - set Deny
+			logger.Infof("disable Maintenance - set Deny maintenance period 90 days (max allowed by GCP)")
+			modifiedInstance.Settings.DenyMaintenancePeriods = []*sqladmin.DenyMaintenancePeriod{{
+				EndDate:   time.Now().Add(time.Hour * 2160).Format("01-02-2006"),
+				StartDate: time.Now().Format("01-02-2006"),
+			}}
+			_, err = sqladminService.ModifyInstance(ctx, strategyConfig.ProjectID, foundInstance.Name, modifiedInstance)
```
adam-cattermole, line 221: I think that there is a chance that we remove the deny period and apply it again before the maintenance is scheduled so I'm not sure if this will work as intended. I think we need to see a maintenance event so that we know how to determine whether the maintenance has been performed or not.

pkg/providers/gcp/provider_postgres.go
```diff
@@ -709,3 +764,29 @@ func buildDefaultCloudSQLSecret(p *v1alpha1.Postgres) (*v1.Secret, error) {
 		Type: v1.SecretTypeOpaque,
 	}, nil
 }
+
+func checkUpdateDenyPeriod(ctx context.Context, sqladminService gcpiface.SQLAdminService,
+	projectID string, instance *sqladmin.DatabaseInstance) error {
+	// if current Time(date) minus lastMaintenanceTime < 90 days (2160 hours without few hours = 2150), do nothing,
+	// otherwise - update Maintenance Deny period
+	currentTime := time.Now()
+	if currentTime.Sub(lastMaintenanceTime) < 2150 {
```
adam-cattermole, line 773: Do we have to save `lastMaintenanceTime`? can we not check the `EndDate` on the instance and compare to current time?

adam-cattermole (commented):
Overall this looks like a good start but I think there's still some work left. My thinking is that we should opt to just extend the deny maintenance period for the time being for Postgres and not try and perform maintenance as I'm uncertain the solution will work.

For Redis the solution seems a lot less concrete and I'm not sure we have a solution to prevent the updates. Interested in your thoughts on this @valerymo

pkg/providers/gcp/provider_postgres.go
valerymo: Thank you! Done

pkg/providers/gcp/provider_postgres.go
valerymo: func main() {
    EndDate := time.Now().Add(time.Hour * 2160).Format("01-02-2006") // 90 days (max allowed by GCP)
    StartDate := time.Now().Format("01-02-2006")
    fmt.Println("StartDate: ", StartDate)
    fmt.Println("EndDate: ", EndDate)
}
$ go run test1.go 
StartDate:  02-15-2023
EndDate:  05-16-2023

pkg/providers/gcp/provider_redis.go
valerymo: Thank you, yes, I saw it, but maybe need try. If will not work, we can use REST, same way as gcloud

pkg/providers/gcp/provider_postgres.go
valerymo: Thank you. If I understand the comment, -  after the Maintenance update applying  and `sqladminService.ModifyInstance(.....)` , need `WAIT` before set Disable update (and modify instance again).
Yes, I thought, but in current scenario - the second modification should not impact, as maintenance update already started. If it's not succeed, -  we will have another/next `maintenanceWindowEnabled` chance. Maybe small sleep, but seems to me no need  check/wait if maintenance update succeed.

pkg/providers/gcp/provider_postgres.go
valerymo: compare EndDate on the instance with  current time?
It looks very good idea!
Thank you very much

pkg/providers/gcp/provider_postgres.go
adam-cattermole: It's because of the processing delay between calls it can be wrong. If first time.Now() happens at day 1 11:59:59.9 the second time.Now() could be done on the next day. If you save it to a variable it'll be the same exact time and won't require two separate calls to the system clock

pkg/providers/gcp/provider_postgres.go
adam-cattermole: My concern is that it will remove the deny from the instance and then apply it again without the maintenance taking place. Normally once CRO successfully reconciles in this case it removes the maintenaceWindow flag from the Postgres CR so the update might not take place until the next service affecting RHOAM release

pkg/providers/gcp/provider_postgres.go
valerymo: yes, sure, 
Hope I understand it correctly now:
		timeNow := time.Now()
.....
		EndDate:   timeNow.Add(time.Hour * 2160).Format("01-02-2006"), 
		StartDate: timeNow.Format("01-02-2006"),
....
Done. Thank you

pkg/providers/gcp/provider_postgres.go
valerymo: agree, maybe risk of such scenario exists, as you noted "remove deny, update not stated yet, apply deny". So update will not happen. Don't know ... maybe sleep (although I also not so like it, but meanwhile and open case to improve it), but not wait for update completion; maybe it will take hours. Seems need somehow check if maintenance upgrade started. Maybe another ticket with exact problem description.

adam-cattermole: Yep I think perhaps we opt to just extend the maintenance deny for now and do not remove it - we can have a follow up ticket that looks into actually performing the update once we have an affected cluster where we can test the issue.

valerymo: sorry, if do not remove Deny - maintenance update is not possible, as denied (it's differ from AWS).  Seems better as is (+ small sleep), as seems to me.

adam-cattermole: Sorry if I'm misremembering but the suggestion we had on a call is to update the deny period continuously for now so that no service maintenance takes place in the first GCP CRO release. Once we get a new maintenance event it will be denied until we implement a solution that we know is working (as we will be able to test it) and we will release a new CRO version and RHOAM release to roll out the maintenance. In my opinion this is a safer approach that gives us time to get the solution working correctly.

valerymo: This is exactly what we have now. Maintenance will never happen if `maintenanceWindowEnabled  is false.
In this case the flow is go to `checkUpdateDenyPeriod`. And here - we continuously update the deny period.

valerymo:
Hi @adam-cattermole , seems I didn't reply yet  on your comment regarding "For Redis the solution seems a lot less concrete and I'm not sure we have a method to prevent the updates. Interested in your thoughts on this @valerymo"
I didn't write too much about this in Spike doc, as after checking Postgres i thought it's easy. But anyway notes for Redis available at the end of doc https://docs.google.com/document/d/1YzqR-zkL_m0-GyMvwNUrJzg5NFWHUI5AajmbCTpVm-o/edit#heading=h.mty7yk5h2ifn  . And implementation for Redis that done in this PR, I hope will work. But I will try add here few notes.   
Redis mainatenance looks a bit differ from Postgres. There are No Deny Period. But maybe it even more simple -  Just reschedule maintenance date, change it for future. It could be done continuosly. And this is I tried to implement in this PR.  The difference with Postgres also in period; in Postgres - it 90 days. In Redis - 7 days.

pkg/providers/gcp/provider_postgres.go
adam-cattermole: That's correct but if there is a serviceAffecting RHOAM release where we have not had time to fix the issue then these untested changes will run. If you really are adamant to keep this in then we can but in my opinion we should only have tested working changes in the releases.

pkg/providers/gcp/provider_redis.go
adam-cattermole: I have commented out a nil pointer error and tested this and it is not possible to set like this:

```
"error": "failed to update gcp redis instance projects/REDACTED/locations/europe-west2/instances/REDACTED: rpc error: code = InvalidArgument desc = Invalid option 'maintenance_schedule': 'maintenance_schedule' cannot be updated on existing instances.\ncom.google.apps.framework.request.StatusException: <eye3 title='INVALID_ARGUMENT'/> generic::INVALID_ARGUMENT: Invalid option 'maintenance_schedule': 'maintenance_schedule' cannot be updated on existing instances.\nerror details: name = BadRequest field = update_mask.paths.0 desc = Invalid value: maintenance_schedule", "errorVerbose": "rpc error: code = InvalidArgument desc = Invalid option 'maintenance_schedule': 'maintenance_schedule' cannot be updated on existing instances.\ncom.google.apps.framework.request.StatusException: <eye3 title='INVALID_ARGUMENT'/> generic::INVALID_ARGUMENT: Invalid option 'maintenance_schedule': 'maintenance_schedule' cannot be updated on existing instances.\nerror details: name = BadRequest field = update_mask.paths.0 desc = Invalid value: maintenance_schedule\nfailed to update gcp redis instance
```

pkg/providers/gcp/provider_redis.go
```diff
@@ -511,3 +524,31 @@ func (p *RedisProvider) buildUpgradeInstanceRequest(instanceConfig *redispb.Inst
...
+
+func (p *RedisProvider) checkRescheduleDeferMaintenance(ctx context.Context, redisClient gcpiface.RedisAPI, instance *redispb.Instance) error {
+	// Redis maintenance can be deferred up to 7 days
+	// If current Time minus lastScheduledMaintananceTime < 7 days (< 168 h), do nothing, otherwise - reschedule
+	if instance == nil {
+		return errorUtil.Errorf("redis Instance is null, can't reschedule redis maintenance")
+	}
+	lastScheduledMaintananceTime := instance.MaintenanceSchedule.StartTime
+	timeNow := time.Now()
+	if timeNow.Sub(lastScheduledMaintananceTime.AsTime()) < 166 {
```
adam-cattermole, line 536: Hitting a nil pointer exception here on normal reconcile:

```
goroutine 2535 [running]:
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile.func1()
        /home/adam/go/src/github.com/integr8ly/cloud-resource-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:118 +0x1f4
panic({0x2502960, 0x46b0a00})
        /usr/lib/go/src/runtime/panic.go:884 +0x213
github.com/integr8ly/cloud-resource-operator/pkg/providers/gcp.(*RedisProvider).checkRescheduleDeferMaintenance(0x2f08b28?, {0x2f08b28, 0xc000ed1a70}, {0x2f09fa0, 0xc001cbb0a0}, 0xc0008f8ea0)
        /home/adam/go/src/github.com/integr8ly/cloud-resource-operator/pkg/providers/gcp/provider_redis.go:536 +0x7e
github.com/integr8ly/cloud-resource-operator/pkg/providers/gcp.(*RedisProvider).createRedisInstance(0xc0011a1040, {0x2f08b28, 0xc000ed1a70}, {0x2f0e690, 0xc0007ae690}, {0x2f09fa0?, 0xc001cbb0a0?}, 0xc0007aff80?, 0xc00106bd40)
        /home/adam/go/src/github.com/integr8ly/cloud-resource-operator/pkg/providers/gcp/provider_redis.go:188 +0xa32
github.com/integr8ly/cloud-resource-operator/pkg/providers/gcp.(*RedisProvider).CreateRedis(0xc0011a1040, {0x2f08b28, 0xc000ed1a70}, 0xc00106bd40)
        /home/adam/go/src/github.com/integr8ly/cloud-resource-operator/pkg/providers/gcp/provider_redis.go:102 +0x365
github.com/integr8ly/cloud-resource-operator/controllers/redis.(*RedisReconciler).Reconcile(0xc001289d00, {0x2f08b28, 0xc000ed1a70}, {{{0xc000f44048?, 0x0?}, {0xc00102c510?, 0x40fb47?}}})
        /home/adam/go/src/github.com/integr8ly/cloud-resource-operator/controllers/redis/redis_controller.go:175 +0x85e
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile(0x2f08a80?, {0x2f08b28?, 0xc000ed1a70?}, {{{0xc000f44048?, 0x28c9580?}, {0xc00102c510?, 0xc000e50d88?}}})
        /home/adam/go/src/github.com/integr8ly/cloud-resource-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:121 +0xc8
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler(0xc001008640, {0x2f08a80, 0xc000fc59a0}, {0x260c2c0?, 0xc001cba4c0?})
        /home/adam/go/src/github.com/integr8ly/cloud-resource-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:320 +0x309
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem(0xc001008640, {0x2f08a80, 0xc000fc59a0})
        /home/adam/go/src/github.com/integr8ly/cloud-resource-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:273 +0x1d9
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2()
        /home/adam/go/src/github.com/integr8ly/cloud-resource-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:234 +0x85
created by sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2
        /home/adam/go/src/github.com/integr8ly/cloud-resource-operator/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:230 +0x587
exit status 2
```

pkg/providers/gcp/provider_postgres.go
valerymo: Thank you @adam-cattermole , maybe we could discuss also in GCP chat, how to progress with this. Any decision is fine for me. Of cause that not verified things should NOT be included into Release. But if feature is Needed, it should be verified.

adam-cattermole:
I also see an error on normal reconcile for the Postgres instances - CR in failed state:

```
status:
  message: >-
    error check and update Maintenance deny period for cloudsql instance: error
    check and update Maintenance deny period for cloudsql instance: failed to
    disable Maintenance - failed to modify cloudsql instance: googleapi: Error
    400: Invalid request: Invalid deny maintenance period. Start and end date
    both must be valid Gregorian calendar dates.., invalid
phase: failed
```

pkg/providers/gcp/provider_redis.go
valerymo: Thank you very much Adam, I will check

valerymo:
Thank you. I will recheck date format, probably it should be like 2008-03-07T17:06:02.000Z /  YYYY-MM-DDTHH:MM:SS.MMMZ.  I will run and check

pkg/providers/gcp/provider_postgres.go
valerymo: Hi @adam-cattermole , postgres maintenance, Deny set/update seems working stable. I added handling of the date format issue.

pkg/providers/gcp/provider_redis.go
valerymo: Hi @adam-cattermole , the logic for Redis - updated. Also, as discussed, it's using now `RescheduleMaintenance()` function that was added to CRO redis client.  Thank you for review and suggestions.

valerymo:
Hi @adam-cattermole , @laurafitzgerald , Current solution is working , no errors, after comments from PR addressing.
Thanks to Adam for Review!  
Regarding following comment of Adam `The number of times isn't really the problem as far as I can tell, it's more so the 7 day limit from the originally scheduled time. I think the ability to reschedule multiple times let's you change the specific time the maintenance will take place as long as it's within a week of the originally scheduled time`. Yes, it could be, it's described in documentation as ` "You may reschedule maintenance multiple times so long as it's not more than one week after the originally scheduled time. You have a few scheduling options for the new maintenance window:" `. But I can't find meanwhile anything about this in GCP forums; but will continue to search.
Regarding PR - it's working with current solution, and I'd like suggest to accept it as is meanwhile.
If Redis maintenance indeed could be deferred Not more than week (if we will have more confirmation, and if we could recheck it) - probably will need revise current "Maintenance window" approach.
But regarding current PR - could you please decide if we could merge it, as it opened long time, and current solution seems working.

pkg/providers/gcp/provider_redis.go
```diff
@@ -511,3 +524,43 @@ func (p *RedisProvider) buildUpgradeInstanceRequest(instanceConfig *redispb.Inst
...
+	// Redis maintenance can be deferred up to 7 days - to next week
+	// If No maintenance has been scheduled yet - do nothing (see next note):
+	// RescheduleMaintenance could be done only for instances with valid upcoming maintenance event, othewise - error "FAILED_PRECONDITION"
+	// If MaintenancePolicy is not set (set to Any day), - Updates may occur any day of the week, need set Policy manually (only once)
+	// or MaintenancePolicy need to be set in Instance creation time (for example - Sunday, 01:00)
+	if instance == nil {
+		return errorUtil.Errorf("redis Instance is null, can't reschedule redis maintenance")
+	}
+	if instance.MaintenancePolicy == nil {
+		p.Logger.Info("missing maintenance Policy, Updates may occur any day of the week. Please set initial Maintenance Policy manually (to be done only once)")
```
adam-cattermole, line 538: Minor suggestion - I think as this is very RHOAM specific and CRO is supposed to be un-opinionated you could leave the message to be more informational rather than request the user apply the maintenance policy.
Maybe something like: "no maintenance policy set, unable to defer maintenance". There should probably also be a return here in case the policy is not set but there is a maintenance scheduled.

Other than this I think it looks good and I'm happy to merge - we can create a follow up if needed

adam-cattermole:
Thanks for the investigation and changes Valery, verified that this works for applying the defer period for Postgres and the Redis solution looks good albeit untestable. Will create a follow up ticket based on a decision whether any changes need to be made!

/lgtm

pkg/providers/gcp/provider_redis.go
valerymo: Done, Message and return - added. Condition - not changed; discussed.
Thank you

adam-cattermole:
We've decided to opt for an alternate solution here @valerymo - we're going to let the user define their maintenance window from RHOAM and update the strategy to let GCP handle the updates. This avoids any of the issues with service affecting releases and us handling the updates ourselves.

Thanks for your hard work, investigation and all the discussion on this - really helped to sculpt the direction for the way forward

valerymo:
As noted in PR - it was decided to opt for an alternate solution - to let the user define their maintenance window using the strategies and GCP handle the updates. This avoids any of the issues with service affecting releases and only being able to extend the service maintenance on Redis by 7 days.
This Spike task can be Closed

PR #683, opened 2023-05-03
MGDAPI-5691 - allow specifying redis size for GCP

adam-cattermole:
## Overview

[MGDAPI-5691](https://issues.redhat.com/browse/MGDAPI-5691)

### Provision a GCP CCS cluster

From the master branch of the [Delorean](https://github.com/integr8ly/delorean) repo create a file in the root of the repo called gcp_service_account.json. Request GCP credentials and add these to this file. Adjust this command to suit your needs, it will generate a config at ocm/cluster.json:
 ```sh
OCM_CLUSTER_NAME=acatterm-ccs OCM_CLUSTER_LIFESPAN=24 COMPUTE_NODES_COUNT=2 BYOC=true CLOUD_PROVIDER=gcp make -f make/ocm.mk ocm/cluster.json
```
Once you have checked the ocm/cluster.json, create the cluster:
```sh
make -f make/ocm.mk ocm/cluster/create
```

### Verify changes

Run the operator from this branch:

```
PROVIDER=gcp make cluster/prepare
make run
```

Create a new Redis CR with a custom size:

```
REDIS_NODE_SIZE=7 PROVIDER=gcp make cluster/seed/redis
```

You should see the redis instance get created with this memory configured. Update the CR to a different value and watch the redis instance change.

Removing the `Spec.Size` will result in the Redis instance reverting to the default of 1GB.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,7 +102,7 @@ cluster/seed/blobstorage:
 
 .PHONY: cluster/seed/redis
 cluster/seed/redis:
-	@cat config/samples/integreatly_v1alpha1_redis.yaml | sed "s/name: REPLACE_ME/name: $(REDIS_NAME)/g" | sed "s/type: REPLACE_ME/type: $(PROVIDER)/g" | sed "s/size: REPLACE_ME/size: $(REDIS_NODE_SIZE)/g" | oc apply -f - -n $(NAMESPACE)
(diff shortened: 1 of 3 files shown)


Fiona-Waters:
Verified that these changes work as expected. Updated `Spec.Size` to 3 and confirmed that the instance updated. 

Screenshot from 2023-05-03 14-46-58

Then removed `Spec.Size` from the CR and can confirm that it updated to the default 1GB.

Screenshot from 2023-05-03 15-05-01

PR #690, opened 2023-07-19
MGDAPI-5815 - configure golangci-lint checks

kramaranya:
## Overview

[MGDAPI-5815](https://issues.redhat.com/browse/MGDAPI-5815)

Configure and add golangci-lint file for running checks

## Verification

- Run the lint make command:
```
make test/lint
```
- It will install the golangci-lint binary to the bin folder of your project if it doesn't already exist there. The output of running should be empty.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,0 +1,33 @@
+output:
+  format: tab
+  sort-results: true
+
+linters:
+  disable:
+    - errcheck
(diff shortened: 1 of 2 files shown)


Patryk-Stefanski:
code looks good, running `make test/lint` installs golanci-lint to the bin folder and returns no issues.

/lgtm

